PR #1, opened 2019-04-15
Control screen debug info output by taget with OcSupportPkg

d1ves:
Control screen debug info output by taget with OcSupportPkg


vit9696:
I provided an explanation why this does not fit in a second pull request:
https://github.com/acidanthera/OcSupportPkg/pull/2#issuecomment-483342524

PR #2, opened 2019-04-26
Coverity Scan workaround

cotsog:


vit9696:
Thanks for your input, let's see how it goes!

PR #6, opened 2019-05-13
Lower "Error" Level of Kernel patcher

christian0101:
As it is set to the maximum error level currently it will halt the system. 

For instance, when booting into recovery the com.apple.driver.AppleIntelFramebufferCapri kext cannot be found and CPU halts.


Download-Fritz:
Thanks! Personally, I agree with this change.
If it should be accepted, I'd ask you to update your PR branch to be a clean master HEAD with your commit at the top of the history. You can avoid merging upstream changes into your modified branch by using "git pull --rebase" in the future, which will undo your commits, update the branch to HEAD, and then re-apply the patches. If the patches shall no longer apply, you will get a chance to merge to amend the commit.

vit9696:
We can do squash merging, and I actually made it the default merge action (and prohibited merge commits) not so long ago.

Thanks for the fix, it makes good sense to me. Merged.

PR #12, opened 2019-08-24
Update README.md

ZISQO:
Anyone can read the topic in the OpenCore Korean forum.

In Korea, scammers are getting from $300 to $600. users who can't install Hackintosh byself.
but scammer's can't speak English as well so most korean communities kept a higher membership levles.

however, I totally agree that you should use a public forum so that anyone can easily install a opencore boot loader.
so, do why i add a link to the there? it's possiblt to use public manual with Korean language.


Andrey1970AppleLife:
It is the third-party guide, we cannot control it.
It is not discussion forum.
It is a small private site.
@BarbaraPalvin - member Acidanthera from Korea, can add a worthy Korean forum if the such is available.

ZISQO:
@Andrey1970AppleLife 
I have already participated in insanelymac github with open core related guide now.
https://github.com/insanelymacdiscord/Getting-Started-With-OpenCore

Check your credit.

PR #14, opened 2019-09-01
Update Configuration.tex

cattyhouse:
Add "DevirtualiseMmio" to "How can I migrate from AptioMemoryFix" 
Correct "AptioInputFix" to "AppleGenericInput"


vit9696:
DevirtualiseMmio is not present in AMF, it is a new feature. I will correct the driver name later on my own.

PR #17, opened 2019-10-09
Fix typos in OpenCoreKernel.c

dhinakg:
There are missing line breaks in the strings on lines 318, 486, and 566.
On lines 313, 481 and 561, MinKernel is set to `OcParseDarwinVersion (OC_BLOB_GET (&XXX->MaxKernel))` instead of `OcParseDarwinVersion (OC_BLOB_GET (&XXX->MinKernel))`.


vit9696:
Thanks for the fix.

PR #18, opened 2019-10-10
Refactor `OcParseDarwinVersion`.

reitermarkus:
With this, you can set both `MinKernel` and `MaxKernel` to `19`, which will then correctly translate to `190000` and `199999`.

This also fixes parsing of e.g. `19.1`, which would previously be `191000` instead of `190100`.

### DIFF
--- a/Platform/OpenCore/OpenCoreKernel.c
+++ b/Platform/OpenCore/OpenCoreKernel.c
@@ -35,6 +35,7 @@ OcParseDarwinVersion (
   )
 {
   UINT32  Version;
+  UINT32  VersionPart;
   UINT32  Index;
   UINT32  Index2;
 
@@ -45,14 +46,26 @@ OcParseDarwinVersion (
   Version = 0;
 
   for (Index = 0; Index < 3; ++Index) {
+    Version *= 100;
+
+    VersionPart = 0;
+
     for (Index2 = 0; Index2 < 2; ++Index2) {
-      Version *= 10;
+      //
+      // Handle single digit parts, i.e. parse 1.2.3 as 010203.
+      //
+      if (*String != '.' && *String != '\0') {
+        VersionPart *= 10;
+      }
+
       if (*String >= '0' && *String <= '9') {
-        Version += *String++ - '0';
+        VersionPart += *String++ - '0';
       } else if (*String != '.' && *String != '\0') {
         return 0;
       }
     }
+
+    Version += VersionPart;
 
     if (*String == '.') {
       ++String;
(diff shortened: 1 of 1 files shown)


camielverdult:
What is happening

vit9696:
Hello,

After the third failure in this code I must say that it is a very bad idea to write code during the vacation. No idea in what senses I was :D. Thank you for another effort to correct it.

- So. I agree with the 19.1 vs 19.10 change, it is indeed a bug.
- However, I do not agree with 19.0.0 meaning 19.99.99 in MaxKernel and 19.0.0 in MinKernel. This makes it impossible to encode 19.0.0 in MaxKernel, it is not how the documentation is written, and even further it is just unintuitive (I asked a few colleagues here).
- There also are some troubles with the code style. We require that all variables are declared in the beginning of the function, and that they are initialised on demand. So your VersionPart needs to be split. We also require that comments are written with empty `//` on the lines above and below.

Please fix the patch and we can merge this.

reitermarkus:
> I do not agree with 19.0.0 meaning 19.99.99 in MaxKernel and 19.0.0 in MinKernel

`19.0.0` is still `190000` in `MaxKernel`, only `19` is `199999` and e.g. `19.1` is `190199`.

vit9696:
This is even worse, unfortunately. For me (and once again for several other people) the intuitive approach is that 19, 19.0, 19.0.0 are the same things no matter where you put them. Think of these writings as decimal fractions, in no way it should expand to 99 depending on the number of dots. Another difficulty is that parsing happens differently for MinKernel, MaxKernel, and CurrKernel. This looks just unacceptable.

reitermarkus:
Then maybe the more intuitive way would be to make `MaxKernel` exclusive? Writing `19.99.99` everywhere is also not really intuitive.

`190000 <= X < 200000`

vit9696:
It actually is intuitive when you need to restrict the match to a particular version, you just put it to max and min. It is very intuitive

reitermarkus:
I don't find `19.99.99` intuitive as the maximum because that format is basically an implementation detail. Why couldn't it be `19.999.999`, for example?

vit9696:
This is in the documentation, and even if it was not you could use the maximum version based on your knowledge about the macOS versions. We know the realistic maximum given the yearly release schedule. In fact, this is the wrong way to do it.

Setting the maximum version prior to the patch breakage (i.e. before this version release) is ridiculous in the first place. Clearly such patch/kext is wrongly written and will likely break even earlier, no matter what you put thereA: 99, 98, or even 13. For this reason I believe that there is no such use for MaxKernel at all, and if there is, it would be rare enough to live with 19.99.99 than go with all the madness.

reitermarkus:
I don't think it's uncommon to have a patch which works on all minor versions of a specific major version.

> Setting the maximum version prior to the patch breakage

If you set the maximum version prior to breakage, how do you find out when it would break if it is disabled already at that point?

vit9696:
That is the point, we *do not* set the maximum version, in e.g. AppleALC, which uses similar scheme, until the patch actually breaks. If it works for two major versions, great, we do not need to change anything. Will merge now.

PR #21, opened 2019-11-23
Kexts patch not work open core

nmano0006:
Intel i7-5960X
Motherboard-X99
 
The 5960X CPU not load without this patch.
Find       483D0000 0040
Replace      483D0000 0080
Identifier       IOPCIFamily

I can 100% Confirm open core any kext patch not work.
I can’t boot my system without this patch.
Can you fix this issue Kernel->patch=not work
Thank you.
/Users/x99/Desktop/Screen Shot 2019-11-23 at 5.34.17 AM.png


vit9696:
Sorry, are you retarded? Why not use bugtracker for such reports? From what we know, kext patching works fine.

PR #23, opened 2020-01-29
[SUGGESTION] RequireSignature if clause improvement in OpenCoreMisc.c

Gardosen:
Jumping into this if clause, makes only sense if the Misc.Security.RequireVault is set to true, as RequireSignature is only used in combination with RequireVault.

A cancel of the boot process just because RequireSignature is true, does not make sense in my opinion.
a simple debug_info about the inconsistent usage is way more user friendly and does not block users  from booting who do not use RequireVault.

### DIFF
--- a/Platform/OpenCore/OpenCoreMisc.c
+++ b/Platform/OpenCore/OpenCoreMisc.c
@@ -227,7 +227,7 @@ OcMiscEarlyInit (
     return EFI_SECURITY_VIOLATION; ///< Should be unreachable.
   }
 
-  if (VaultKey == NULL && Config->Misc.Security.RequireSignature) {
+  if (Config->Misc.Security.RequireVault && VaultKey == NULL && Config->Misc.Security.RequireSignature) {
     DEBUG ((DEBUG_ERROR, "OC: Configuration requires signed vault but no public key provided!\n"));
     CpuDeadLoop ();
     return EFI_SECURITY_VIOLATION; ///< Should be unreachable.
--- a/Platform/OpenCore/OpenCoreMisc.c
+++ b/Platform/OpenCore/OpenCoreMisc.c
@@ -227,7 +227,9 @@ OcMiscEarlyInit (
     return EFI_SECURITY_VIOLATION; ///< Should be unreachable.
   }
 
-  if (Config->Misc.Security.RequireVault && VaultKey == NULL && Config->Misc.Security.RequireSignature) {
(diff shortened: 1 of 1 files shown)


vit9696:
Thank you for your contribution, but I disagree. The code currently present in the repository allows to ensure that the user will be aware of his misconfiguration, although in a fairly brutal way. What it tries to protect from is a false assumption of "security" working by forcing to correct the configuration. It also makes sure that this false assumption never spreads by making it a boot invariant.

Turning this check into a debug print that is not even visible in release builds adds an additional security risk of letting people to be unaware of this problem at all.

Gardosen:
@vit9696 blocking the boot process of a user who is not even using a specific feature, just because a variable is not set to false, which is totally unimportant to the entire runtime while that feature is off, is bad practice and more or less points out that the architecture opencore is based off is either not existing, broken, or badly designed.
If there would be one point, where RequireSignature is used WITHOUT RequireVault, it would make sense, but how it is right now is just garbage code.

also as a side note, i just used what is already used by the code itself to keep the consistency, these DEBUG infos appear also in the release version of OpenCore, so it does not add an additional security risk.

Kind regards
Gardosen

vit9696:
DEBUG macros with DEBUG_INFO level are not present in release builds, this is documented in the reference manual. WARN and ERROR are, but the messages will appear too fast to be possibly noticed or may fail to appear at all due to misconfigured or yet unconfigured console control, for instance.

I believe the code is correct and works as designed. I also believe that the measures taken are adequate to the security risks. Yet just in case I also ask @Download-Fritz to comment on this.

Download-Fritz:
Agreed.

Gardosen:
there is no security risk if you don't block the boot for people without RequireVault.

i'm sorry to say this but this is badly designed. it should either not block the user from booting or inform him correctly about why he got blocked in this situation. but just saying

"OC: Configuration requires signed vault but no public key provided!" is not even not helpful, it is wrong. because it says "Configuration requires signed vault" but that is not true in the case of RequireVault = false. 
As RequireVault is not checked on this point, users getting blocked for no reason.

Download-Fritz:
It is fair to say users do get blocked out for no reason (in their sepcific situation), and I agree. However, this is a needed sacrafice to protect those that did intend for the security to be on when they misconfigured their setup. Security is an important paradigm for OpenCore and hence comfort is and will be sacraficed to ensure a secure setup. Here, it's being forced to flip a boolean value (comfort, our approach) vs unintentionally booting an untrusted environment (security, your approach), potentially unnoticed.

I think that is all there is to say about this topic.

PR #25, opened 2020-04-03
fix X299 TB3 hotplug (SB.SCK0.PR00)

ZISQO:
Users add a Thunderbolt.aml file to acpi for TB3 hot plug work.


vit9696:
This has nothing to do to TB3, as this is about CPU power management, but committed. Note, that you are expected to remove the irrelevant entries in any case. Thanks.

PR #26, opened 2020-04-03
Added CPU naming

Pavo-IM:
These are the CPU names for AMD Ryzen Zen and Zen 2 CPUs.

### DIFF
--- a/Docs/AcpiSamples/SSDT-PLUG.dsl
+++ b/Docs/AcpiSamples/SSDT-PLUG.dsl
@@ -65,7 +65,7 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "CpuPlug", 0x00003000)
     }
     
     If (CondRefOf (\_PR.C000)) {
-        Scope (\_PR.P000) {
+        Scope (\_PR.C000) {
             Method (_DSM, 4, NotSerialized) {
                 If (LEqual (Arg2, Zero)) {
                     Return (Buffer (One) { 0x03 })
--- a/Docs/AcpiSamples/SSDT-PLUG.dsl
+++ b/Docs/AcpiSamples/SSDT-PLUG.dsl
@@ -12,6 +12,8 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "CpuPlug", 0x00003000)
 {
     External (_SB_.CPU0, DeviceObj)
     External (_PR_.CPU0, DeviceObj)
+    External (_PR_.C000, DeviceObj)
+    External (_PR_.P000, DeviceObj)
     External (_SB_.PR00, DeviceObj)
     External (_PR_.PR00, DeviceObj)
     External (_SB_.SCK0.CP00, DeviceObj)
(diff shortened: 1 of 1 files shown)


vit9696:
AMD CPUs cannot work with Intel XCPM, I think you missed something.

Pavo-IM:
Not missing anything SSDT-PLUG works perfectly fine with AMD Zen and Zen 2 CPUs. This has been tested and working on all 17h Family AMD CPUs 1000x, 2000x and 3000x.
Screenshot 2020-04-03 19 15 47
If you need anymore proof just ask.

Download-Fritz:
XCPM is somehow nuked either way on AMD, yes...

Andrey1970AppleLife:
com.apple.driver.AppleIntelCPUPowerManagement - not used at xcpm

Download-Fritz:
@Andrey1970AppleLife That is likely because they currently fake Penryn, but it also works this way when faking Skylake. I think the AMD patches implicitly or explicitly disable XCPM, but it's not like there is literally any sort of documentation. Either, it works.

Docs/AcpiSamples/SSDT-PLUG.dsl
```diff
@@ -61,6 +63,36 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "CpuPlug", 0x00003000)
             }
         }
     }
+    
+    If (CondRefOf (\_PR.C000)) {
+        Scope (\_PR.P000) {
```
Download-Fritz, line 68: Incorrect CPU name

Pavo-IM: Ah ok I see what you mean, my bad will fix

vit9696:
Ugh, are you really sure this is needed? `plugin-type` makes use of advanced power management for Intel and results in loading of frequency vectors. I can merge this just in case since it is a sample anyway, and there can be Intel CPUs with such paths, but I really do not think AMD needs it.

vit9696:
Ok, after talking with @Download-Fritz I somewhat understood what is going on. To get AGPM loaded (and potentially working graphics power management) you need to do two things:

- CPU PM
- Platform PM

CPU PM can be one of the following: Null, AICPUPM, XCPM. Null does not work due to several macOS drivers checking for kext name, so either AICPUPM or XCPM. Neither of them can work with AMD, but the patch in OpenCore (`DummyPowerManagement`) makes AICPUPM behave the same way as Null CPU PM.

Platform PM can be one of the following: ACPI_SMC_PlatformPlugin or X86PlatformPlugin. Normally you have the former for AICPUPM and the latter for XCPM. However, on AMD you cannot have ACPI_SMC_PlatformPlugin as it does CPUID checks. Therefore the hack on AMD is to make X86PlatformPlugin load.

This is very ugly, but it makes sense to me. Merged.

Pavo-IM:
Thanks for the merge but I did want to clear something up. Yes it is true that AMD cannot use the advanced power management  for Intel by loading of frequency vectors but on my system which is a Ryzen 9 3950X. I do not use `DummyPowerManagement` Quirk and my runs better when I use SSDT-PLUG to inject the `plugin-type`. I can show you some benchmarks with and without SSDT-PLUG if you like. But thanks again for the merge.

aluveitie:
@Pavo-IM Would be interesting to see the difference

Pavo-IM:
With SSDT-PLUG enabled:
Screenshot 2020-04-04 09 33 37

Without SSDT-PLUG enabled:
Screenshot 2020-04-04 09 40 53

aluveitie:
Does disabling `DummyPowerManagement` make any difference performance wise?

Pavo-IM:
Those test were done without `DummyPowerManagement` enabled. I have tested it with the Quirk enabled and it doesn't make any difference in scores.

Download-Fritz:
As far as I know you can fake high scores by raising FSBFrequency (at least on some systems, Skylake+ does not use FSB)... especially on systems with garbage metrics (only God knows what Cinebench uses and what Intel code outputs for AMD), scores are more or less meaningless. Either way, out of scope of this PR.

PR #27, opened 2020-04-03
for HP Z620 dual cpus socket

ZISQO:
.

### DIFF
--- a/Docs/AcpiSamples/SSDT-PLUG.dsl
+++ b/Docs/AcpiSamples/SSDT-PLUG.dsl
@@ -16,6 +16,7 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "CpuPlug", 0x00003000)
     External (_PR_.PR00, DeviceObj)
     External (_SB_.SCK0.CP00, DeviceObj)
     External (_SB_.SCK0.PR00, DeviceObj)
+    External (_SB_.SCK1.PR00, DeviceObj)
 
     If (CondRefOf (\_SB.CPU0)) {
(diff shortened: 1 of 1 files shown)


vit9696:
That's Sandy Bridge, Sandy Bridge does not support XCPM. Adding plugin-type to one socket is also enough.

ZISQO:
> That's Sandy Bridge, Sandy Bridge does not support XCPM. Adding plugin-type to one socket is also enough.

also it's possible to do with legacy AICPUPM pathed file of.
already i was successed with Z420, 620 for Catalina 10.15.x and Sleep and wake up perfectly.

usr-sse2:
> That's Sandy Bridge, Sandy Bridge does not support XCPM. Adding plugin-type to one socket is also enough.
Plugin-type=1 is not only for XCPM. Ivy Bridge doesn't support XCPM, but requires plugin-type 1 so it loads X86PlatformPlugin.

vit9696:
Why do you load X86PlatformPlugin on Sandy Bridge, when it is supposed to use ACPI_SMC_PlatformPlugin?

BrushXue:
> > That's Sandy Bridge, Sandy Bridge does not support XCPM. Adding plugin-type to one socket is also enough.
> 
> also it's possible to do with legacy AICPUPM pathed file of.
> already i was successed with Z420, 620 for Catalina 10.15.x and Sleep and wake up perfectly OC opencore. so could you please say to quickly?

Could please share your EFI folder? I have dual E5 and want to switch to OC but didn't find any tutorial for X79 or C602.

PR #28, opened 2020-04-05
GUI improvements by @usr-sse2 scheduled for 0.5.7

vit9696:


crushers:
Can u add support for mouse?

Platform/OpenCanopy/OpenCanopy.c
```diff
@@ -1196,6 +1196,49 @@ GuiDrawLoop (
...
+     && ImageData[Index    ] == 'i'
+     && ImageData[Index + 1] == 'c'
+     && ImageData[Index + 2] == '0'
+     && ImageData[Index + 3] == '7') {
+      return GuiPngToImage(Image, IcnsImage + Index + 8, RecordLength - 8);
+    } else if (Scale == 2
+     && ImageData[Index    ] == 'i'
+     && ImageData[Index + 1] == 'c'
+     && ImageData[Index + 2] == '1'
+     && ImageData[Index + 3] == '3') {
```
Download-Fritz, line 1234: Maybe consider ReadUanaligned32()

Platform/OpenCanopy/OpenCanopy.c
```diff
@@ -1196,6 +1196,49 @@ GuiDrawLoop (
...
+  UINT32 RecordLength;
+  UINT8 *ImageData;
+
+  //
+  // We do not need to support 'it32' 128x128 icon format,
+  // because Finder automatically converts the icons to PNG-based
+  // when assigning volume icon
+  //
+
+  Index = 8;
```
Download-Fritz, line 1217: Magic constant

Platform/OpenCanopy/OpenCanopy.h
```diff
@@ -275,4 +284,8 @@ GuiPngToClickImage (
   IN     CONST EFI_GRAPHICS_OUTPUT_BLT_PIXEL  *HighlightPixel
   );
 
-#endif // BOOT_LIQUOR_H
+#define OPENCANOPY_USE_VOLUME_ICON         (1 << 10)
+#define OPENCANOPY_USE_DISK_LABEL_FILE     (1 << 9)
+#define OPENCANOPY_USE_GENERIC_LABEL_IMAGE (1 << 8)
```
Download-Fritz, line 289: Please use the BIT* macros

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -718,42 +726,180 @@ GLOBAL_REMOVE_IF_UNREFERENCED GUI_OBJ mBootPickerView = {
...
+
+  if (RawData == NULL || DataLength < 5) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Image->Width = RawData[1] << 8 | RawData[2];
+  Image->Height = RawData[3] << 8 | RawData[4];
+  Image->Buffer = (EFI_GRAPHICS_OUTPUT_BLT_PIXEL *) AllocatePool(Image->Width * Image->Height * sizeof(EFI_GRAPHICS_OUTPUT_BLT_PIXEL));
+
+  if (DataLength != 5 + Image->Width * Image->Height) {
```
Download-Fritz, line 767: Are Image->* fields validated here? I don't believe so. In this case we need overflow-aware arithmetics

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -718,42 +726,180 @@ GLOBAL_REMOVE_IF_UNREFERENCED GUI_OBJ mBootPickerView = {
...
   }
 
-  VolumeEntry->Context = EntryContext;
+  VolumeEntry->Context = Entry;
 
-  if (!IsExternal) {
+  if (UseVolumeIcon
+   && EFI_SUCCESS == OcGetBootEntryIcon(Context, AppleBootPolicy, Entry, &IconFileData, &IconFileSize)
+   && (EntryIcon = (GUI_IMAGE *) AllocatePool(sizeof(GUI_IMAGE)))
+   && EFI_SUCCESS == GuiIcnsToImage128x128(EntryIcon, IconFileData, IconFileSize, GuiContext->Scale)) {
```
Download-Fritz, line 898: if false, this will leak EntryIcon... avoid side-effects in if-conditions, especially allocations

Platform/OpenCanopy/OpenCanopy.c
```diff
@@ -1196,6 +1196,49 @@ GuiDrawLoop (
...
+  ImageData = (UINT8 *) IcnsImage;
+  while (Index < IcnsImageSize - 8) {
```
Download-Fritz, line 1219: Maybe this should be a for-loop?

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -995,6 +1191,15 @@ BootPickerViewInitialize (
   mBootPicker.Hdr.Obj.OffsetX = mBootPickerView.Width / 2;
   mBootPicker.Hdr.Obj.OffsetY = (mBootPickerView.Height - mBootPicker.Hdr.Obj.Height) / 2;
 
+  // TODO: animations should be tied to UI objects, not global
+  // Each object has its own list of animations.
```
Download-Fritz, line 1195: Logically yes, but for performance reasons, I think we should stick to a centralised list

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -718,42 +726,180 @@ GLOBAL_REMOVE_IF_UNREFERENCED GUI_OBJ mBootPickerView = {
...
+  OUT GUI_IMAGE *Image,
+  IN  UINT8     *RawData,
+  IN  UINT32    DataLength
+  )
+{
+  UINT32 PixelIdx;
+
+  if (RawData == NULL || DataLength < 5) {
+    return EFI_INVALID_PARAMETER;
+  }
```
Download-Fritz, line 761: We generally use ASSERT() to ensure function preconditions. DataLength < 5 may be a precondition, but then I'd say to remove the condition from the file reading function to have one single spot of validation. Also, magic constant.

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -718,42 +726,180 @@ GLOBAL_REMOVE_IF_UNREFERENCED GUI_OBJ mBootPickerView = {
...
+
+  Image->Width = RawData[1] << 8 | RawData[2];
+  Image->Height = RawData[3] << 8 | RawData[4];
```
Download-Fritz, line 764: Maybe use ReadUanligned16()?

Platform/OpenCanopy/GuiApp.c
```diff
@@ -38,22 +43,101 @@ InternalContextDestruct (
...
+    return RETURN_OUT_OF_RESOURCES;
+  }
+
+  UnicodeSPrint (Path, BufferSize, L"%s.%a", ImageFilePath, Scale == 2 ? "l2x" : "lbl");
+
+  *FileData = OcStorageReadFileUnicode (Storage, Path, FileSize);
+
+  if (*FileData == NULL || *FileSize == 0) {
+    DEBUG ((DEBUG_WARN, "Failed to load %s\n", Path));
+    return RETURN_NOT_FOUND;
```
Download-Fritz, line 129: This leaks path and possibly *FileData

Library/OcBootManagementLib/BootEntryManagement.c
```diff
@@ -131,6 +132,165 @@ OcDescribeBootEntry (
...
+
+  if (BootEntry->Type == OC_BOOT_EXTERNAL_TOOL || BootEntry->Type == OC_BOOT_SYSTEM) {
+    if (Context->CustomDescribe == NULL) {
+      DEBUG((DEBUG_ERROR, "CustomDescribe is null\n"));
+      return EFI_NOT_FOUND;
+    }
+    Status = Context->CustomDescribe(
+      Context->CustomEntryContext,
+      BootEntry,
+      (UINT8)Scale,
```
Download-Fritz, line 173: I think you can safely declare Scale as UINT8 in the first place

Platform/OpenCore/OpenCoreMisc.c
```diff
@@ -159,6 +159,59 @@ OcToolDescribeEntry (
   HasIcon  = FALSE;
   HasLabel = FALSE;
 
+  if (ChosenEntry->Type == OC_BOOT_SYSTEM) {
+
+    if (StrCmp(ChosenEntry->Name, OC_MENU_RESET_NVRAM_ENTRY) == 0) {
```
Download-Fritz, line 164: Hmm, maybe we should have a secondary identifier? I really do not like the idea of using name for this.
cc @vit9696

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -718,42 +726,180 @@ GLOBAL_REMOVE_IF_UNREFERENCED GUI_OBJ mBootPickerView = {
...
+      return RETURN_UNSUPPORTED;
+    }
   }
 
-  VolumeEntry->Context = EntryContext;
+  VolumeEntry->Context = Entry;
 
-  if (!IsExternal) {
+  if (UseVolumeIcon
+   && EFI_SUCCESS == OcGetBootEntryIcon(Context, AppleBootPolicy, Entry, &IconFileData, &IconFileSize)
```
Download-Fritz, line 896: Where is IconFileData supposed to be freed?

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -718,42 +726,180 @@ GLOBAL_REMOVE_IF_UNREFERENCED GUI_OBJ mBootPickerView = {
...
-    return RETURN_UNSUPPORTED;
+  AppleBootPolicy = OcAppleBootPolicyInstallProtocol (FALSE);
+  if (AppleBootPolicy == NULL) {
+    DEBUG ((DEBUG_ERROR, "OCB: AppleBootPolicy locate failure\n"));
+    return EFI_NOT_FOUND;
+  }
+
+  if (UseDiskLabel
+   && EFI_SUCCESS == OcGetBootEntryLabelImage(Context, AppleBootPolicy, Entry, 1, &IconFileData, &IconFileSize)
+   && EFI_SUCCESS == DecodeAppleDiskLabelImage(&VolumeEntry->Label, IconFileData, IconFileSize)) {
```
Download-Fritz, line 840: please stick to EFI_ERROR()

Platform/OpenCanopy/Views/BootPicker.c
Download-Fritz: please stick to EFI_ERROR()

Library/OcBootManagementLib/BootEntryManagement.c
```diff
@@ -131,6 +132,165 @@ OcDescribeBootEntry (
...
+      *DataLength = 0;
+    }
+    DEBUG((DEBUG_INFO, "Get label for external tool %s - %r\n", BootEntry->Name, Status));
+    return Status;
+  }
+
+  if (BootEntry->DevicePath == NULL) {
+    DEBUG((DEBUG_ERROR, "DevicePath for entry %s is NULL\n", BootEntry->Name));
+    return EFI_UNSUPPORTED;
+  }
```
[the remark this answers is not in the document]
vit9696: Think not, could probably just assert.

Platform/OpenCanopy/Views/BootPicker.c
vit9696: They are UINT16 if I remember correctly, so they cannot overflow (since they will automatically promote to int).

Platform/OpenCanopy/Views/BootPicker.c
vit9696: There is a byte swap here, unsure whether it is more readable the other way.

vit9696: Most likely we want a structure really.

Platform/OpenCore/OpenCoreMisc.c
vit9696: Yes, that's what I asked @usr-sse2 to do for now when he asked me that it is not right. Maybe indeed turn system into reset nvram and then add another type when we need?

Download-Fritz: Depends, how many System entries could we possibly have? Either this or a secondary type

Platform/OpenCanopy/Views/BootPicker.c
usr-sse2: Hm, need to free it right here, but also in case of decode failure. Will extract it into a function so the if-else if-else control flow is still applicable

Platform/OpenCanopy/OpenCanopy.c
```diff
@@ -1356,12 +1493,11 @@ GuiGetInterpolatedValue (
   UINT32 DeltaTime;
 
   ASSERT (Interpol != NULL);
-  ASSERT (Interpol->StartTime <= CurrentTime);
```
vit9696, line 1359: Fritz, is this ok to remove this?

Platform/OpenCanopy/OpenCanopy.c
```diff
@@ -1356,12 +1493,11 @@ GuiGetInterpolatedValue (
...
   ASSERT (Interpol->Duration > 0);
 
   STATIC CONST UINT32 InterpolFpTimeFactor = 1U << 12U;
 
-  if (CurrentTime == Interpol->StartTime) {
+  if (CurrentTime <= Interpol->StartTime) {
```
vit9696, line 1500: @Download-Fritz, is this correct?

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -413,7 +418,11 @@ InternalBootPickerEntryDraw (
   ASSERT (Context != NULL);
 
   Entry       = BASE_CR (This, GUI_VOLUME_ENTRY, Hdr.Obj);
-  EntryIcon   = Entry->EntryIcon;
+  /*if (mBootPickerImageIndex < 5) {
+    EntryIcon = &((BOOT_PICKER_GUI_CONTEXT *) DrawContext->GuiContext)->Poof[mBootPickerImageIndex];
+  } else */{
+    EntryIcon   = Entry->EntryIcon;
+  }
```
vit9696, line 425: Does the commented out code above make sense?

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -441,8 +450,8 @@ InternalBootPickerEntryDraw (
   //
   // Draw the label horizontally centered.
   //
-  ASSERT (Label->Width  <= BOOT_ENTRY_DIMENSION);
-  ASSERT (Label->Height == BOOT_ENTRY_LABEL_HEIGHT);
+  //ASSERT (Label->Width  <= BOOT_ENTRY_DIMENSION);
+  //ASSERT (Label->Height == BOOT_ENTRY_LABEL_HEIGHT);
```
vit9696, line 454: Is this correct?

usr-sse2: Height: macOS makes height 12, built-in renderer makes height 13. And as we are reading it from file, it may contain any height if future macOS decides to change it.

usr-sse2: Width: as we don't support scrolling labels now, do we want to be unable to boot in DEBUG because of this assert?

Platform/OpenCanopy/Views/BootPicker.c
usr-sse2: It's for testing Poof animation.

Platform/OpenCanopy/OpenCanopy.c
usr-sse2: It's to support sequential animations: one animation starts after the end of the previous one, so StartTime can be not 0.

Platform/OpenCanopy/OpenCanopy.c
usr-sse2: It's to support sequential animations: one animation starts after the end of the previous one, so StartTime can be less than CurrentTime.

usr-sse2:
> Can u add support for mouse?

Mouse is already supported. It works.

Platform/OpenCanopy/Views/BootPicker.c
Download-Fritz: I think it should at least be sanitised, but sticking to what Apple has now sounds good to me.

crushers:
> > Can u add support for mouse?
> 
> Mouse is already supported. It works.

Sorry but for me does not work.

telepati:
@crushers what kind of mouse are you using? Because mouses (Apple Magic Mouses) always worked since @Download-Fritz merge the OpenCanopy to OpenCore. Response time little slow but it always worked.

crushers:
> @crushers what kind of mouse are you using? Because mouses (Apple Magic Mouses) always worked since @Download-Fritz merge the OpenCanopy to OpenCore. Response time little slow but it always worked.

I use classic USB mouse.

usr-sse2:
@crushers Do you have USB mouse UEFI driver installed or built-in in the firmware? Does the mouse work in FileVault password screen or in Clover?

usr-sse2:
@telepati Wow, Bluetooth mouse can work in UEFI?! What Bluetooth dongle are you using?

usr-sse2:
@crushers @telepati 
Moved the discussion of mouse support to https://github.com/acidanthera/bugtracker/issues/866

PR #35, opened 2020-04-12
Added a check to fix 'Cannot override' bug #814

MacOS86-it:
This one should fix issue #814 once and for all. The issue was caused by the SSDT that was adding _DSM Method to a Package (forbidden by ACPI Specification) defined on the DSDT (\_SB.PR00). The new if-condition checks if the object is a Processor. By doing this way, if it finds a Package it doesn't add the _DSM Method.

### DIFF
--- a/Docs/AcpiSamples/SSDT-PLUG.dsl
+++ b/Docs/AcpiSamples/SSDT-PLUG.dsl
@@ -32,73 +32,89 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "CpuPlug", 0x00003000)
     }
 
     If (CondRefOf (\_SB.CPU0)) {
-        Scope (\_SB.CPU0) {
-            Method (_DSM, 4, NotSerialized)  
-            {
-                Return (PMPM (Arg0, Arg1, Arg2, Arg3))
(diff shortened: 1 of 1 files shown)


vit9696:
But in this case the SSDT will just not work, is that desired?

MacOS86-it:
In the case of the issue two of the If CondRefOf... conditions are true: both the processor (_PR.CPU0) and a package (PR00).
ObjectType(_PR.CPU0) is equal to 12 as per ACPI Specification. The condition is true, so the SSDT adds the plugin-type property.
In the case of _SB.PR00, ObjectType returns 4, since PR00 is a Package. The condition is false, so the _DSM method is skipped.
The error was caused because the table added a _DSM Method to a package.
With this fix the SSDT still adds the DSM property for the processor, but it skips the _DSM Method for Package.
In conclusion, in this case the SSDT will work as intended (it adds plugin-type only to the processor).

0ECCC882-F8A2-4883-B715-589FAF659589
5EEB6C49-F133-443B-954A-561A313FCDF8

vit9696:
Whoever wrote this DSDT is mad, but the change makes sense to me now. Thanks, merged.

PR #36, opened 2020-04-22
Gui timeout

TaiPhamD:
This change allows the OpenCanopy GUI to auto select the default boot entry after config.plist->MISC->Boot->Timeout expires. 

FYI - I had to move structs declare from BootPicker.c into BootPicker.h so that we can reference those struct from OpenCanopy.c


vit9696:
There are several codestyle issues with this change, but as for the logic I think for the time being we can live with this. I will most likely slightly refactor it myself and will merge in the coming days. Thanks.

PR #39, opened 2020-04-26
OpenCanopy: Support HiDPI in fontless mode

usr-sse2:


Platform/OpenCanopy/GuiApp.c
```diff
@@ -154,8 +154,8 @@ LoadImageFileFromStorage (
           FileData,
           FileSize,
           Scale,
-          MatchWidth,
-          MatchHeight
+          0, //MatchWidth * Scale,
```
vit9696, line 157: This is a security check, we cannot replace it with 0.

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -1176,15 +1177,19 @@ BootPickerViewInitialize (
     InternalBootPickerExitLoop,
     GuiContext
     );
+  DrawContext->Scale = GuiContext->Scale;
```
vit9696, line 1180: Was it not set before?

usr-sse2: There are no other assignments to this variable

Platform/OpenCanopy/GuiApp.c
usr-sse2: - I found that MatchWidth is multiplied by Scale several times, I'll fix it;
- Current cursor image is 24x24 for 1x scale and 80x56 for 2x scale, both old and new, though it should be possible to crop it to 48x48 or extend both so they become 48x48 and 96x96.

I don't understand why:
- it's a *security* issue;
- it's an issue at all, because nothing in the layout is tied to the size of a cursor, so if a user wants a 256x128 cursor – he just puts it in Cursor.icns and nothing breaks.

Platform/OpenCanopy/Views/BootPicker.c
usr-sse2: Yes, there are no other assignments.

Platform/OpenCanopy/GuiApp.c
vit9696: It is a security issue, because providing a very large (or very small) image may make the interface unusable and/or unstable.

vit9696: As for cursor, I think the current code in boot.efi expects 80x56, so we probably want just ensure the right dimensions for 2x scale.

usr-sse2: 1) does it expect 80x56 for HiDPI and 24x24 for non-HiDPI as we currently have?
2) boot.efi uses its own cursor, not our?

usr-sse2: If the cursor is at least 1x1, the UI is usable, as you can see in VNC apps. If the cursor is 24x24 but fully transparent, the UI is unusable *by mouse*, but everything is still accessible from the keyboard. If the cursor is a 1920x1080 screenshot of OpenCore UI, it would still be usable by mouse, though the frame rate would be low. Though it may become not usable by keyboard, because until the user moves mouse, he will not understand that something is wrong. So, to protect from such attack it's sufficient to limit maximum cursor size, and strict size requirements don't do anything useful.

vit9696: 1. boot.efi does indeed have a custom cursor loaded from FirmwareVolume protocol. Sorry, I forgot about it. But I am pretty sure our cursors come from the firmware and they match with boot.efi cursors.

2. You are right that with the cursor we want to limit cursor size not restrict it to a particular size. I was probably unnecessarily harsh about it. Could you maybe add a boolean flag `StrictDimensions` or anything similar to let `<=` size comparison for some images?

usr-sse2: efa6c42cce08724712f1b3f0a8f1cca2c93cbee0

PR #40, opened 2020-05-03
Updated Discussion Links

Pavo-IM:
Added Proxmox VE 6.1 configuration.


vit9696:
Thanks, but I am not including it. That one is AMD specific. We do not target to linking to every guide or similar. Leoyzen's configuration is there mainly because he is the one responsible for adding OC QEMU/KVM support and is somewhat in charge in it.

Pavo-IM:
Well it would probably show the community that your product also supports AMD systems, but oh well.

PR #41, opened 2020-05-03
Address Misc Macbuild Tool Issues

dakanji:
Addresses a few issues flagged by ShellCheck in macbuild.tool.

ShellCheck Refs:
https://github.com/koalaman/shellcheck/wiki/SC2155
https://github.com/koalaman/shellcheck/wiki/SC2164
https://github.com/koalaman/shellcheck/wiki/SC2046
https://github.com/koalaman/shellcheck/wiki/SC2034

### DIFF
--- a/macbuild.tool
+++ b/macbuild.tool
@@ -69,7 +69,7 @@ package() {
   popd || exit 1
 }
 
-cd "$(dirname $0)" || exit 1
+cd dirname "$0" || exit 1
 ARCHS=(X64 IA32)
 SELFPKG=OpenCorePkg
 DEPNAMES=('EfiPkg' 'MacInfoPkg' 'DuetPkg')
--- a/macbuild.tool
+++ b/macbuild.tool
@@ -69,7 +69,7 @@ package() {
   popd || exit 1
 }
 
-cd dirname "$0" || exit 1
+cd $(dirname "$0") || exit 1
 ARCHS=(X64 IA32)
 SELFPKG=OpenCorePkg
 DEPNAMES=('EfiPkg' 'MacInfoPkg' 'DuetPkg')
--- a/macbuild.tool
+++ b/macbuild.tool
@@ -79,6 +79,13 @@ DEPURLS=(
   'https://github.com/acidanthera/DuetPkg'
 )
 DEPBRANCHES=('master' 'master' 'master')
+
+export ARCHS
+export SELFPKG
+export DEPNAMES
+export DEPURLS
+export DEPBRANCHES
+
 src=$(/usr/bin/curl -Lfs https://raw.githubusercontent.com/acidanthera/ocbuild/master/efibuild.sh) && eval "$src" || exit 1
 
 if [ "$BUILD_UTILITIES" = "1" ]; then
--- a/macbuild.tool
+++ b/macbuild.tool
@@ -69,7 +69,7 @@ package() {
   popd || exit 1
 }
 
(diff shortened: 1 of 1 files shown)


macbuild.tool
```diff
@@ -6,7 +6,8 @@ package() {
     exit 1
   fi
 
-  local ver=$(cat Include/OpenCore.h | grep OPEN_CORE_VERSION | sed 's/.*"\(.*\)".*/\1/' | grep -E '^[0-9.]+$')
+  local ver
+  ver=$(cat Include/OpenCore.h | grep OPEN_CORE_VERSION | sed 's/.*"\(.*\)".*/\1/' | grep -E '^[0-9.]+$')
```
vit9696, line 10: Why is that needed?

macbuild.tool
```diff
@@ -68,7 +69,7 @@ package() {
   popd || exit 1
 }
 
-cd $(dirname "$0")
+cd "$(dirname $0)" || exit 1
```
vit9696, line 72: This may not work if `$0` contains spaces as far as I know.

macbuild.tool
```diff
@@ -78,6 +79,13 @@ DEPURLS=(
   'https://github.com/acidanthera/DuetPkg'
 )
 DEPBRANCHES=('master' 'master' 'master')
+
+export ARCHS
+export SELFPKG
+export DEPNAMES
+export DEPURLS
+export DEPBRANCHES
```
vit9696, line 87: Why is that needed?

macbuild.tool
dakanji: macbuild.tool works perfectly well as it is ... run it dozens of times in last week.

So change not "Needed" ... just desirable in terms of maintaining standards.
I.E., no issue here but there is a potential issue that may happen elsewhere in development if such local variables are declared and assigned together.

See: https://github.com/koalaman/shellcheck/wiki/SC2155#problematic-code-in-the-case-of-local

macbuild.tool
dakanji: Indeed.

On review, issue is actually that command expansion is not needed as not assigning output.

Will modify to `cd dirname "$0" || exit 1`

macbuild.tool
dakanji: Basically same as local variables item.

The declared variables are not actually used anywhere in this script file but are for efibuild.sh which is run in a subshell.

While things work as it, technically should be exported in such instances. Again, something that is not an issue here but potentially elsewhere if not following such by default

vit9696: Hmm, does the tool unconditionally warn about these variables? If we cannot silence the warnings, I am fine to merge this, but in this case I request all the scripts to be checked with this tool on CI level, so that we do not violate these recommendations anywhere else in the future.

Could you update your PR and include such a check in `.travis.yml`?

dakanji: ShellCheck raises the error messages and warnings by default if that's what you mean.

In my case it is installed as a plugin in my editor (Atom).

I believe TravisCI comes with it by default but will need to read up on how to add.

I will squash the commits into one clean one and push. If I can figure the Travis thing out,  I will add it.

Otherwise will add an issue and look into dealing with it separately. 

PS. ShellCheck messages can be silenced by adding a line telling it to ignore a condition above where that condition appears.

This is needed as there can be reasons why something is done but you will obviously see the addition when reviewing.

vit9696:
Ok, I merged this.

PR #43, opened 2020-05-04
Fonts

usr-sse2:
- Fix BMF font loading (how could it work before?!)
- Support 2x font
- Display both rendered and stored labels with alpha on non-black background
- Support fonts with negative offsets (blend pixels instead of overwriting)

Fonts are in acidanthera/OcBinaryData#3


Platform/OpenCanopy/BitmapFont.c
```diff
@@ -84,6 +84,9 @@ BmfGetKerningPair (
   ASSERT (Context != NULL);
 
   Pairs = Context->KerningPairs;
+
+  if (Pairs == NULL)
+    return NULL;
```
vit9696, line 89: Should this be in curly braces?

Platform/OpenCanopy/BitmapFont.c
```diff
@@ -301,13 +304,23 @@ BmfContextInitialize (
...
   }
 
-  if (Context->Info         == NULL
-   || Context->Common       == NULL
-   || Context->Pages        == NULL
-   || Context->Chars        == NULL
-   || Context->KerningPairs == NULL
-    ) {
-    DEBUG ((DEBUG_WARN, "BMF: Missing block\n"));
+  if (Context->Info         == NULL) {
```
vit9696, line 307: Maybe remove the extra indentations now the if is different?

Platform/OpenCanopy/BitmapFont.c
```diff
@@ -398,68 +411,71 @@ BmfContextInitialize (
...
+        if (Pairs[Index - 1].first > Pairs[Index].first) {
+          DEBUG ((DEBUG_WARN, "BMF: First Character IDs are not sorted\n"));
           return FALSE;
         }
+
+        if (Pairs[Index - 1].first == Pairs[Index].first) {
+          if (Pairs[Index - 1].second > Pairs[Index].second) {
+            DEBUG ((DEBUG_WARN, "BMF: Second Character IDs are not sorted\n"));
+            return FALSE;
+          } else if (Pairs[Index - 1].second == Pairs[Index].second) {
```
vit9696, line 471: Probably else is not needed here for better readability.

Platform/OpenCanopy/BitmapFont.c
```diff
@@ -693,6 +740,13 @@ GuiGetLabel (
     InitialWidthOffset = 0;
   }
 
+  // Apply palette because Apple does it with stored labels
```
vit9696, line 743: Please use surrounding `//` for upper and lower lines for comments.

Platform/OpenCanopy/BitmapFont.c
```diff
@@ -586,26 +602,55 @@ BmfGetTextInfo (
...
+    // We assume that the font is generated by dpFontBaker
+    // and has only gray channel, which should be interpreted as alpha.
+    GuiBlendPixel(Dst, Color, AlphaSrc->Red);
     ++Dst;
-    ++Src;
+    ++AlphaSrc;
   }
 }
 
+STATIC CONST UINT8 clut[] = {
```
vit9696, line 633: Please name global module variables with `m` prefix (e.g. `mClut`).

Platform/OpenCanopy/BitmapFont.c
```diff
@@ -586,26 +602,55 @@ BmfGetTextInfo (
...
   )
 {
   UINTN  Index;
 
   for (Index = 0; Index < PixelCount; ++Index) {
-    Dst->Red       = 255 - Src->Red;
-    Dst->Green     = 255 - Src->Green;
-    Dst->Blue      = 255 - Src->Blue;
-    Dst->Reserved  = Src->Reserved;
+    // We assume that the font is generated by dpFontBaker
```
vit9696, line 625: Please use surrounding `//` for upper and lower lines for comments.

Platform/OpenCanopy/BitmapFont.c
```diff
@@ -586,26 +602,55 @@ BmfGetTextInfo (
...
+  0x80, /* 0x99 = (3*43) = 129 colors*/
+  0xFC, /* 0xAA */
+  0xFD, /* 0xBB */
+  0xAB, /* 0xCC = 4*43 = 172 colors */
+  0xFE, /* 0xDD */
+  0xFF, /* 0xEE */
+  0xD6, /* 0xFF = 5*43 = 215 */
+};
+
+extern CONST UINT8 mAppleDiskLabelImagePalette[256];
```
vit9696, line 652: I think that needs to be in the header.

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -418,8 +418,8 @@ InternalBootPickerEntryDraw (
   //
   // Draw the label horizontally centered.
   //
-  //ASSERT (Label->Width  <= BOOT_ENTRY_DIMENSION);
-  //ASSERT (Label->Height == BOOT_ENTRY_LABEL_HEIGHT);
+  //ASSERT (Label->Width  <= BOOT_ENTRY_DIMENSION * DrawContext->Scale);
```
vit9696, line 421: I believe we want a FIXME explaining why we cannot enable it (the limit is 340, but we cannot properly draw more).

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -418,8 +418,8 @@ InternalBootPickerEntryDraw (
...
+  ASSERT_EQUALS (Label->Height, BOOT_ENTRY_LABEL_HEIGHT * DrawContext->Scale);
```
vit9696, line 422: I think this wants to become <= following the `APPLE_DISK_LABEL_MAX_WIDTH` check during loading.

Also, can we have static assertions for these dimensions being <= `APPLE_DISK_LABEL_MAX_WIDTH` and `APPLE_DISK_LABEL_MAX_HEIGHT`.

Platform/OpenCanopy/GuiApp.c
```diff
@@ -404,7 +409,9 @@ InternalContextConstruct (
       FontData,
       FontDataSize
       );
-    if (Context->FontContext.BmfContext.Height  != BOOT_ENTRY_LABEL_HEIGHT) {
+    if (Context->FontContext.BmfContext.Height != BOOT_ENTRY_LABEL_HEIGHT * Context->Scale) {
+        DEBUG((DEBUG_WARN, "OCUI: Font has height %d instead of %d",
```
vit9696, line 413: This looks a bit unindented, please use:

```C
DEBUG ((
  DEBUG_WARN,
  "OCUI: …",
  …
  );
```

Platform/OpenCanopy/BitmapFont.c
usr-sse2: This is not my change, I just moved it inside if

Platform/OpenCanopy/BitmapFont.c
usr-sse2: I wanted to discuss if we need this code or not. You say that Apple has pngs in addition to the label format. Are they already used or it's for the future?

Platform/OpenCanopy/BitmapFont.c
usr-sse2: And how it should be named, if it's no more static?

Platform/OpenCanopy/BitmapFont.c
vit9696: Yeah, but still better to use `if (x) { return; } if (y) { return; }`. Probably an overlook from the older review or even initial merge.

Platform/OpenCanopy/BitmapFont.c
vit9696: - They do not have PNG images ondisk at least presently.
- Disk labels can be in two formats, with the second being BGRA format, which support I recently added to disklabel utility.
- Apple uses PNG or BMP in the firmware for prebuilt labels (e.g. `EFI Boot`), but their protocol accepts any format, including `icns` and disk label (either type).
- Ondisk disk labels are processed with the same protocol, so it can technically process them in any format, but `icns` will be likely unsupported.
- Presently I do not know of any ondisk labels using anything but type 1 (alpha-only 8-bit format).

I am not sure we need to bother to support all other formats to be honest, maybe with the time of if we find them being used in the wild.

Platform/OpenCanopy/BitmapFont.c
vit9696: Normally with `g` prefix (`gAppleDiskLabelImagePalette`).  Maybe `gAppleDiskLabelPalette` and `gAppleDiskLabelClut`?

Platform/OpenCanopy/BitmapFont.c
usr-sse2: Maybe we should implement some option in OcPngLib to allow loading gray-only images as alpha-only, so it would set its `Reserved` component instead of `Blue`, `Green`, `Red`?

Platform/OpenCanopy/BitmapFont.c
usr-sse2: Clut now has no reason to be global

Platform/OpenCanopy/BitmapFont.c
vit9696: I do not mind, but I am not sure how to do it best. What would you like to change it to in more detail?

Platform/OpenCanopy/BitmapFont.c
vit9696: GuiApp.h maybe?

Platform/OpenCanopy/Views/BootPicker.c
usr-sse2: <= or >=? 

We can also test (on Mac) if Apple enforces these requirements, e. g. what if we generate a  64x64 label.

vit9696: That height does not exceed 12. Testing on a Mac is a good idea, though I would say we should be better than them in terms of restrictions.

Platform/OpenCanopy/GuiApp.c
```diff
@@ -410,8 +410,12 @@ InternalContextConstruct (
       FontDataSize
       );
     if (Context->FontContext.BmfContext.Height != BOOT_ENTRY_LABEL_HEIGHT * Context->Scale) {
-        DEBUG((DEBUG_WARN, "OCUI: Font has height %d instead of %d",
-               Context->FontContext.BmfContext.Height, BOOT_ENTRY_LABEL_HEIGHT * Context->Scale));
+        DEBUG((
+          DEBUG_WARN,
+          "OCUI: Font has height %d instead of %d",
```
vit9696, line 415: Hmm, just realised that newline is missing.

Platform/OpenCanopy/Views/BootPicker.c
Download-Fritz: Right. Regarding size,
1) we can probably adapt offsets and such based of a Mac BootPicker screenshot, mine were just best guesses as I had no such available at the time.
2) Apple BootPicker implements label scrolling, I believe we want something like this too. A primitive version would just have a container GUI_OBJ with our maximum dimensions, and the actual label would be inside as wide as it wants to be. The inner object can then be slid freely to achieve the scrolling effect.

Download-Fritz:
Except for what is already noted, looks great to me, thanks a lot!

Platform/OpenCanopy/BitmapFont.c
usr-sse2: Looks more like OpenCanopy.h, because it's used in OpenCanopy.c and BitmapFont.c, and BitmapFont.c includes OpenCanopy.h and doesn't include GuiApp.h.

Platform/OpenCanopy/BitmapFont.c
usr-sse2: No, let's leave as is, because it would involve changes to lodepng, not our code.

Platform/OpenCanopy/BitmapFont.c
usr-sse2: The current question is: do we want to post-process our rendered labels (decrease their color depth by mapping them via clut and inverse clut), because that's what happens when generating and decoding .disklabel files, or just display what we rendered, supposing that Apple doesn't do it with builtin labels? Is "EFI Boot" saved after such processing or it's raw CoreText-rendered image?

Platform/OpenCanopy/BitmapFont.c
vit9696: Well, I showed you the decoder function from Apple firmware. There is no clut there, they just map 16 values to opacity levels with a step of 0x11. I believe we should follow.

Library/OcAppleDiskImageLib/OcAppleDiskImageLibInternal.c
```diff
@@ -239,9 +239,9 @@ InternalParsePlist (
     goto DONE_ERROR;
   }
 
-  Result = OcOverflowMulU32 (NumDmgBlocks, sizeof (*DmgBlocks), &DmgBlocksSize);
-  if (Result) {
-    return FALSE;
+  Result = !OcOverflowMulU32 (NumDmgBlocks, sizeof (*DmgBlocks), &DmgBlocksSize);
+  if (!Result) {
```
vit9696, line 243: Ugh, double negation? Please do not do it. This is very hard to read for no reason.

Platform/OpenCanopy/BitmapFont.c
```diff
@@ -398,68 +413,72 @@ BmfContextInitialize (
...
+                  Pairs[Index].amount,
+                  &Advance
+                  );
+      if (Result
+       || 0 > Width || Width > MAX_UINT16
+       || 0 > Advance || Advance > MAX_UINT16) {
+         DEBUG ((
+           DEBUG_WARN,
+           "BMF: Pair at index %d insane\n"
+           " first %u\n"
```
vit9696, line 448: Please remove newlines here actually as we do not support multiline logging. Just spaces will be fine.

Platform/OpenCanopy/BitmapFont.c
vit9696: Yeah, agree.

Library/OcAppleDiskImageLib/OcAppleDiskImageLibInternal.c
usr-sse2: The `Result` value is checked at DONE_ERROR, and it should be FALSE in case of error.

Platform/OpenCanopy/BitmapFont.c
usr-sse2: Not my change again (I just added index) but I'll fix.

Library/OcAppleDiskImageLib/OcAppleDiskImageLibInternal.c
vit9696: Ok, but maybe make it

```
Result = Oc…
if (Result) {
  Result = FALSE; ///< Result should be false in case of error.
  goto DONE_ERROR;
}
```

I also found another issue with this code:

```
    if (!Result || (BlockDictChildDataSize < sizeof (*Block))) {
      goto DONE_ERROR;
    }
```

Platform/OpenCanopy/BitmapFont.c
vit9696: Thanks!

Library/OcAppleDiskImageLib/OcAppleDiskImageLibInternal.c
usr-sse2: But then Result will be FALSE both in case of error and no error. I think that a variable should have consistent semantics across the whole function.

PR #45, opened 2020-05-04
Build: Fix ShellCheck warning in macbuild

dakanji:
Fix for final outstanding warning. Passes ShellCheck.
Tested and builds without error locally ... Should pass TravisCI

### DIFF
--- a/macbuild.tool
+++ b/macbuild.tool
@@ -69,7 +69,7 @@ package() {
   popd || exit 1
 }
 
-cd $(dirname "$0") || exit 1
+cd "$(dirname "$0")" || exit 1
 ARCHS=(X64 IA32)
 SELFPKG=OpenCorePkg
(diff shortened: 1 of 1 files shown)


vit9696:
Hmm, are you sure it is POSIX compliant? Well, we do not care about POSIX, but at least whether it works everywhere.

dakanji:
Seems to be. Changing shebang to `#!/bin/sh` results in 11 POSIX related warnings in ShellCheck from zero with `#!/bin/bash` ... this is not one of them.

POSIX_ERRS

vit9696:
Well, did you you actually check the standard or at least ran with different shells? Or is that blind relying on the instrument?

dakanji:
I did not check the standards and do blindly rely on it here.

In any case... 
- The file is clearly not POSIX compliant as you say
- The item clearly builds cleanly 
- Shebang is present and any interpreter would use this ... what exactly needs to be checked again?

PR #46, opened 2020-05-05
Fix Doxygen comments

usr-sse2:


Library/OcAudioLib/OcAudio.c
```diff
@@ -128,7 +128,7 @@ InternalMatchCodecDevicePath (
         &gEfiAudioIoProtocolGuid,
         (VOID **) &Private->AudioIo
         );
-      return EFI_SUCCESS;
+      return Status;
```
vit9696, line 131: This change looks good to me, but I guess it needs to be mentioned in the commit message at least. Marking not to forget.

vit9696:
Can we actually generate PDF or alike documentation from the code with Doxygen?

PR #47, opened 2020-05-09
Attempt to Fix acidanthera/bugtracker/issues/902

dakanji:
An attempt to fix acidanthera/bugtracker/#902 by running ShellCheck after Macbuild when all the various repos have been pulled in. It then checks each *.sh and *.tool file.

Alternative (better?) would be to update travis.yml for each repo. Perhaps even better is to check them here but also check in the specific repos

**Note:** There will be failures on current setup:
ShellCheck_1

ShellCheck_2

ShellCheck_3

Some are just warning that will need to be specifically disabled. 
Basically needs commitment to embrace ShellCheck

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,14 +8,23 @@ env:
 matrix:
   include:
   - os: osx
+    name: "Check Shell Scripts"
+    osx_image: xcode10.2
+    compiler: clang
+
+    script:
+      - HOMEBREW_NO_AUTO_UPDATE=1 brew install shellcheck
+      - "./macbuild.tool --skip-tests --skip-build --skip-package"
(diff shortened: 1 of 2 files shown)


dakanji:
Fails as expected due to ShellCheck Errors so working. 
NB: Means every commit will fail until these are errors are fixed if merged.

dakanji:
Messed up. Will redo

vit9696:
Thanks!

1. Do not call `macbuild.tool`, in my opinion. This way EDK II will not be cloned, and thus you will not have to add more exceptions than necessary. We do not control EDK II scripts and will not change them.

2. Also make sure to include the necessary changes to the scripts as I will not merge changes that break compilation.

PR #48, opened 2020-05-09
Add ShellCheck to OpenCorePkg CI

dakanji:
Closes acidanthera/bugtracker/#902

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,19 @@ env:
 
 matrix:
   include:
+ - os: linux
+    name: "Shellcheck on Linux"
+    addons:
+      apt:
+        packages:
+        - shellcheck
+
+    script:
+      # Check Parent Dir (Change "*.tool" to "*.{sh,tool}" as required if needed later)
+      - shellcheck ./*.tool
+      # Check Sub Dirs (Change "*.tool" to "*.{sh,tool}" as required if needed later)
+      - shellcheck ./**/*.tool
+
   - os: osx
     name: "Build macOS XCODE5"
     osx_image: xcode10.2
@@ -77,7 +90,7 @@ matrix:
     name: "Analyze Coverity"
     osx_image: xcode11.3
     compiler: clang
-    
+
     before_install:
       - HOMEBREW_NO_AUTO_UPDATE=1 brew install openssl
       - curl -Ls https://entrust.com/root-certificates/entrust_l1k.cer -o ~/entrust_l1k.crt || exit 1
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,2 @@
+# Ignore sourced files if any
+disable=SC1091
--- a/Debug/efidebug.tool
+++ b/Debug/efidebug.tool
@@ -26,23 +26,23 @@
 #
 
 RUNDIR=$(dirname "$0")
-pushd "${RUNDIR}" >/dev/null
+pushd "${RUNDIR}" >/dev/null  || exit 1
 RUNDIR=$(pwd)
-popd >/dev/null
+popd >/dev/null || exit 1
 
-cd "$RUNDIR"
+cd "$RUNDIR" || exit 1
 
 find_gdb() {
   if [ "${GDB}" = "" ]; then
-    GDB=$(which ggdb)
+    GDB=$(command -v ggdb)
   fi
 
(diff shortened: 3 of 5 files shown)


.travis.yml
```diff
@@ -7,6 +7,18 @@ env:
...
   include:
+  - os: osx
+    name: "Check Shell Scripts"
+    osx_image: xcode10.2
+    compiler: clang
+
+    script:
+      - HOMEBREW_NO_AUTO_UPDATE=1 brew install shellcheck
+      # Check Parent Dir (No *.sh files curently. Change "*.tool" to "*.{sh,tool}" if any is added later)
+      - shellcheck /home/travis/build/acidanthera/OpenCorePkg/*.tool
```
vit9696, line 18: Use `$(pwd)` if you want full path. We should not rely on travis internal paths.

.travis.yml
```diff
@@ -7,6 +7,18 @@ env:
...
+      # Check Sub Dirs (No *.sh files curently. Change "*.tool" to "*.{sh,tool}" if any is added later)
+      - shellcheck /home/travis/build/acidanthera/OpenCorePkg/**/*.tool
```
vit9696, line 20: Here too.

.travis.yml
```diff
@@ -15,9 +15,9 @@ matrix:
     script:
       - HOMEBREW_NO_AUTO_UPDATE=1 brew install shellcheck
       # Check Parent Dir (No *.sh files curently. Change "*.tool" to "*.{sh,tool}" if any is added later)
-      - shellcheck /home/travis/build/acidanthera/OpenCorePkg/*.tool
+      - shellcheck ./*.tool
```
vit9696, line 18: In fact, I think you want `shellcheck ./*.tool || exit 1` here, because otherwise the error may not be reported?

dakanji: It will exit with "1" once there is an error.

dakanji:
Can you advise whether `compiler: clang` and `osx_image: xcode10.2` are actually needed for this section?

vit9696:
You could use Linux VMs really, as they are faster. E.g.

```
  - os: linux
    name: "Shellcheck on Linux"
    addons:
      apt:
        packages:
        - shellcheck
```

.travis.yml
vit9696: shellcheck will, but travis does not abort build on a command returning non-zero code.

dakanji: It will continue the build but report that test as failed.
That way can see that Test A failed and Test D failed for amendments.

dakanji: Please confirm.

Do you prefer to abort on the first error met (Your Suggested Change ... Should really then be done for each test. I.e., running macbuild etc should also have similar ) or continue and say so many tests failed (This is how everything is currently setup)?

vit9696: I guess either way is fine to me in this case, I will merge.

dakanji:
Found out that there are some shell scripts with a `command` extension that were not checked before and `/Utilities/LogoutHook/LogoutHook.command` has quite a few issues flagged starting with the shebang which is plain sh but there is one non POSIX compliant item in the code.

That one is easy to fix by changing to bash but will be adding exceptions for the others with links to the ShellCheck wiki as there is no test on this script and can't be sure of implications of changes.

Bearing your reluctance to open many issues, I think an issue on this should be opened for you to evaluate and remove the exceptions later.

Thoughts?

vit9696:
Basically it is like this:

- We will add more scripts soon, e.g. I will likely move `https://github.com/acidanthera/Lilu/blob/master/Lilu/Scripts/kpdescribe.sh` to OpenCorePkg.
- I am ready to review changes and/or ask people to ensure that the scripts continue working after the changes.
- I am not ready to invest time into into shellcheck at least in the nearest future, so if it causes problems I will rather disable it entirely.
- We are slowly changing `bash` to `sh`, because newer macOS use `zsh` by default and we cannot rely on `bash` features anymore.

So if you could fix the script and submit a PR about it I can always help, but the rest is beyond our scope as we have important things to handle.

dakanji:
OK. Will leave the exceptions in place on this one and they will be skipped.
Can be looked at when you are doing your changes.

Note though that shellcheck will catch the sh incompatibilities such as seen here and will help your reviewers later

PR #49, opened 2020-05-09
Update ShellCheck CI

dakanji:
Updates #48

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,23 +7,12 @@ env:
 
 matrix:
   include:
-  - os: linux
-    name: "Shellcheck on Linux"
+  - os: osx
+    name: "Check Shell Scripts"
 
     script:
-      # Travis shellcheck is borked
-      - file="shellcheck-v0.7.1"
-      - suf=".linux.x86_64.tar.xz"
-      - curl -LO "https://github.com/koalaman/shellcheck/releases/download/v0.7.1/${file}${suf}" || exit 1
-      - sum=$(shasum -a 256 "${file}${suf}" | cut -f1 -d' ')
-      - expsum="64f17152d96d7ec261ad3086ed42d18232fcb65148b44571b564d688269d36c8"
-      - if [ "$sum" != "$expsum" ]; then echo "Invalid checksum $sum" ; exit 1 ; fi
-      - tar -xf "${file}${suf}" || exit 1
-      - export PATH="$(pwd)/${file}:$PATH"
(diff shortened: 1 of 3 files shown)


vit9696:
I do not think it is a good idea to pollute the source code of the shell script rather than fix it. In my opinion, either exclude the file from checking or fix the issues.

Rodion (@roddy20) could make sure that the script remains functional.

dakanji:
I don't think comments count as pollution but it's your call.

Script was not included before so the PR can just be closed and the matter will be solved if that is the preference.

vit9696:
I agree that this PR is an enhancement, and I thought you would like to drive shellcheck incorporation in Acidanthera anyway, so all the scripts would be changed sooner or later anyway. Not sure what you mean.

dakanji:
Trying to the best of my knowledge to contribute in a small way and don't want to break anything ... especially with this one as it doesn't have a test to verify if I change according to the suggestions.

That's why I thought it best to flag where there are issues that I can't resolve right now so that I can either get back to them later or someone more knowledgeable can pick them up.

I am only able to handle problems in discrete chunks which is why I thought it was a reasonable incremental step of adding informative comments.

I understand that projects have all sorts of considerations to ponder but If you are willing to accommodate a step by step process, then the added comments would ultimately be dealt with.

Pollution is a bit of  strong word and it was in that context that made me think deleting the PR was a possible way forward as the files in question were not touched before.

vit9696:
Ok, this sounds fine to me as long as it is a temporary measure. I guess handling these changes in a separate PR will indeed be better as first we need to have the tool working and then we could address its issues.

dakanji:
BTW. Left the shebang alone as "sh" but there are two non POSIX compliant items

1. Use of local variable on Line 52: `local key`
2. Use of expansion on 77: `Boot008{1..3}`

Not sure how the script actually worked/works.

Personally I think shebang should change to bash.

Regarding comments elsewhere on not being able to rely on bash, i think that is a bit of a leap as while Apple has indeed changed the DEFAULT shell to zsh, as long as bash is not actually REMOVED, and there is no indication this is imminent, any script with a bash shebang would be interpreted accordingly. 

Same way you could/can run zsh scripts when bash was/is the default as long as you indicated this in the shebang ... Since OS X Tiger when zsh was added (IIRC).

Anyway, that's an item for the project team to decide on but should be done in full disclosure.

vit9696:
In my opinion, we should expect Apple to remove bash in the nearest future as it is GPL and end of life. The use of `local` keyword and expansion could be easily fixed I believe.

PR #55, opened 2020-05-18
fix build.sh exit immediately on macOS

cattyhouse:


cattyhouse:
close as it was fixed by vit

PR #60, opened 2020-05-24
Added VboxHfs.efi driver

ZISQO:
Added missing files for VboxHfs.efi driver users


Andrey1970AppleLife:
HfsPlus works better.

PR #61, opened 2020-05-24
Added VboxHfs.efi driver for Sample.plist

ZISQO:
Added missing files for VboxHfs.efi driver users


vit9696:
We believe if anybody needs this driver he can add it manually.

PR #71, opened 2020-06-08
Added ControlMsrE2 

derBrumbaer:
ControlMsrE2 to edit CFG Lock BIOS Option.
Includes VerifyMsrE2 functionality.

### DIFF
--- a/Application/ControlMsrE2/ControlMsrE2.c
+++ b/Application/ControlMsrE2/ControlMsrE2.c
@@ -59,6 +59,9 @@ UefiMain (
                 ListHeaderCount++;
             }
 
+            // Keep list alive 'til program finishes.
+            // So that all lists can be searched, the results be displayed together.
+            // And from all those one Option will be selected to be changed
             EFI_HII_PACKAGE_LIST_HEADER *ListHeaders[ListHeaderCount];
             
             OptionsCount = 0;
--- a/Application/ControlMsrE2/ControlMsrE2.c
+++ b/Application/ControlMsrE2/ControlMsrE2.c
@@ -1,0 +1,207 @@
+/** @file
+ Control CFGLock BIOS Option
+ 
+ Copyright (c) 2020, Brumbaer. All rights reserved.<BR>
+ This program and the accompanying materials
+ are licensed and made available under the terms and conditions of the BSD License
+ which accompanies this distribution.  The full text of the license may be found at
+ http://opensource.org/licenses/bsd-license.php
+ 
+ THE PROGRAM IS DISTRIBUTED UNDER THE BSD LICENSE ON AN "AS IS" BASIS,
+ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
+ 
+ **/
+#include "ControlMsrE2.h"
+
+#define CONTEXTS_MAX 8
+
+EFI_GUID FormSetupClassGuid = EFI_HII_PLATFORM_SETUP_FORMSET_GUID;
+
+EFI_STATUS
+EFIAPI
+UefiMain (
+          IN EFI_HANDLE        ImageHandle,
+          IN EFI_SYSTEM_TABLE  *SystemTable
+          )
+{
+    EFI_HII_HANDLE * handles;
+    UINT32           OptionsCount;
+    UINT32           ContextsCount;
+    OneOfContext     Contexts[CONTEXTS_MAX];
+    
+    EFI_STATUS Status = InterpretArguments();
+    
+    if (!EFI_ERROR(Status)) {
+        Status = VerifyMSRE2 (ImageHandle, SystemTable);
+    }
+    
+    if (!EFI_ERROR(Status)) {
+        Print(L"\nBIOS Options:\n");
+
+        EFI_STRING SearchString = AllocateStrFromAscii ("cfg");
+        
+        if (IS_INTERACTIVE()) {
+            SearchString = ModifySearchString(SearchString);
+        }
+        
+        handles = HiiGetHiiHandles (NULL);
+        
+        if (handles == NULL) {
+            Print (L"Could not retrieve HiiHandles.\n");
+            Status = EFI_OUT_OF_RESOURCES;
+        }
+        else {
+            UINT32 ListHeaderCount;
+            UINT32 ListHeaderIndex;
+            
+            ListHeaderCount = 0;
+            for (EFI_HII_HANDLE* h = (EFI_HII_HANDLE*) handles; *h != NULL; h++) {
+                ListHeaderCount++;
+            }
+
+            EFI_HII_PACKAGE_LIST_HEADER *ListHeaders[ListHeaderCount];
+            
+            OptionsCount = 0;
+            ContextsCount = 0;
+            ListHeaderIndex = 0;
+            
+            // For Each Handle
+            for (EFI_HII_HANDLE* h = (EFI_HII_HANDLE*) handles; (*h != NULL) && (ContextsCount < CONTEXTS_MAX); h++, ListHeaderIndex++) {
+                
+                ListHeaders[ListHeaderIndex] = HiiExportPackageLists(*h);
+                
+                if (ListHeaders[ListHeaderIndex]) {
+                    
+                    if (IS_VERBOSE()) {
+                        Print (L"Package List: ");
+                        PrintGuid (&ListHeaders[ListHeaderIndex]->PackageListGuid);
+                        Print (L"\n");
+                    }
+                    
+                    // First package in list
+                    EFI_HII_PACKAGE_HEADER* PkgHeader = PADD(ListHeaders[ListHeaderIndex], sizeof(EFI_HII_PACKAGE_LIST_HEADER));
+                    
+                    // For each package in list
+                    do {
+                        if (PkgHeader->Type == EFI_HII_PACKAGE_END) {
+                            break;
+                        }
+                        
+                        if (IS_VERBOSE()) {
+                            Print (L"Package Type: %02X ", PkgHeader->Type);
+                        }
+                        
+                        if (PkgHeader->Type == EFI_HII_PACKAGE_FORMS) {
+                            
+                            EFI_IFR_OP_HEADER* IfrHeader = PADD(PkgHeader, sizeof(EFI_HII_PACKAGE_HEADER));
+                            
+                            if (IfrHeader->OpCode == EFI_IFR_FORM_SET_OP) { // Form Definition must start with FORM_SET_OP
+                                
+                                // Print some Info
+                                if (IS_VERBOSE()) {
+                                    Print (L"Form: ");
+                                    PrintGuid (&(((EFI_IFR_FORM_SET*) IfrHeader)->Guid));
+                                }
+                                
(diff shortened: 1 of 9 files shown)


vit9696:
Hi,

Thanks for the initial draft. That's a good start to begin with, but I will point out several straight-forward issues that do not need my direct comment:

- Currently there are compilation issues caused by an automatically generated memcpy by the compiler and so on. Please check the logs in Travis CI and address those: https://travis-ci.com/github/acidanthera/OpenCorePkg/builds/170286146
- The codestyle looks unpolished, and this needs some cleanup. Check `3.4 Coding Conventions` in Configuration.pdf, it lists most of our recommendations and deviations from [EDK II Codestyle](https://github.com/tianocore/tianocore.github.io/wiki/Code-Style-C), which we adopted throughout in OpenCore. Currently the code is "not there" yet, so I would suggest a cleanup before I could point out any codestyle comments.
- I will also mark some logic issues and choices inline soon.

Thanks!

Application/ControlMsrE2/ControlMsrE2.c
```diff
@@ -0,0 +1,210 @@
...
+    EFI_STATUS Status = InterpretArguments();
+    
+    if (!EFI_ERROR(Status)) {
+        Status = VerifyMSRE2 (ImageHandle, SystemTable);
+    }
+    
+    if (!EFI_ERROR(Status)) {
+        Print(L"\nBIOS Options:\n");
+
+        EFI_STRING SearchString = AllocateStrFromAscii ("cfg");
```
vit9696, line 41: You could use `AsciiStrCopyToUnicode` from OcStringLib.h here.

Application/ControlMsrE2/ControlMsrE2.c
```diff
@@ -0,0 +1,210 @@
...
+        
+        if (IS_INTERACTIVE()) {
+            SearchString = ModifySearchString(SearchString);
+        }
+        
+        handles = HiiGetHiiHandles (NULL);
+        
+        if (handles == NULL) {
+            Print (L"Could not retrieve HiiHandles.\n");
+            Status = EFI_OUT_OF_RESOURCES;
```
vit9696, line 51: There is a memory leak of `SearchString` here.

Application/ControlMsrE2/ControlMsrE2.c
```diff
@@ -0,0 +1,210 @@
...
+                        Print (L"Package List: ");
+                        PrintGuid (&ListHeaders[ListHeaderIndex]->PackageListGuid);
+                        Print (L"\n");
+                    }
+                    
+                    // First package in list
+                    EFI_HII_PACKAGE_HEADER* PkgHeader = PADD(ListHeaders[ListHeaderIndex], sizeof(EFI_HII_PACKAGE_LIST_HEADER));
+                    
+                    // For each package in list
+                    do {
```
vit9696, line 88: Maybe make this `while (ContextsCount < CONTEXTS_MAX || PkgHeader->Type != EFI_HII_PACKAGE_END)`?

Application/ControlMsrE2/ControlMsrE2.c
```diff
@@ -0,0 +1,210 @@
...
+ 
+ THE PROGRAM IS DISTRIBUTED UNDER THE BSD LICENSE ON AN "AS IS" BASIS,
+ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
+ 
+ **/
+#include "ControlMsrE2.h"
+
+#define CONTEXTS_MAX 8
+
+EFI_GUID FormSetupClassGuid = EFI_HII_PLATFORM_SETUP_FORMSET_GUID;
```
vit9696, line 18: Cannot we use gEfiHiiPlatformSetupFormsetGuid by simply including `<Guid/HiiPlatformSetupFormset.h>`, and then adding `MdeModulePkg/MdeModulePkg.dec` and `gEfiHiiPlatformSetupFormsetGuid` dependencies to ControlMsrE2.inf?

Application/ControlMsrE2/ControlMsrE2Hii.c
```diff
@@ -0,0 +1,256 @@
...
+
+typedef VOID EFIAPI OpCodeHandler (EFI_IFR_OP_HEADER* IfrHeader, UINT8* Stop, VOID* Context);
+
+EFI_HII_PACKAGE_LIST_HEADER* HiiExportPackageLists (IN EFI_HII_HANDLE Handle) {
+    
+    UINTN BufferSize = 0;
+    
+    gHiiDatabase->ExportPackageLists (gHiiDatabase, Handle, &BufferSize, NULL);
+    
+    if (BufferSize > 0) {
```
vit9696, line 30: Should not this check for return status?

Application/ControlMsrE2/ControlMsrE2Hii.c
```diff
@@ -0,0 +1,256 @@
...
+        EFI_HII_PACKAGE_LIST_HEADER* result = (EFI_HII_PACKAGE_LIST_HEADER*) AllocatePool (BufferSize);
```
vit9696, line 31: Should not this check for allocation failure?

Application/ControlMsrE2/ControlMsrE2Hii.c
```diff
@@ -0,0 +1,256 @@
...
+        
+        EFI_STATUS Status = gHiiDatabase->ExportPackageLists (gHiiDatabase, Handle, &BufferSize, result);
+        
+        return (Status) ? NULL : result;
```
vit9696, line 35: Better use `EFI_ERROR` and make sure to free memory on failure.

Application/ControlMsrE2/ControlMsrE2Hii.c
```diff
@@ -0,0 +1,256 @@
...
+                 IN void* Context,
+                 IN OpCodeHandler Handler
+                 ) {
+    
+    do {
+        if (Header->OpCode == EFI_IFR_END_OP)
+            return Header;
+        if (Header->OpCode == OpCode) {
+            Handler (Header, Stop, Context);
+            if ((Stop != NULL) && *Stop)
```
vit9696, line 53: If Stop can be NULL and is modified, maybe it should be `IN OUT UINT8 *Stop OPTIONAL` in the header?

Application/ControlMsrE2/ControlMsrE2UI.c
```diff
@@ -0,0 +1,277 @@
...
+    CHAR16 *rp = r;
+    
+    for (CHAR8 *sp = s; *sp != 0; sp++, rp++)
+        *rp = *sp;
+    
+    *rp = 0;
+    return r;
+}
+
+UINT8 EFIAPI Char16IgnoreCaseCompare (CHAR16 a, CHAR16 b) {
```
vit9696, line 32: You could use OcStriCmp from OcStringLib.

Application/ControlMsrE2/ControlMsrE2UI.c
```diff
@@ -0,0 +1,277 @@
...
+    if (a >= 'A' && a <= 'Z') {
+        a = a | 0x20;
+    }
+    if (b >= 'A' && b <= 'Z') {
+        b = b | 0x20;
+    }
+    return a - b;
+}
+
+UINT8 EFIAPI StrContains (CHAR16* a, CHAR16* b) {
```
vit9696, line 42: You could use `OcStriStr` from OcStringLib.

Application/ControlMsrE2/ControlMsrE2UI.c
```diff
@@ -0,0 +1,277 @@
...
+    Print (L"%08x-%04x-%04x-%02x-%02x-%02x-%02x-%02x-%02x-%02x-%02x",
+           Guid->Data1,
+           Guid->Data2,
+           Guid->Data3,
+           Guid->Data4[0], Guid->Data4[1], Guid->Data4[2], Guid->Data4[3], Guid->Data4[4], Guid->Data4[5], Guid->Data4[6], Guid->Data4[7]
+           );
+}
+
+VOID PrintUINT8Str (UINT8 n[1]) {
+    AsciiPrint ((CHAR8*)(&n[0]));
```
vit9696, line 175: This looks really ugly and unsafe, cannot you use `AsciiPrint ("%a", (CHAR8 *) X);`? Perhaps even without this function.

Application/ControlMsrE2/ControlMsrE2.c
derBrumbaer: I can do this.

Application/ControlMsrE2/ControlMsrE2.c
derBrumbaer: There isn't. It's not obvious from this line, but in ModifySearchString the "old" search string is deallocated if changed, otherwise it stays the same.

Application/ControlMsrE2/ControlMsrE2.c
derBrumbaer: Could do that, but I wanted the packages to be handled like in a switch, so it is obvious that this is a  package like all the others but which marks the end of the list.

Application/ControlMsrE2/ControlMsrE2.c
derBrumbaer: Can do that

Application/ControlMsrE2/ControlMsrE2Hii.c
derBrumbaer: I assumed Buffersize will not be changed on a fail, but your right, I don't know.

Application/ControlMsrE2/ControlMsrE2Hii.c
derBrumbaer: Will do

Application/ControlMsrE2/ControlMsrE2.c
vit9696: Right, we will need to do allocation failure checking though, as you cannot pass NULL to FreePool for instance.

Application/ControlMsrE2/ControlMsrE2Hii.c
derBrumbaer: I'll have to check all the Specifiers

Application/ControlMsrE2/ControlMsrE2UI.c
derBrumbaer: Will do

Application/ControlMsrE2/ControlMsrE2UI.c
derBrumbaer: Will do

Application/ControlMsrE2/ControlMsrE2UI.c
derBrumbaer: I can if you insist, but it's a convenience method and IMHO PrintUInt8Str (whatever) look nicer than AsciiPrint ("%a", (CHAR8 *) whatever);

derBrumbaer:
I have to acquaint myself with the libraries, different compilers and your code style first.

Doing it inside a pull request doesn't make much sense.

vit9696:
Right, we have no problems with work being done in pull-requests and we usually do this ourselves by prefixing [WIP] to the name. This helps us to utilise all the checking environment. Yet please choose what is most convenient for you, thanks!

PR #76, opened 2020-06-18
Allow specification of max slide for unavailable range

zhen-zen:
https://github.com/acidanthera/bugtracker/issues/979
Some higher slides may be unavailable and sometimes can cause boot failure when generated random slide fall into this range.

### DIFF
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -1460,6 +1460,22 @@ To view their current state use \texttt{pmset -g} command in Terminal.
   this option is to be enabled.
 
 \item
+  \texttt{ProvideMaxSlide}\\
+  \textbf{Type}: \texttt{plist\ integer}\\
+  \textbf{Failsafe}: \texttt{0}\\
+  \textbf{Description}: Provide maximum KASLR slide when higher ones are unavailable.
+
+  This option overrides the maximum slide instead of \texttt{TOTAL\_SLIDE\_NUM} (256)
+  when \texttt{ProvideCustomSlide} is enabled. When generating random slides, this
+  option can avoid bad slides that are unavailable and cause boot failure.
+
+  \emph{Note}: The necessity of this quirk is determined by random boot failure when
+  \texttt{ProvideCustomSlide} is enabled. Usually the debug log may contain messages
+  like \texttt{AAPL: [EB|`LD:LKC] \} Err(0x9)} when \texttt{AppleDebug} is enabled.
+  To find the optimal value, manually append \texttt{slide=X} to \texttt{boot-args}
+  and log the largest one that won't cause boot failure.
+
+\item
   \texttt{RebuildAppleMemoryMap}\\
   \textbf{Type}: \texttt{plist\ boolean}\\
   \textbf{Failsafe}: \texttt{false}\\
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -1465,13 +1465,17 @@ To view their current state use \texttt{pmset -g} command in Terminal.
   \textbf{Failsafe}: \texttt{0}\\
   \textbf{Description}: Provide maximum KASLR slide when higher ones are unavailable.
 
-  This option overrides the maximum slide instead of \texttt{TOTAL\_SLIDE\_NUM} (256)
-  when \texttt{ProvideCustomSlide} is enabled. When generating random slides, this
-  option can avoid bad slides that are unavailable and cause boot failure.
+  This option overrides the maximum slide instead of \texttt{TOTAL\_SLIDE\_NUM}
+  (256) when \texttt{ProvideCustomSlide} is enabled. It is believed that modern
+  firmwares allocate pool memory from top to bottom, effectively resulting in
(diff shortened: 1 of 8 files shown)


Library/OcAfterBootCompatLib/CustomSlide.c
```diff
@@ -89,11 +89,18 @@ GenerateSlideValue (
...
 
-  do {
-    DivU64x32Remainder (GetPseudoRandomNumber64 (), SlideSupport->ValidSlideCount, &Slide);
-  } while (Slide == 0);
+  //
+  // Handle maximum available slide case.
+  //
+  DivU64x32Remainder (
+    GetPseudoRandomNumber64 (),
+    SlideSupport->ProvideMaxSlide == 0
```
vit9696, line 97: Ugh, this is still wrong. In this implementation `SlideSupport->ProvideMaxSlide` will be maximum number of slides used. I.e. it says first N slides, but these first slides can be anything, they can be 1, 2, 3, 5, 6 or they can be 1, 2, 3, 5, 9, 100, 101, 102, 103 as `SlideSupport->ValidSlides` is a sparse list of values (!).

What I believe this should really do is to add a check [here](https://github.com/acidanthera/OpenCorePkg/blob/796765c/Library/OcAfterBootCompatLib/CustomSlide.c#L365), replacing it with something like:

```C
if (StartAddr + AvailableSize) != EndAddr
  || (SlideSupport->ProvideMaxSlide > 0 && Slide > SlideSupport->ProvideMaxSlide))
```

zhen-zen: Or if we can break the outer for-loop before that line and skip evaluating remaining higher slides?

vit9696: Hm, or actually that. Yeah, that's more obvious of course.

Library/OcAfterBootCompatLib/CustomSlide.c
```diff
@@ -89,11 +89,9 @@ GenerateSlideValue (
     return SlideSupport->ValidSlides[0];
   }
 
-  do {
-    DivU64x32Remainder (GetPseudoRandomNumber64 (), SlideSupport->ValidSlideCount, &Slide);
-  } while (Slide == 0);
+  DivU64x32Remainder (GetPseudoRandomNumber64 (), SlideSupport->ValidSlideCount - 1, &Slide);
```
vit9696, line 92: This change looks wrong to me, as now it avoids the first slide entry in ValidSlides, but it is not necessary 0 (e.g. on Z77 it normally starts with 12, 13, and so on). Please change it back to the original code, that is what we reverse-engineered from Efiboot.

Library/OcAfterBootCompatLib/CustomSlide.c
```diff
@@ -374,6 +380,11 @@ ShouldUseCustomSlideOffset (
     }
   }
 
+  if (SlideSupport->ValidSlideCount < SlideSupport->ProvideMaxSlide) {
```
vit9696, line 383: I do not know what exactly this does, but it looks like some artifact of the previous change. Please remove.

Docs/Configuration.tex
```diff
@@ -1459,6 +1459,22 @@ \subsection{Quirks Properties}\label{booterpropsquirks}
...
 
+\item
+  \texttt{ProvideMaxSlide}\\
+  \textbf{Type}: \texttt{plist\ integer}\\
+  \textbf{Failsafe}: \texttt{0}\\
+  \textbf{Description}: Provide maximum KASLR slide when higher ones are unavailable.
+
+  This option overrides the maximum slide instead of \texttt{TOTAL\_SLIDE\_NUM} (256)
+  when \texttt{ProvideCustomSlide} is enabled. When generating random slides, this
+  option can avoid bad slides that are unavailable and cause boot failure.
```
vit9696, line 1470: Could you add the following afterwards? It is believed that modern firmwares allocate pool memory from top to bottom, effectively resulting in free memory at the time of slide scanning being later used as temporary memory during kernel loading.

Library/OcAfterBootCompatLib/CustomSlide.c
zhen-zen: Or if the condition should be `SlideSupport->ValidSlides[Slide] == 0`  ?

vit9696: Good question, now I see the issue. Let me recheck EfiBoot.

Library/OcAfterBootCompatLib/CustomSlide.c
```diff
@@ -91,7 +91,7 @@ GenerateSlideValue (
 
   do {
     DivU64x32Remainder (GetPseudoRandomNumber64 (), SlideSupport->ValidSlideCount, &Slide);
-  } while (Slide == 0);
+  } while (Slide == 0); // or SlideSupport->ValidSlides[Slide] == 0
```
vit9696, line 94: My bad, should be `SlideSupport->ValidSlides[Slide] == 0` indeed.

PR #77, opened 2020-06-22
ToCompile: Had to declare gOcVendorVariableGuid extern to avoid linke…

olwimo:
…r error when compiled into both the library and utility


vit9696:
Thanks, cannot it just be removed as we have that in includes now?

olwimo:
Sure, I think, just #include User/Include/GlobalVar.h instead of the declaration, much better style to keep declarations centralized; but I'm completely new to this project, so I went with the simplest solution I could see for my first edit :)

PR #78, opened 2020-06-25
Docs: Optimise recommended clone cmd

dakanji:
Add depth option to reduce cloned size (recursion to submodules)


Andrey1970AppleLife:
Don't edit Differences.tex

cattyhouse:
> Add depth option to reduce cloned size (recursion to submodules)

it does nothing, still 1GB

dakanji:
>it does nothing, still 1GB

Drops to 650mb for me with this. Not a fan of the new EDK with the massive size overhead (~150mb for the v0.5.9 one) but must assume the devs have determined it brings even more massive benefits ... I hope!

PR #80, opened 2020-06-26
update SSDT-EC.dsl

naveenkrdy:
Add support for AMD.
Add conditinal detection and disabling of existing EC device.


khronokernel:
Closing due to a few reasons:

- Missing numerous variation of EC naming 
- Missing STA check(many desktop ECs are already disabled, trying to set STA again will error out)
- Not verifying whether valid EC(some desktops use EC0 without setting an HID value, so you'd want to verify it has HID, CRS and GPE as per ACPI spec)

The SSDT-EC provided in this repo is just a sample, so would be best for users to find the EC themselves. You can find several EC naming examples here: [SSDT-EC-DESKTOP.dsl](https://github.com/dortania/Getting-Started-With-ACPI/blob/master/extra-files/decompiled/SSDT-EC-DESKTOP.dsl#L13L65 )

PR #81, opened 2020-06-26
Update SSDT-PLUG.dsl

naveenkrdy:
No point in checking further conditions when one is met.


foskvs:
Actually there is a reason. There are systems where multiple devices are present (example: both _SB.PR00 and _PR.CPU0) but only one is a processor object.

naveenkrdy:
> Actually there is a reason. There are systems where multiple devices are present (example: both _SB.PR00 and _PR.CPU0) but only one is a processor object.
ok ! i see 
then i will close the pull !
thanks !

PR #83, opened 2020-06-27
Add RTC0 patch for ASUS X299 System

ZISQO:
Asus forgot to map the 0x72 and 0x73 regions in RTC device. Look in my original DSDT and find RTC. 
Then see what regions the CRS is missing, for me you can see the length is 0x02 meaning it only contains 0x70 and 0x71.
and the next region starts at 0x74 meaning we have 2 regions missing

so, MykolaG was find this value 0x04.

### DIFF
--- a/Docs/AcpiSamples/SSDT-X299-RTC0.aml
+++ b/Docs/AcpiSamples/SSDT-X299-RTC0.aml
@@ -1,0 +1,56 @@
+/*
+ * Intel ACPI Component Architecture
+ * AML/ASL+ Disassembler version 20180427 (64-bit version)(RM)
+ * Copyright (c) 2000 - 2018 Intel Corporation
+ * 
+ * Disassembling to non-symbolic legacy ASL operators
+ *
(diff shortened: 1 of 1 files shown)


vit9696:
Yes, we are aware of that issue. Why do not you map 0x74~0x77 registers used for no interrupt CMOS access starting with Ivy Bridge though? Pretty sure @khronokernel did it right after the explanation…

Also, the code is incorrect for Windows, as it disables RTC entirely for it by `Name (_SB.PC00.LPC0.RTC._STA, Zero)`.

ZISQO:
Even with this code, I have no problem using Windows 10 via OpenCore. But without this code, x299 TUF mk1 hangs during Big sur boot. that's it. 

also @khronokernel  were refer my bootloader for Big sur and he using Overrides path but that doen't working for my system.

khronokernel:
Added sample in master: https://github.com/acidanthera/OpenCorePkg/commit/62e6ab1ec56628e1ea26c12042ca6c6f2fa7662d

ZISQO:
thank you so much @khronokernel!!!

PR #91, opened 2020-07-13
Load icons for custom boot entries

tkashkin:
This PR adds support for icons for custom entries (acidanthera/bugtracker#1042).

It should try to load an icon for custom entry from the same directory and fallback to trying to load `.VolumeIcon.icns`.

For entry with `Path` = `.../\EFI\path\to\bootloader.efi` it should try to load `.../\EFI\path\to\bootloader.efi.icns`.

I'm not familiar with either UEFI programming nor with OpenCore code, so let me know if I somehow managed to do something wrong in a few lines of code.

### DIFF
--- a/Library/OcBootManagementLib/BootEntryInfo.c
+++ b/Library/OcBootManagementLib/BootEntryInfo.c
@@ -460,7 +460,9 @@ OcGetBootEntryIcon (
   }
 
   if (BootEntry->Type == OC_BOOT_EXTERNAL_OS && BootEntry->PathName != NULL) {
-    // Try to load the icon from the same path with appended .icns extension
+    //
+    // Try to load the icon from the same path with appended .icns extension.
+    //
     Status = InternalGetAppleImage (
       FileSystem,
       BootEntry->PathName,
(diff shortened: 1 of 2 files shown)


Library/OcBootManagementLib/BootEntryInfo.c
```diff
@@ -459,6 +459,25 @@ OcGetBootEntryIcon (
     return Status;
   }
 
+  if (BootEntry->Type == OC_BOOT_EXTERNAL_OS && BootEntry->PathName != NULL) {
+    // Try to load the icon from the same path with appended .icns extension
```
vit9696, line 463: Please use EDK II-style comments with extra surrodunging `//` here and below, i.e.:

```C
//
// Something 
//
```

vit9696:
Thanks! This looks good to me, @Download-Fritz also does not mind. @usr-sse2 is the change ok to you? If so, we could merge this after the minor codestyle issue is addressed. Other than that, please update the `Configuration.tex` in regards to the new behaviour for the icon paths. Do not regenerate the PDFs to reduce the conflict possibility though.

vit9696:
According to documentation we have .icns suffix variant used for custom entries, and root volume icns for everything else. Yet currently the former will fallback to the latter, which is inconsistent. I do not mind, which route we take: fixing up the docs or updating the code to stop using .VolumeIcon.icns, yet please do either.

tkashkin:
@vit9696 I've tried to update docs but I'm very bad at explaining and documenting things.

> Yet currently the former will fallback to the latter, which is inconsistent.

It looks like more correct behaviour to me and, more importantly, it does not change current behaviour unless `<ENTRY_PATH>.icns` actually exists and loads correctly.

PR #92, opened 2020-07-27
correct EC scope in SSDT-EC-USBX.dsl

rektimus:
can not nest \_SB under \_SB, possible better ways to organize scopes

### DIFF
--- a/Docs/AcpiSamples/SSDT-EC-USBX.dsl
+++ b/Docs/AcpiSamples/SSDT-EC-USBX.dsl
@@ -78,7 +78,7 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "SsdtEC", 0x00001000)
             }
         }
 
-        Scope (\_SB.PCI0.LPCB)
+        Scope (PCI0.LPCB)
         {
             Device (EC)
             {


williambj1:
It’s not nested, / starts from root

asgmeonerandom:
educate yourself https://uefi.org/sites/default/files/resources/ACPI_6_3_final_Jan30.pdf

rektimus:
> It’s not nested, / starts from root

It's ...

Scope (\_SB)
{
    Scope (\_SB.PCI0.LPCB
    {
        ...
    }
}

that's nested. using \ or ^ does not work in SSDTs as it's not running from within the firmware.

I just verified this, error was AE_ALREADY_EXISTS, changing scope removed error and allowed SSDT to load.

rektimus:
Found similar issue here:

https://github.com/acidanthera/bugtracker/issues/678

zhen-zen:
> > It’s not nested, / starts from root
> 
> It's ...
> 
> ```
> Scope (\_SB)
> {
>     Scope (\_SB.PCI0.LPCB
>     {
>         ...
>     }
> }
> ```
> 
> that's nested. using \ does not work in SSDTs scope when nested (^ doesn't work at all) as it's not running from within the firmware.
> 
> I just verified this, error was AE_ALREADY_EXISTS, changing scope removed error and allowed SSDT to load.

That's not a issue raised by this SSDT. Probably that guy forgot to rename their own EC or left EC0 to EC enabled. On an untouched DSDT table, there should be no device named `EC` which is a bogus device injected by this SSDT.

And what @williambj1 said is correct, you can start from root in any scope. Both DSDT and SSDTs share the same namespace and there's no difference between "firmware" and injected SSDT. Spend some time and read the ACPI spec.

Personally I'm not in favor of this kind of merging. The External declaration above announced the same scope for LPCB. For readability, it makes more sense just moving the EC device part out of _SB scope.

And for the commented part in this SSDT, I'm afraid that the part won't work on some of the laptops? @PMheart once wrote an experimental driver to break `AppleACPIEC`. When that blocked, all EC fields are non-functional on some laptops. Maybe `AppleACPIEC ` has something to do with EC initialization? @Andrey1970AppleLife @vit9696

Andrey1970AppleLife:
Fictitious EC only for desktops.
On laptops EC is necessary as is, rename EC if a name another.

vit9696:
@zhen-zen is correct. I kind of missed that patch. @Andrey1970AppleLife, could you revert the change and instead move the `Scope (\_SB.PCI0.LPCB)` to the very end of the file out of the `\_SB` scope? This also makes it more convenient to edit LPCB path if necessary.

rektimus:
Just to clarify, I reviewed the DSDT, EC was already disabled in DSDT, and was not disabled in SSDT. All I did was correct nested SB and it worked.

zhen-zen:
> Just to clarify, I reviewed the DSDT, EC was already disabled in DSDT, and was not disabled in SSDT. All I did was correct nested SB and it worked.
> 
> eta: moving the devices into separate scopes is what I meant by "possible better ways to organize scopes", in fact moving them to separate ssdts is even better, you're following my advice but claiming I'm wrong and need to read more...
> 
> if the ssdt is fine as is, why the need to change it? because clearly it is wrong.

Don't be that furious. Apparently you didn't read my explanation and neither checked the spec.

The following extracts are from 5.3 ACPI Namespace in [ACPI Specification, Version 6.3](https://uefi.org/sites/default/files/resources/ACPI_6_3_final_Jan30.pdf) (Page 266):

> • A name proceeded with ‘\’ causes the name to refer to the root of the namespace (‘\’ is not part of the 32-bit fixed-length name).
> Except for names preceded with a ‘\’, the current namespace determines where in the namespace hierarchy a name being created goes and where a name being referenced is found.

So the old SSDT worked as expected and your claim on nested scope with `\` is false.

Also for your claim about "^ doesn't work at all" (on the same page).

> There are two types of namespace paths: an absolute namespace path (that is, one that starts with a ‘\’ prefix), and a relative namespace path (that is, one that is relative to the current namespace). The namespace search rules discussed above, only apply to single NameSeg paths, which is a relative namespace path. For those relative name paths that contain multiple NameSegs or Parent Prefixes, ‘^’, the search rules do not apply. If the search rules do not apply to a relative namespace path, the namespace object is looked up relative to the current namespace. For example:
```
ABCD.        //search rules apply
^ABCD.       //search rules do not apply
XYZ.ABCD.    //search rules do not apply
\XYZ.ABCD.   //search rules do not apply
```

Honestly, ACPI programming is not that as easy as it seems and there's much to learn from from the doc.

And for “possible better ways to organize scopes”, there's no opposition. However, it's not simply merging objects under the same scope. IIRC, only Apple or some old machines have a such clean DSDT with everything rest in their parent scope. Most vendors, including Intel reference code, have intensive use of absolute path and that sometimes can help to organize objects with related function but not in the same scope. And sometimes just disordered :P.

For this case, I have already described my suggestions above. Since there's no enforced code-style for ACPI programming like other language, it's more about convenience and readability.

Separating SSDTs is another issue. It allows control on a smaller granularity, which can be enabled and disabled separately. And currently it's typical to group them by functionality. While some people do merge their SSDTs into single one after fully tested.

PR #93, opened 2020-07-30
OcInputLib: Do not reset input protocols

varahash:
On some firmware the reseting input protocols upon OpenCore initialization will discard all currently pressed keys on keyboard. This leads to issue where user unable to trigger showing boot picker by just holding OPT key. https://github.com/acidanthera/bugtracker/issues/564 

Alternatively, we can make this behaviour configurable with option `ResetInputProtocols=true/false`, but this seems unnecessary and just removing 'reset' should work everywhere.

### DIFF
--- a/Library/OcInputLib/Keycode/AIKShim.c
+++ b/Library/OcInputLib/Keycode/AIKShim.c
@@ -21,27 +21,6 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 
 EFI_STATUS
 EFIAPI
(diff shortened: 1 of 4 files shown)


vit9696:
Hello, thanks a lot for this. We will test this thoroughly and merge into 0.6.1 to avoid a potential regression. This code is rather old, and I do not clearly remember why a reset was done this way. Will also let us decide whether a configuration option is needed. Could you provide more details on which hardware you were able to reproduce the issue and the fix?

varahash:
Hello, here is hardware:
**Motherboard**: Asus ROG MAXIMUS VIII GENE
**Keyboard**: Logitech K270

PR #94, opened 2020-08-05
Virtual folders implementation

vit9696:
Created for review.

### DIFF
--- a/Include/Acidanthera/Library/OcVirtualFsLib.h
+++ b/Include/Acidanthera/Library/OcVirtualFsLib.h
@@ -19,6 +19,8 @@
 
 #include <Protocol/SimpleFileSystem.h>
 
+#include <Guid/FileInfo.h>
+
 /**
   Creates read-only EFI_FILE_PROTOCOL instance over a buffer allocated
   from pool. On success FileName and FileData ownership is transferred
@@ -46,6 +48,32 @@ CreateVirtualFile (
   );
 
 /**
+  Creates read-only EFI_FILE_PROTOCOL instance over a buffer allocated
+  from pool, but copies FileName internally. On success FileData ownership is transferred
+  to the resulting EFI_FILE_PROTOCOL, which frees them with FreePool
+  upon closing EFI_FILE_PROTOCOL.
+
+  Resulting EFI_FILE_PROTOCOL has 2nd revision, but may be downgraded
+  to 1st by updating the corresponding field.
+
+  @param[in]      FileName         Pointer to the file's name.
+  @param[in]      FileBuffer       Pointer to the file's data.
+  @param[in]      FileSize         File size of FileData.
+  @param[in]      ModificationTime File modification date, optional.
+  @param[in,out]  File             Resulting file protocol.
+
+  @return  EFI_SUCCESS if instance was successfully created.
+**/
+EFI_STATUS
+CreateVirtualFileFileNameCopy (
+  IN     CHAR16             *FileName,
+  IN     VOID               *FileBuffer,
+  IN     UINT64             FileSize,
+  IN     EFI_TIME           *ModificationTime OPTIONAL,
+  IN OUT EFI_FILE_PROTOCOL  **File
+  );
+
+/**
   Creates virtual file system instance around any file.
   CreateRealFile or CreateVirtualFile must be called from
   any registered OpenCallback.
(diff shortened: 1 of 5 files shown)


Include/Acidanthera/Library/OcVirtualFsLib.h
```diff
@@ -45,6 +47,32 @@ CreateVirtualFile (
   IN OUT EFI_FILE_PROTOCOL  **File
   );
 
+/**
+  Creates read-only EFI_FILE_PROTOCOL instance over a buffer allocated
+  from pool, but copies FileName internally. On success FileData ownership is transferred
```
vit9696, line 52: 1. Is `ModificationTime` also copied? I guess it is, so better to remove `, but copies FileName internally` and add `All the other fields are copied.` right after `upon closing EFI_FILE_PROTOCOL`.

2. To make copying even more explicit, please add `CONST` parameter to the not owned pointer arguments (`FileName`, `ModificationTime`).

3. File should be just out, as we do not read its value, but only write. This is in fact true for several other functions in this file.

Include/Acidanthera/Library/OcVirtualFsLib.h
```diff
@@ -66,6 +94,82 @@ CreateRealFile (
...
+
+  @param[in]      FileName         Pointer to the file's name.
+  @param[in]      ModificationTime File modification date, optional.
+  @param[in]      UnderlyingFile   The underlying file protocol, optional.
+  @param[in,out]  File             Resulting file protocol.
+
+  @return  EFI_SUCCESS if instance was successfully created.
+**/
+EFI_STATUS
+VirtualDirCreateOverlay (
```
vit9696, line 115: Points 1 and 2 from `CreateVirtualFileFileNameCopy` are valid for this one.

Include/Acidanthera/Library/OcVirtualFsLib.h
```diff
@@ -66,6 +94,82 @@ CreateRealFile (
...
+
+  @param[in]      FileName         Pointer to the file's name.
+  @param[in]      ModificationTime File modification date, optional.
+  @param[in]      UnderlyingFile   The underlying file protocol, optional.
+  @param[in,out]  File             Resulting file protocol.
+
+  @return  EFI_SUCCESS if instance was successfully created.
+**/
+EFI_STATUS
+VirtualDirCreateOverlayFileNameCopy (
```
vit9696, line 140: Points 1 and 2 from `CreateVirtualFileFileNameCopy` are valid for this one.

Include/Acidanthera/Library/OcVirtualFsLib.h
```diff
@@ -66,6 +94,82 @@ CreateRealFile (
...
+  IN     CHAR16             *FileName,
+  IN     EFI_TIME           *ModificationTime OPTIONAL,
+  IN     EFI_FILE_PROTOCOL  *UnderlyingFile OPTIONAL,
+  IN OUT EFI_FILE_PROTOCOL  **File
+  );
+
+/**
+  Adds an EFI_FILE_INFO instance to the virtual directory.
+  Ownership of the instance is retained until the EFI_FILE_PROTOCOL is closed.
```
vit9696, line 149: It is not clear who frees that memory and what does retain mean in this regard.

Library/OcVirtualFsLib/VirtualDir.c
```diff
@@ -0,0 +1,586 @@
...
+  OUT EFI_FILE_PROTOCOL       **NewHandle,
+  IN  CHAR16                  *FileName,
+  IN  UINT64                  OpenMode,
+  IN  UINT64                  Attributes
+  )
+{
+  EFI_STATUS         Status;
+  VIRTUAL_DIR_DATA   *Data;
+
+  ASSERT (This != NULL);
```
vit9696, line 45: Please do not use `ASSERT` for protocol functions, as we want the checks to be valid in RELEASE builds and actually return an error — `EFI_INVALID_PARAMETER`. This is an EDK II convention at least.

This applies to all the functions below.

Library/OcVirtualFsLib/VirtualDir.c
```diff
@@ -0,0 +1,586 @@
...
+  //
+  if (Data->CurrentEntry != NULL && IsNodeAtEnd (&Data->Entries, Data->CurrentEntry)) {
+    Data->CurrentEntry = NULL;
+
+    *BufferSize = 0;
+    return EFI_SUCCESS;
+  }
+
+  *BufferSize = ReadSize;
+  DirEntryLink = Data->CurrentEntry != NULL ? GetNextNode (&Data->Entries, Data->CurrentEntry) : GetFirstNode (&Data->Entries);
```
vit9696, line 173: Maybe make this an `if`? Looks really hard to read.

Library/OcVirtualFsLib/VirtualDir.c
```diff
@@ -0,0 +1,586 @@
...
+  //
+  ReadSize = *BufferSize;
+  if (Data->CurrentEntry == NULL && Data->UnderlyingProtocol != NULL) {
+    Status = Data->UnderlyingProtocol->Read (
+      Data->UnderlyingProtocol,
+      BufferSize,
+      Buffer
+      );
+
+      if (EFI_ERROR (Status) || *BufferSize != 0) {
```
vit9696, line 149: Please leave a comment on `*BufferSize != 0` check. I guess it is about routing to the original function till we have data from it.

Library/OcVirtualFsLib/VirtualDir.c
```diff
@@ -0,0 +1,586 @@
...
+
+  *BufferSize = ReadSize;
+  DirEntryLink = Data->CurrentEntry != NULL ? GetNextNode (&Data->Entries, Data->CurrentEntry) : GetFirstNode (&Data->Entries);
+
+  //
+  // Get next file info structure.
+  //
+  DirEntry = GET_VIRTUAL_DIR_ENTRY_FROM_LINK (DirEntryLink);
+  if (DirEntry->FileInfo->Size < SIZE_OF_EFI_FILE_INFO) {
+    return EFI_DEVICE_ERROR;
```
vit9696, line 180: Does that mean a memory corruption as I do not see how it could happen in normal functioning. If so, please add an `ASSERT (FALSE)` here to detect it early.

Library/OcVirtualFsLib/VirtualDir.c
```diff
@@ -0,0 +1,586 @@
...
+  }
+
+  //
+  // Ensure entry size matches string length.
+  //
+  FileStrSize = StrSize (DirEntry->FileInfo->FileName);
+  if (OcOverflowAddUN (SIZE_OF_EFI_FILE_INFO, FileStrSize, &ReadSize) || ReadSize != DirEntry->FileInfo->Size) {
+    return EFI_DEVICE_ERROR;
```
vit9696, line 188: Same here.

Library/OcVirtualFsLib/VirtualDir.c
```diff
@@ -0,0 +1,586 @@
...
+
+    ZeroMem (FileInfo, InfoSize - NameSize);
+    FileInfo->Size = InfoSize;
+
+    CopyMem (&FileInfo->CreateTime, &Data->ModificationTime, sizeof (FileInfo->ModificationTime));
+    CopyMem (&FileInfo->LastAccessTime, &Data->ModificationTime, sizeof (FileInfo->ModificationTime));
+    CopyMem (&FileInfo->ModificationTime, &Data->ModificationTime, sizeof (FileInfo->ModificationTime));
+
+    //
+    // Return zeroes for timestamps.
```
vit9696, line 324: This comment is clearly no longer valid. Maybe remove it?

Library/OcVirtualFsLib/VirtualDir.c
```diff
@@ -0,0 +1,586 @@
...
+VirtualDirCreateOverlay (
+  IN     CHAR16             *FileName,
+  IN     EFI_TIME           *ModificationTime OPTIONAL,
+  IN     EFI_FILE_PROTOCOL  *UnderlyingFile OPTIONAL,
+  IN OUT EFI_FILE_PROTOCOL  **File
+  )
+{
+  VIRTUAL_DIR_DATA  *Data;
+
+  ASSERT (FileName != NULL);
```
vit9696, line 480: Just in case. Unlike in protocol functions, in library code we want `ASSERT` indeed. Please do **not** change `ASSERT` from here onwards.

Include/Acidanthera/Library/OcVirtualFsLib.h
```diff
@@ -66,6 +94,82 @@ CreateRealFile (
   IN OUT EFI_FILE_PROTOCOL  **File
   );
 
+/**
+  Creates read-only EFI_FILE_PROTOCOL virtual directory instance,
+  optionally as an overlay over an existing EFI_FILE_PROTOCOL instance.
```
vit9696, line 99: I do not see `optionally as an overlay over an existing EFI_FILE_PROTOCOL instance` in the code. `File` is always overwritten regardless of the passed value. What do I miss?

Include/Acidanthera/Library/OcVirtualFsLib.h
```diff
@@ -66,6 +94,82 @@ CreateRealFile (
...
+VirtualDirCreateOverlay (
+  IN     CHAR16             *FileName,
+  IN     EFI_TIME           *ModificationTime OPTIONAL,
+  IN     EFI_FILE_PROTOCOL  *UnderlyingFile OPTIONAL,
+  IN OUT EFI_FILE_PROTOCOL  **File
+  );
+
+/**
+  Creates read-only EFI_FILE_PROTOCOL virtual directory instance,
+  optionally as an overlay over an existing EFI_FILE_PROTOCOL instance.
```
vit9696, line 124: I do not see `optionally as an overlay over an existing EFI_FILE_PROTOCOL instance` in the code. `File` is always overwritten regardless of the passed value. What do I miss?

Library/OcVirtualFsLib/VirtualFsInternal.h
```diff
@@ -40,8 +41,27 @@
     VIRTUAL_FILE_DATA_SIGNATURE  \
     )
 
+#define VIRTUAL_DIR_DATA_SIGNATURE  \
+  SIGNATURE_32 ('V', 'F', 'S', 'd')
+
+#define VIRTUAL_DIR_FROM_PROTOCOL(This) \
+  CR (                           \
+    This,                        \
+    VIRTUAL_DIR_DATA,           \
```
vit9696, line 50: Would you mind aligning `\` here?

Library/OcVirtualFsLib/VirtualFsInternal.h
```diff
@@ -55,11 +75,38 @@ struct VIRTUAL_FILE_DATA_ {
...
+//
+#define VIRTUAL_DIR_ENTRY_SIGNATURE  SIGNATURE_32 ('V', 'S', 'd', 'L')
+
+/**
+  Gets the next element in list of VIRTUAL_DIR_ENTRY.
+
+  @param[in] This  The current ListEntry.
+**/
+#define GET_VIRTUAL_DIR_ENTRY_FROM_LINK(This)  \
+  (CR (                                     \
```
vit9696, line 105: Would you mind aligning `\` here?

vit9696:
Merged, thx

PR #96, opened 2020-08-08
OcAppleKernelLib: Implement 64-bit cacheless kext injection

vit9696:
references acidanthera/bugtracker#358

### DIFF
--- a/Library/OcAppleKernelLib/CachelessContext.c
+++ b/Library/OcAppleKernelLib/CachelessContext.c
@@ -151,15 +151,18 @@ ScanExtensions (
   File->SetPosition (File, 0);
 
   do {
+    //
+    // Apple's HFS+ driver does not adhere to the spec and will return zero for
+    // EFI_BUFFER_TOO_SMALL. EFI_FILE_INFO structures larger than 1KB are
+    // unrealistic as the filename is the only variable.
+    //
     FileInfoSize = SIZE_1KB - sizeof (CHAR16);
     Status = File->Read (File, &FileInfoSize, FileInfo);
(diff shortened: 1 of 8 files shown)


Library/OcAppleKernelLib/CachelessContext.c
```diff
@@ -0,0 +1,1142 @@
...
+
+  do {
+    FileInfoSize = SIZE_1KB - sizeof (CHAR16);
+    Status = File->Read (File, &FileInfoSize, FileInfo);
+    if (Status == EFI_BUFFER_TOO_SMALL) {
+      FileInfoSize = SIZE_2KB - sizeof (CHAR16);
+      FileInfo = ReallocatePool (SIZE_2KB, SIZE_1KB, FileInfo);
+      Status = File->Read (File, &FileInfoSize, FileInfo);
+      //
+      // TODO? would the buffer ever be this big realistically.
```
vit9696, line 161: This should contain file info for a single file info, correct? I believe it is safe to assume it will not be that large in a valid scenario. For an invalid scenario (e.g. FS corruption) we should make sure that the failure here is handled and error and exits the function. Please add an error check and an explanation.

Library/OcAppleKernelLib/CachelessContext.c
```diff
@@ -0,0 +1,1142 @@
...
+      //
+    }
+
+    if (FileInfoSize > 0) {
+      if (StrCmp (&FileInfo->FileName[StrLen (FileInfo->FileName) - 5], L".kext") == 0) {
```
vit9696, line 166: Here `FileName` is not guaranteed to be over 5 characters.

PR #97, opened 2020-08-09
Implement kernel verification via Apple Secure Boot

vit9696:
Before merging this still needs to be done:

- [x] Interoperation between kernel patcher and secure boot
- [ ] Lowering secure boot policy when performing DMG loading


vit9696:
Merged. Will track personalised DMG loading in https://github.com/acidanthera/bugtracker/issues/1071.

PR #98, opened 2020-08-11
Fix typo from Configuration.tex

1alessandro1:


vit9696:
Thanks! Fixed in master.

PR #107, opened 2020-08-23
Updated BIOSVersion for IM201.yaml and CPUs for IM201 and IM202.yaml.

bootmacos:
Updated BIOSVersion in IM201.yaml to reflect the one that spotted in an actual iMac20,1 machine.
Updated supporting CPUs in IM201 and IM202.yaml.
Auto-generated related header files, i.e., ones for OcMacInfoLib and macserial.

### DIFF
--- a/Library/OcAppleSecureBootLib/OcAppleSecureBootLib.c
+++ b/Library/OcAppleSecureBootLib/OcAppleSecureBootLib.c
@@ -601,6 +601,12 @@ InternalGetImg4ByPath (
     return EFI_NO_MEDIA;
   }
 
+  OcFileDevicePathFullName (
+    Path,
+    (FILEPATH_DEVICE_PATH *)DevicePath,
+    ImagePathSize
+    );
+
   Result = mSbHardwareModel[0] != '\0';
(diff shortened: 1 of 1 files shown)


bootmacos:
I think that a BIOSVersion belongs to the hardware, not to macOS, and it should not be changed unless firmware-update comes along with os-update, like when APFS-boot was introduced. The 1037.147.4.0.0 is shown in two of actual iMac 2020 (iMac20,1, Catalina) by running macserial, and most probably this will not change after updating them to Big Sur. Moreover,     1037.147.4.0.0 is feasible considering that the one in iMac19,1 is 1037.140.50.0.0. The difference suggests proper update has been applied to the successor (iMac20,1).

vit9696:
> I think that a BIOSVersion belongs to the hardware, not to macOS, and it should not be changed unless firmware-update comes along with os-update, like when APFS-boot was introduced.

Correct, and for T2 machines every macOS update (including minor) means firmware update. This is how T2 machines work.

bootmacos:
Thank you very much for the explanation. I understand.

PR #110, opened 2020-08-26
 Reset timeout on cursor interaction, spelling fix

reimertz:
My keyboard is connected to a Bluetooth reciever and therefore doesn't work.
Sadly, my Logitech G603 + 2.4Ghz reciever is a tad laggy and this results in me 50% of the time not being able to select what I'd like to boot up. 

So I added the feature of also disabling the timeout on pointer events. 

Does this make sense? Please review this carefully. Was a looong time ago I wrote any C code.. :)

Also fixed a spelling error.

### DIFF
--- a/Platform/OpenCanopy/OpenCanopy.c
+++ b/Platform/OpenCanopy/OpenCanopy.c
@@ -1201,7 +1201,7 @@ GuiDrawLoop (
         }
       }
       //
-      // If detected key press then disable menu timeout
+      // If detected pointer event then disable menu timeout
       //
       TimeOutSeconds = 0;
     }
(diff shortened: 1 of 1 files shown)


vit9696:
Where is the cursor interaction check here? To me it looks like it always stops the timeout as first iteration. Also, I believe that some mouse actually do very slight movements even when still. This may be undesired?..

Platform/OpenCanopy/OpenCanopy.c
```diff
@@ -986,7 +986,7 @@ GuiLibConstruct (
     DEBUG ((DEBUG_WARN, "OCUI: Failed to initialise pointer\n"));
   }
 
-  mKeyContext = GuiKeyConstruct (PickerContet);
+  mKeyContext = GuiKeyConstruct (PickerContext);
   if (mKeyContext == NULL) {
     DEBUG ((DEBUG_WARN, "OCUI: Failed to initialise key input\n"));
```
sergeev17, line 991: ***

sergeev17 (approved):

PR #113, opened 2020-08-31
Update recommendation on ACPI patches

zhen-zen:


vit9696:
Looks fine to me, but needs the merge of https://github.com/acidanthera/VoodooPS2/pull/22. I am yet to clean it up, as I am still not satisfied with the methods used for detection.

zhen-zen:
I'm sorry for urging that. My laptop suddenly died and I just want to ensure if I didn't mess up things or have leftovers before regaining access for testing.

vit9696:
Yes, I am also sorry for not giving any reply on that issue for some time. I will try to get my hand on it, but I am afraid we will only make it with 0.6.2 as there is only one week left for 0.6.1 and currently we are still not fully over Secure Boot and Big Sur, and everyone is very loaded.

PR #115, opened 2020-09-03
Return Status on Setting Renderer

dakanji:
Closes acidanthera/bugtracker#1133


dakanji:
-Werror=unused-but-set-variable is a bit on the extreme side!
Suppose could just call without setting anything.

vit9696:
I do not agree with EFIAPI, as this is not a protocol function, but otherwise looks fine. Please drop EFIAPI and I will merge. Also, there are some unnecessary newlines around OcSetupConsole.

PR #117, opened 2020-09-05
Return Status on Providing Console GOP

dakanji:
Closes acidanthera/bugtracker#1135

### DIFF
--- a/Library/OcConsoleLib/ConsoleGop.c
+++ b/Library/OcConsoleLib/ConsoleGop.c
@@ -89,7 +89,7 @@ ConsoleHandleProtocol (
   return EFI_UNSUPPORTED;
 }
 
-VOID
+EFI_STATUS
 OcProvideConsoleGop (
   IN BOOLEAN  Route
   )
@@ -132,7 +132,7 @@ OcProvideConsoleGop (
     //
     if (OriginalGop->Mode->MaxMode > 0) {
       mConsoleGraphicsOutput = OriginalGop;
-      return;
+      return EFI_ALREADY_STARTED;
     }
 
     DEBUG ((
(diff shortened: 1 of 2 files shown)


Library/OcConsoleLib/ConsoleGop.c
```diff
@@ -150,51 +150,52 @@ OcProvideConsoleGop (
 
     if (EFI_ERROR (Status)) {
       DEBUG ((DEBUG_INFO, "OCC: No handles with GOP protocol - %r\n", Status));
-      return;
```
vit9696, line 153: Do not change the control flow here and increase the indentation level. Return status here instead and update return calls below.

Library/OcConsoleLib/ConsoleGop.c
```diff
@@ -192,8 +192,10 @@ OcProvideConsoleGop (
     }
 
     mConsoleGraphicsOutput = Gop;
+    return EFI_SUCCESS;
   } else {
```
vit9696, line 196: Drop the else, do not use else after `return`.

dakanji: A bit confused. Wouldn't deleting the else block result in some items in the flow (those that have an error status at the LocateProtocol command) get to the end without returning anything? 

Actually, the information about the missing compatible GOP at this stage is important (why I am looking at this in the first instance) and needed to return a different status to ID this condition.

Alternative is to move the EFI_SUCCESS to the end of the flow.

Thoughts?

vit9696: Not deleting the `else` block, but rewriting the control flow to:

```C
if (x) {
  …
  return EFI_SUCCESS;
}

DEBUG ((DEBUG_WARN, "OCC: Missing compatible GOP - %r\n", Status));
return EFI_NOT_FOUND;
```

It is considered bad practice to write an `else` case when the previous case does not return control. See more details [here](http://clang.llvm.org/extra/clang-tidy/checks/readability-else-after-return.html).

dakanji: Got it.

Anyway, already pushed a change that does not need this. Just returns status after the if/else block.

PR #118, opened 2020-09-10
32-bit Mach-O and patching support

Goldfish64:
- [X] 32-bit OcMachoLib functions
- [X] 32-bit kext injection and patching
- [x] Port relevant quirks to 32-bit
- [x] Port CPUID Emulate functionality to 32-bit
- [x] 32-bit kext blocker


Library/OcAppleKernelLib/KextPatcher.c
```diff
@@ -84,33 +90,66 @@ PatcherInitContextFromBuffer (
...
-    &Context->MachContext,
-    &InnerContext,
-    Buffer,
-    BufferSize,
-    NULL
-    );
-  if (EFI_ERROR (Status)) {
-    return Status;
+  Context->Is32Bit        = Is32Bit;
+  Context->VirtualBase    = VirtualAddress < FileOffset ? VirtualAddress - FileOffset : FileOffset; // FIXME: 32-bit MH_OBJECT
```
vit9696, line 137: Not sure I understand the FIXME. Maybe this value needs to be signed now or we need to rely on unsigned wraparound? `VirtualBase` is essentially the value we should subtract from a virtual pointer to translate a virtual address (for instance, an address of a symbol) to a file offset for patching.

Library/OcAppleKernelLib/KextPatcher.c
```diff
@@ -84,33 +90,66 @@ PatcherInitContextFromBuffer (
...
-    "__TEXT"
-    );
-  if (Segment == NULL || Segment->VirtualAddress < Segment->FileOffset) {
-    return EFI_NOT_FOUND;
+  //
+  // 32-bit can be of type MH_OBJECT, which has all sections in a single unnamed segment.
+  // We'll fallback to that if there is no __TEXT segment.
+  //
+  if (Is32Bit) {
+    Segment32 = MachoGetSegmentByName32 (
```
vit9696, line 103: Do any 32-bit kexts have `__TEXT` segment? If not, maybe it is better to only use section lookup and/or fallback to `__TEXT` if it is less common?

Include/Acidanthera/Library/OcMachoLib.h
```diff
@@ -32,18 +32,62 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 /// only.  Members are not guaranteed to be sane.
 ///
 typedef struct {
-  MACH_HEADER_64        *MachHeader;
+  MACH_HEADER_ANY       *MachHeader;
   UINT32                FileSize;
+  BOOLEAN               Is32Bit;
```
vit9696, line 37: Might be better to move it to the end of the structure to reduce the padding.

Include/Acidanthera/Library/OcMachoLib.h
```diff
@@ -32,18 +32,62 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
...
+
   UINT32                ContainerOffset;
   MACH_SYMTAB_COMMAND   *Symtab;
-  MACH_NLIST_64         *SymbolTable;
+  MACH_NLIST            *SymbolTable32;
```
vit9696, line 41: Any reason not to make these a `union`? Same with `IndirectSymbolTable` below.

Include/Acidanthera/Library/OcMachoLib.h
```diff
@@ -329,7 +615,37 @@ MachoGetSymbolName64 (
...
+  Retrieves a Symbol's name.
+
+  @param[in,out] Context  Context of the Mach-O.
+  @param[in]     Symbol   Symbol to retrieve the name of.
+
+  @retval symbol name.
+
+**/
+CONST CHAR8 *
+MachoGetSymbolNameAny (
```
vit9696, line 627: The choice of `Any`-suffixed and "none"-suffixed versions (as opposed to `32`/`64`) for generic function names is a little worrying to me. Do we really need `Any` in most places we have it? Is there any logic behind `Any`?

Library/OcMachoLib/CxxSymbolsX.h
```diff
@@ -0,0 +1,114 @@
...
+  IN OUT OC_MACHO_CONTEXT     *Context,
+  IN     CONST MACH_NLIST_X   *Smcp
+  )
+{
+  MACH_NLIST_X  *Symbol;
+  BOOLEAN       Result;
+
+  ASSERT (Context != NULL);
+  ASSERT (Smcp != NULL);
+  MACH_ASSERT_X;
```
vit9696, line 28: I might actually request `MACH_ASSERT_X` to be a functional macro taking `Context` as an argument, because relying on `Context` being the Mach-O context is very dirty and counter-intuitive.

Library/OcMachoLib/MachoX.h
```diff
@@ -0,0 +1,68 @@
...
+#define MACH_SECTION_X              MACH_SECTION
+#define MACH_SEGMENT_COMMAND_X      MACH_SEGMENT_COMMAND
+#define MACH_NLIST_X                MACH_NLIST
+
+#define MACH_LOAD_COMMAND_SEGMENT_X MACH_LOAD_COMMAND_SEGMENT
+
+#define MACH_X(a)                   a##32
+#define MACH_ASSERT_X               ASSERT (Context->Is32Bit)
+
+#define MACH_UINT32_CAST(a)         (a)
```
vit9696, line 47: Are we sure this does not need a cast? Probably it does not, but it might want a comment on what the contract for that macro is.

vit9696:
The changes look pretty much good, and I believe we should be able to merge this relatively soon.

Library/OcAppleKernelLib/KextPatcher.c
```diff
@@ -282,23 +365,28 @@ PatcherBlockKext (
     return EFI_UNSUPPORTED;
   }
 
-  KmodOffset = Context->VirtualKmod - Context->VirtualBase;
-  KmodInfo   = (KMOD_INFO_64_V1 *)((UINT8 *) MachoGetMachHeader64 (&Context->MachContext) + KmodOffset);
-  StartAddr  = KcFixupValue (KmodInfo->StartAddr, NULL);;
+  KmodOffset = Context->VirtualKmod - Context->VirtualBase + Context->FileOffset;
+  KmodInfo   = (UINT8 *) MachoGetMachHeader (&Context->MachContext) + KmodOffset;
```
vit9696, line 369: This looks wrong as it needs to be done after the two checks on `KmodOffset` and `TmpOffset`. (It was wrong in the past as well.)

Library/OcAppleKernelLib/KextPatcher.c
```diff
@@ -282,23 +365,28 @@ PatcherBlockKext (
...
-  if (OcOverflowAddU64 (KmodOffset, sizeof (KMOD_INFO_64_V1), &TmpOffset)
+  if (OcOverflowAddU64 (KmodOffset, Context->Is32Bit ? sizeof (KMOD_INFO_32_V1) : sizeof (KMOD_INFO_64_V1), &TmpOffset)
     || TmpOffset > MachoGetFileSize (&Context->MachContext)
     || StartAddr == 0
     || Context->VirtualBase > StartAddr) {
     return EFI_INVALID_PARAMETER;
   }
 
-  TmpOffset = StartAddr - Context->VirtualBase;
+  TmpOffset = StartAddr - Context->VirtualBase + Context->FileOffset;
```
vit9696, line 384: This needs overflow checks now as adding `Context->FileOffset` may overflow in theory.

PR #119, opened 2020-09-14
Deactivate EC

mbarbierato:
If the _STA method exists in DSDT origin the EC0 or H_EC doesn't deactivate, with this is deactivate


mbarbierato:
I've tried this solution on my Intel NUC8i3BEH to deactivate the H_EC

Devil-66:
that's right, even in some Dell all in one, and it works! With @mbarbierato we're on the team @Hackintoshlifeit

Docs/AcpiSamples/SSDT-EC-USBX.dsl
```diff
@@ -35,8 +35,9 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "SsdtEC", 0x00001000)
      */
 
     /**
+    
     External (_SB_.PCI0.LPCB.EC0, DeviceObj)
-
+
```
vit9696, line 40: Why did you change whitespace here and above? Please avoid.

Docs/AcpiSamples/SSDT-EC-USBX.dsl
```diff
@@ -51,7 +52,20 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "SsdtEC", 0x00001000)
             }
         }
     }
-    **/
+    
+    if the STA method exists, use
+    
+    External (_SB_.PCI0.LPCB.EC0_._STA, UnknownObj)
```
vit9696, line 58: Better make this a separate comment somewhat like:
```C
/*
 * Uncomment the following to disable EC if the STA method exists.
 */

/**

…

 **/
```

Devil-66: ok, I'm working on ssdt plug, let's make a new request? or do I update you here?

Devil-66: the change you propose is ok, it reads better, easier to understand

Docs/AcpiSamples/SSDT-EC-USBX.dsl
mbarbierato: ok

Devil-66:
tested on more than 10 hacks

Devil-66:
I've already tried what you say, it doesn't work, I wanted to simplify, but the darwin is important and it has to be defined, on every cpu path, and using dsm as in the past,  test ssdt plug.

armenio:
> I've already tried what you say, it doesn't work, I wanted to simplify, but the darwin is important and it has to be defined, on every cpu path, and using dsm as in the past, test ssdt plug.

review the suggestions carefully because there’s no reason why they don’t work.
did you understand the exchange of multiple "if" by logical operator "and" just to simplify?

Devil-66:
edit and update 😘

Devil-66:
not working, rest on the current one, joining breaks something, I have to study it, I stay with this current with darwin. thanks for 😘 advice

armenio:
> not working, rest on the current one, joining breaks something, I have to study it, I stay with this current with darwin. thanks for 😘 advice

Can you post your test files/results?

Look for Darwin searching form **&& _OSI ("Darwin")**

of course, this should be used in every possible Device Scope

Devil-66:
Schermata 2020-09-15 alle 00 09 02
Schermata 2020-09-15 alle 00 09 02
see not working , with method written yourself

armenio:
> Schermata 2020-09-15 alle 00 09 02
> Schermata 2020-09-15 alle 00 14 19
> 
> see not working , with method written yourself

OK. Don't worry. Maybe I do that in a PR later

Devil-66:
to disable EC controllers on the desktop pc in some cases EC0 in other H_EC you have to use the method inserted in ssdt ec and ec usbx, I wrote them and tested them with my staff, see the nuc of @mbarbierato where I deactivated H_EC, then I added _STA with darwin in some ssdt, and I modified ssdt plug with darwin, so it does not affect windows and its controls that has on the cpu, and ssdt plug l I tested on multiple machines, including those with PR cpu. PR. PR00 and SB. PR00 and works this latest version, the ssdt versions I made with darwin on these CPUs do not work.
a greeting to the whole acidanthera team, it would be... it's would be an honor to work for you.

Docs/AcpiSamples/SSDT-EC-USBX.dsl
```diff
@@ -80,6 +97,17 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "SsdtEC", 0x00001000)
                     0x0834
                 })
             }
+            Method (_STA, 0, NotSerialized)  // _STA: Status
```
vit9696, line 100: The indentation s wrong.

Docs/AcpiSamples/SSDT-EC-USBX.dsl
```diff
@@ -80,6 +97,17 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "SsdtEC", 0x00001000)
...
+        {
+            If (_OSI ("Darwin"))
+            {
+                Return (0)
```
vit9696, line 104: This looks wrong, I believe you wanted to enable `USBX` on Darwin, not disable it.

Docs/AcpiSamples/SSDT-IMEI.dsl
```diff
@@ -11,6 +11,17 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "IMEI", 0x00000000)
         Device (IMEI)
         {
             Name (_ADR, 0x00160000)  // _ADR: Address
+            Method (_STA, 0, NotSerialized)  // _STA: Status
+        {
+            If (_OSI ("Darwin"))
+            {
+                Return (0)
```
vit9696, line 18: Same here.

Docs/AcpiSamples/SSDT-PLUG.dsl
```diff
@@ -20,24 +20,26 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "CpuPlug", 0x00003000)
...
-       })
-    }
-
     If (CondRefOf (\_SB.CPU0)) {
         If ((ObjectType (\_SB.CPU0) == 0x0C)) {
             Scope (\_SB.CPU0) {
-                Method (_DSM, 4, NotSerialized)  
-                {
-                    Return (PMPM (Arg0, Arg1, Arg2, Arg3))
+                If (_OSI ("Darwin")) {
```
vit9696, line 26: Is it any harmful? I am actually unsure there are any examples of these methods inducing harm in other systems.

Docs/AcpiSamples/SSDT-PLUG.dsl
```diff
@@ -20,24 +20,26 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "CpuPlug", 0x00003000)
...
+                    Method (_DSM, 4, NotSerialized)  // _DSM: Device-Specific Method
```
vit9696, line 27: Please avoid the copy-pasting, using `PMPM` method is cleaner to read.

Docs/AcpiSamples/SSDT-SBUS-MCHC.dsl
```diff
@@ -11,6 +11,17 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "MCHCSBUS", 0x00000000)
         Device (MCHC)
         {
             Name (_ADR, Zero)  // _ADR: Address
+            Method (_STA, 0, NotSerialized)  // _STA: Status
+        {
+            If (_OSI ("Darwin"))
+            {
+                Return (0)
```
vit9696, line 18: Same here.

Docs/AcpiSamples/SSDT-SBUS-MCHC.dsl
```diff
@@ -11,6 +11,17 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "MCHCSBUS", 0x00000000)
         Device (MCHC)
         {
             Name (_ADR, Zero)  // _ADR: Address
+            Method (_STA, 0, NotSerialized)  // _STA: Status
+        {
```
vit9696, line 15: Indentation is wrong here too.

Docs/AcpiSamples/SSDT-SBUS-MCHC.dsl
```diff
@@ -39,6 +50,17 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "MCHCSBUS", 0x00000000)
                 })
             }
         }
+        Method (_STA, 0, NotSerialized)  // _STA: Status
+        {
+            If (_OSI ("Darwin"))
+            {
+                Return (0)
```
vit9696, line 57: Same here.

Docs/AcpiSamples/SSDT-PLUG.dsl
Devil-66: starting windows from opencore, you noticed or temp high cpu and fans turn fast, with darwin method ssdt plug, then start windows from oc and temp low and fans low regimes

Docs/AcpiSamples/SSDT-SBUS-MCHC.dsl
Devil-66: repaired, sorry I did it at night 😇🙈

Docs/AcpiSamples/SSDT-PLUG.dsl
vit9696: weird, but ok.

Docs/AcpiSamples/SSDT-SBUS-MCHC.dsl
```diff
@@ -8,9 +8,12 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "MCHCSBUS", 0x00000000)
 
     Scope (_SB.PCI0)
     {
+    If (_OSI ("Darwin"))
```
vit9696, line 11: This is way too fugly. Try to avoid conditional declarations as they may be harder to handle for the ACPI implementation. Better use STA method here and elsewhere.

Devil-66: on mchc and imei better the method _sta? as you like ok 💪🙈

vit9696: Well, the issue is rather conceptual. The device exists, it is just that you do not want to use it e.g. in Windows.

dreamwhite:
Excuse me but what's the purpose of this PR, aside the amount of formatting errors? Basically it adds a "If _OSI("Darwin") which may not be needed if you boot only macOS. Am I wrong? @vit9696

1alessandro1:
Is this a sport or something @Devil-66?
Why are you making this changes since Windows/Linux etc do not care about plugin-type=1?

dreamwhite:
> Is this a sport or something @Devil-66?
> Why are you making this changes since Windows/Linux etc do not care about plugin-type=1?

Since he is a hackintosh master, with all of his knowledge he stated that if you don't use "_OSI("Darwin")" conditional, the CPU temps raises up as long as the fans. That doesn't make much sense but okay.

1alessandro1:
Nobody ever reported this problem, only EC was known to cause some issues in Windows, hence the contidional there.

Docs/AcpiSamples/SSDT-IMEI.dsl
```diff
@@ -11,6 +11,17 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "IMEI", 0x00000000)
         Device (IMEI)
         {
             Name (_ADR, 0x00160000)  // _ADR: Address
+            Method (_STA, 0, NotSerialized)  // _STA: Status
+        {
```
vit9696, line 15: Please indent this properly.

Docs/AcpiSamples/SSDT-ALS0.dsl
```diff
@@ -7,6 +7,8 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "ALS0", 0x00000000)
 {
     Scope (_SB)
     {
+    If (_OSI ("Darwin"))
```
vit9696, line 10: Please revert this and leave `_STA` as before.

Docs/AcpiSamples/SSDT-PLUG.dsl
```diff
@@ -35,99 +35,117 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "CpuPlug", 0x00003000)
...
+              }
             }
+          }
         }
-    }
     
     If (CondRefOf (\_PR.CPU0)) {
         If ((ObjectType (\_PR.CPU0) == 0x0C)) {
             Scope (\_PR.CPU0) {
+                If (_OSI ("Darwin")) {
```
vit9696, line 51: Please fix the indentation.

Docs/AcpiSamples/SSDT-SBUS-MCHC.dsl
```diff
@@ -7,11 +7,22 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "MCHCSBUS", 0x00000000)
     External (_SB_.PCI0.SBUS, DeviceObj)
 
     Scope (_SB.PCI0)
-    {
+     {
```
vit9696, line 10: Please fix the indentation.

vit9696 (commented):
The formatting of the files is really bad. Please carefully address this.

Docs/AcpiSamples/SSDT-IMEI.dsl
```diff
@@ -11,6 +11,17 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "IMEI", 0x00000000)
         Device (IMEI)
         {
             Name (_ADR, 0x00160000)  // _ADR: Address
+            Method (_STA, 0, NotSerialized)  // _STA: Status
+        	{
```
vit9696, line 15: This one is not resolved.

Docs/AcpiSamples/SSDT-EC-USBX.dsl
```diff
@@ -53,8 +53,27 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "SsdtEC", 0x00001000)
...
+            \_SB.PCI0.LPCB.EC0._STA = Zero
+        }
+        
+    }    
+    **/
+
     Scope (\_SB)
     {
+    If (_OSI ("Darwin"))
+        {
```
vit9696, line 76: This one is not resolved.

Docs/AcpiSamples/SSDT-SBUS-MCHC.dsl
```diff
@@ -11,7 +11,18 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "MCHCSBUS", 0x00000000)
         Device (MCHC)
         {
             Name (_ADR, Zero)  // _ADR: Address
+            Method (_STA, 0, NotSerialized)  // _STA: Status
+        {
+            If (_OSI ("Darwin"))
```
vit9696, line 16: This one is not resolved.

Docs/AcpiSamples/SSDT-IMEI.dsl
mbarbierato: Sorry

Docs/AcpiSamples/SSDT-EC-USBX.dsl
mbarbierato: Sorry

Docs/AcpiSamples/SSDT-SBUS-MCHC.dsl
mbarbierato: Sorry

Docs/AcpiSamples/SSDT-EC-USBX.dsl
```diff
@@ -53,32 +53,52 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "SsdtEC", 0x00001000)
...
+            \_SB.PCI0.LPCB.EC0._STA = Zero
+        }
+        
+    }    
+    **/
+
     Scope (\_SB)
     {
-        Device (USBX)
+        If (_OSI ("Darwin"))
```
vit9696, line 75: All good, but this check should be in _STA.

Devil-66: ok

Sniki:
On native Apple Hardware they have OSDW as a method to emulate Darwin.
@osy86 create an example: https://github.com/osy86/HaC-Mini/blob/master/ACPI/SSDT-OsxDetect.asl
Would it be cleaner to make a SSDT-OSX or whatever name you prefer to come up with and simply add the IF (OSDW) check in all SSDT(s) like in @osy86 ACPI folder example of his [repo](https://github.com/osy86/HaC-Mini/tree/master/ACPI)

dreamwhite:
> On native Apple Hardware they have OSDW as a method to emulate Darwin.
> @osy86 create an example: https://github.com/osy86/HaC-Mini/blob/master/ACPI/SSDT-OsxDetect.asl
> Would it be cleaner to make a SSDT-OSX or whatever name you prefer to come up with and simply add the IF (OSDW) check in all SSDT(s) like in @osy86 ACPI folder example of his [repo](https://github.com/osy86/HaC-Mini/tree/master/ACPI)

that's a much cleaner option than adding unnecessary "If _OSI("Darwin")" in my honest opinion.
What does @vit9696 says about this? Is it better to add to every single SSDT the Darwin check or the SSDT-OSX method?

vit9696:
Makes very little sense as OSI is described in the spec and is clear for anyone.

dreamwhite:
> Makes very little sense as OSI is described in the spec and is clear for anyone.

Alright but at this point I have a question: let's suppose that a user has only a working macOS installation on his machine without any other OS. It doesn't make any difference if the SSDTs contains or not the _OSI method since he only boots macOS, right?

vit9696:
Well, how does this relate to OSI/OSDW choice?

dreamwhite:
> Well, how does this relate to OSI/OSDW choice?

Correct me if I'm wrong: since the "John Doe" user (who boots only macOS with his PC) uses some SSDTs (e.g. SSDT-EC-USBX and SSDT-PLUG) there should be no problems of any kind if the OSI/OSDW method is used or not in those SSDTs, since he boots only macOS. Am I wrong?

vit9696:
Well, in that sense it does not matter indeed. Though, detecting the OS is cleaner of course. I only meant that OSDW to simply wrap OSI makes no sense as it is not really shorter, and OSI is easier to recognise.

Devil-66:
> Well, in that sense it does not matter indeed. Though, detecting the OS is cleaner of course. I only meant that OSDW to simply wrap OSI makes no sense as it is not really shorter, and OSI is easier to recognise.


I think the same way "detecting the OS is cleaner of course."

dreamwhite:
Just a stupid question: why [line 61](https://github.com/acidanthera/OpenCorePkg/blob/master/Docs/AcpiSamples/SSDT-EC-USBX.dsl#L61) of SSDT-EC-USBX has `UnknownObj` while we already know that _STA method is a MethodObj?

vit9696:
Yes, I am actually reviewing this at the moment. This hack may actually fail badly on many systems. Can those whom it works for upload the ACPI patches? Because on our systems `_STA` is a method, and you cannot write `Zero` to a method, only to a variable.

Devil-66:
> Yes, I am actually reviewing this at the moment. This hack may actually fail badly on many systems. Can those whom it works upload the ACPI patches? Because on our systems `_STA` is a method, and you cannot write `Zero` to a method, only to a variable.

sorry I'm late, I was working.
I leave you the final version to disable EC0, H_EC or whatever it is

    External (_SB_.PCI0.LPCB.EC0_._STA, UnknownObj)


Scope (\)
    {
        Method (_INI, 0, NotSerialized)  // _INI: Initialize
        {
            If (_OSI ("Darwin"))
            {
                Store (Zero, \_SB.PCI0.LPCB.EC0._STA)
            }
        }
    }

today had friends tested, this works great, evaluate what you want to do, thank you for reading me, await news.

mbarbierato:
> Yes, I am actually reviewing this at the moment. This hack may actually fail badly on many systems. Can those whom it works upload the ACPI patches? Because on our systems `_STA` is a method, and you cannot write `Zero` to a method, only to a variable.

ok, you can also revert ssdt-ec, it's the same thing.
We will carry out other tests to disable this if the _STA method is present in DSDT.
if you agree we will make another pull request with other ssdts we are working on.

Thank you

dreamwhite:
> Yes, I am actually reviewing this at the moment. This hack may actually fail badly on many systems. Can those whom it works upload the ACPI patches? Because on our systems `_STA` is a method, and you cannot write `Zero` to a method, only to a variable.

@vit9696 excuse me but for avoiding ACPI errors, can't we just check the `ObjectType` of `_STA` and decide if it's a method (so returning a value) or a variable and consequently store the value as they did?

vit9696:
If it is a method, you cannot update it (at all), so the workaround you previously submitted cannot (and did not) work on most implementations, including all the ones we tested it on. Initially I was under the assumption that macOS ACPI implementation has a bug, causing implicit method overriding (somewhat similar to [Linux](https://www.kernel.org/doc/html/latest/firmware-guide/acpi/method-customizing.html) or opcode patching in this situation, but neither is the case actually.

There are tools that generate EC patches (e.g. [SSDTTime](https://github.com/corpnewt/SSDTTime)), which is probably the way to go when dealing with EC.

1alessandro1:
@vit9696 I've just a little doubt, but I've heard that sometimes this way of implementing the _OSI method could fail (even if the compiler doesn't complain about that). 

`DefinitionBlock ("", "SSDT", 2, "ACDT", "CpuPlug", 0x00003000)
  {
    If(_OSI("Darwin")) 
    { 
      // SSDT contents
    }
  }
`

Here is an example:

image

Of course the code should be intended correctly (at the bottom there's the other curly bracket)

It isn't clear to me why on some instances that `If(_OSI("Darwin"))` method can behave weirdly

May you please explain why is that?

vit9696:
Not that I am aware of.

1alessandro1:
Thanks, so what's in the example isn't improperly written ACPI code, right?

vit9696:
At the very least this does look ACPI-conformant. Writing conditions outside the object scope may not be the best idea for older fragile implementations, but it should work in all modern operating systems.

PR #124, opened 2020-09-27
OpenCorePlatform: Add `SystemMemoryStatus` option to `Generic`

al3xtjames:
This allows `PT_FEATURE_HAS_SOLDERED_SYSTEM_MEMORY` to be set when using automatic PlatformInfo generation. Does it make sense to have an `Auto` option, or should we just make this a boolean (like AdviseWindows)?


vit9696:
We have already done it once, and it did not work. The reason for it is because almost every model name with soldered memory is hardcoded in the operating system framework and will not change the displayed tab in about system. Use https://github.com/Goldfish64/SystemProfilerMemoryFixup instead.

al3xtjames:
Huh, I must have missed the previous attempt. It appears to work with MBP14,x and 15,x, but I guess this isn't needed with SystemProfilerMemoryFixup.

vit9696:
Hmm. If we can make the statistics where it works and where it does not we could consider merging it, but it will require a clear table in the documentation. This might be important, because userspace patcher is not ready for Big Sur yet and we do not have resources to work on it.

The previous attempt was here: https://github.com/acidanthera/OpenCorePkg/commit/f654218e0c751b2391fa36e44d87b5f6f6d4c4b3#diff-f5c935ab63596eba1212f2823153290b

al3xtjames:
I took another look at SPMemoryReporter in 10.15.7, it doesn't look like much has changed. The only hardcoding appears to be for product names beginning with MacBookAir and MacBookPro10; otherwise, the value from `platform-feature` is used. I confirmed that the tab unconditionally disappears if MacBookAirx,x or MacBookPro10,x are used. `PT_FEATURE_HAS_SOLDERED_SYSTEM_MEMORY ` should be respected otherwise (I tried with MacBookPro13,x-15,x and the memory tab showed up).

Include/Acidanthera/Library/OcConfigurationLib.h
```diff
@@ -417,7 +417,8 @@ typedef enum {
   _(UINT16                      , ProcessorType      ,     , 0                                            , () )                   \
   _(UINT8                       , Rom                , [6] , {0}                                          , () )                   \
   _(BOOLEAN                     , SpoofVendor        ,     , FALSE                                        , () )                   \
-  _(BOOLEAN                     , AdviseWindows      ,     , FALSE                                        , () )
+  _(BOOLEAN                     , AdviseWindows      ,     , FALSE                                        , () )                   \
+  _(OC_STRING                   , SystemMemoryStatus ,     , OC_STRING_CONSTR ("Auto", _, __)             , OC_DESTR (OC_STRING) )
```
vit9696, line 421: Move this above to reduce padding.

Docs/Configuration.tex
```diff
@@ -4335,6 +4335,24 @@ \subsection{Generic Properties}\label{platforminfogeneric}
...
+\item
+  \texttt{SystemMemoryStatus}\\
+  \textbf{Type}: \texttt{plist\ string}\\
+  \textbf{Failsafe}: \texttt{Auto}\\
+  \textbf{Description}: Indicates whether system memory is upgradable in \texttt{PlatformFeature}.
+  This controls the visibility of the Memory tab in About This Mac.
+
+  Valid values:
+
+  \begin{itemize}
```
vit9696, line 4347: Add notes, where it does not work when set to `Upgradable`.

Platform/OpenCore/OpenCorePlatform.c
```diff
@@ -34,6 +34,7 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 #include <Guid/AppleVariable.h>
 
 STATIC CHAR8    mCurrentSmbiosProductName[48];
+STATIC UINT32   mPlatformFeature = 0;
```
vit9696, line 37: Please make this local. Global variable for SMBIOS is only needed because of the usage outside of the function.

PR #125, opened 2020-09-29
Update SSDT-PLUG.dsl

teedy8023mars:


Pavo-IM:
I do not agree of removing all of the other CPUs in this PR, the purpose of using the `If (CondRefOf ()` is to have the capability of having multiple references to sources. If you only have one source there is no point in using `If (CondRefOf ()` at all.

Gengik84:
@teedy8023mars maybe you have done an involuntary PR?
otherwise what is the real motivation?

PR #127, opened 2020-10-03
macserial: Fix Issue #1195 (false-positive year warning for serials)

matis-schotte:
Fix https://github.com/acidanthera/bugtracker/issues/1195

macserial: Fix Issue #1195 (false-positive year warning for serials)

Also I really wanted to say thanks to you all guys, great project!! Wonderful 👍

### DIFF
--- a/Utilities/macserial/macserial.c
+++ b/Utilities/macserial/macserial.c
@@ -1,7 +1,8 @@
 //
 // Decode mac serial number
 //
-// Copyright (c) 2018 vit9696
+// Copyright (c) 2018-2020 vit9696
+// Copyright (c) 2020 Matis Schotte
 //
 
 #include <limits.h>
--- a/Utilities/macserial/macserial.c
+++ b/Utilities/macserial/macserial.c
@@ -355,7 +355,7 @@ static bool get_serial_info(const char *serial, SERIALINFO *info, bool print) {
     // Since year can be encoded ambiguously, check the model code for 2010/2020 difference.
     // Old check relies on first letter of model to be greater than or equal to H, which breaks compatibility with iMac20,2 (=0).
     // Added logic checks provided model years `AppleModelYear` first year greater than or equal to 2020.
-    if (AppleModelYear[info->modelIndex][0] >= 2020 || (info->decodedYear == 0 && info->model[0] >= 'H')) {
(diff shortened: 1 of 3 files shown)


Utilities/macserial/macserial.c
```diff
@@ -352,7 +353,9 @@ static bool get_serial_info(const char *serial, SERIALINFO *info, bool print) {
     // New encoding started in 2010.
     info->decodedYear = alpha_to_value(info->year[0], AppleTblYear, AppleYearBlacklist);
     // Since year can be encoded ambiguously, check the model code for 2010/2020 difference.
-    if (info->decodedYear == 0 && info->model[0] >= 'H') {
+    // Old check relies on first letter of model to be greater than or equal to H, which breaks compatibility with iMac20,2 (=0).
+    // Added logic checks provided model years `AppleModelYear` first year greater than or equal to 2020.
+    if (AppleModelYear[info->modelIndex][0] >= 2020 || (info->decodedYear == 0 && info->model[0] >= 'H')) {
```
vit9696, line 358: What if modelIndex remains negative at this step?

vit9696:
This is actually a duplicate of https://github.com/acidanthera/bugtracker/issues/1092. Also need to fix the python tool.

Utilities/macserial/macserial.c
matis-schotte: Right, thanks for the pointer! Would it suffice to change this to `(info->modelIndex != -1 && AppleModelYear[info->modelIndex][0] >= 2020)` ?

matis-schotte:
Alright, I cannot seem to find `update_products.py ` in the macserial project ... Also not a python expert, but should I try anyways in this scope?

Utilities/macserial/macserial.c
vit9696: Would rather use `>= 0` than `!= -1` just in case more negative values land. This is probably ok... My bigger concern is whether the generator will still work correctly.

vit9696:
It is the code that provides information to macserial:
https://github.com/acidanthera/OpenCorePkg/tree/master/AppleModels

matis-schotte:
I updated the PR and also include a fix for the AppleModels db project https://github.com/acidanthera/bugtracker/issues/1092

Please note that the python script is probably going to take a while to scan all 4-digit codes (0000 - ZZZZ = 37^4 = 1.874.161 HTTP requests ;) )

vit9696:
I actually tried to simplify this a little bit. Thanks a lot for working on this!

PR #131, opened 2020-10-03
Ability to run from any folder.

jief666:


dakanji:
Related to acidanthera/bugtracker#920

vit9696:
Well, this patch cannot be merged as is because of the basics — the quality of the patch is very low.

- The changes do not address Bootstrap.efi compatibility:
     - When OpenCore is launched via Bootstrap.efi, which is located within OpenCore directory, it uses `OPEN_CORE_DRIVER_PATH`. If we assume that the directory can be different, The code will need to be adapted to look into OpenCore in a relative (parent) folder first.
     - When OpenCore registers Bootstrap.efi in the booter options it uses `OPEN_CORE_BOOTSTRAP_PATH`, but that will also need to be updated if the directory can be changed.
- There are several other functional errors, I will mark them inline.
- The codestyle must follow the codestyle we have in OpenCore (refer to the development section, it is similar to EDK II).
    - Variable declarations should be at the top
    - Names should be capitilised & camel case
    - Variable initialisation should be made separately
    - Pointers should be checked for not being null explicitly (`!= NULL` instead of `!`)
    - Function calls should be indented with a space after the function name (e.g. `Func (…)`)
    - All the `if` clauses should be guarded with curly braces (even single line)
    - Comments should be surrounded by empty `//` above and below
    - etc.
- Unrelated changes like the version string should be contributed separately (we have an issue for that https://github.com/acidanthera/bugtracker/issues/384)
- The changes should be compiler-agnostic, i.e. you cannot use gcc/clang-specific attributes without a particular need, and if you do, you must ensure compatibility with all the other compilers by introducing preprocessor checks.
- The changes cause CI failure, click on the `Details` and read the reasons why they do.

Let's start with a simple question, what is the point? From what I understand the only use-case of this is on Macs trying to boot OpenCore from different locations. I can let this in at this point, but it will need to be properly written following all our guidelines and addressing all the issues first.

Platform/OpenCore/OpenCore.c
```diff
@@ -175,13 +185,43 @@ OcBootstrapRerun (
...
+
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "OC: Failed to locate loaded image - %r\n", Status));
+      return EFI_NOT_FOUND;
+    }
+
+    // Extract dirname from LoadedImage->FilePath
+    CHAR16* UnicodeDevicePath = NULL;
+    if ( LoadedImage->FilePath ) {
+      UnicodeDevicePath = ConvertDevicePathToText(LoadedImage->FilePath, FALSE, FALSE);
```
vit9696, line 204: Allocation failure should be checked here.

Platform/OpenCore/OpenCore.c
```diff
@@ -175,13 +185,43 @@ OcBootstrapRerun (
...
 
+
+    EFI_LOADED_IMAGE_PROTOCOL* LoadedImage = NULL;
+    Status = gBS->HandleProtocol (
+      gImageHandle,
+      &gEfiLoadedImageProtocolGuid,
+      (VOID **) &LoadedImage
+      );
+
+    if (EFI_ERROR (Status)) {
```
vit9696, line 196: Missing Loaded image protocol should not abort the startup, can always fallback to the hardcoded default path.

Platform/OpenCore/OpenCore.c
```diff
@@ -175,13 +185,43 @@ OcBootstrapRerun (
...
+    CHAR16* UnicodeDevicePath = NULL;
+    if ( LoadedImage->FilePath ) {
+      UnicodeDevicePath = ConvertDevicePathToText(LoadedImage->FilePath, FALSE, FALSE);
+      CHAR16* p = UnicodeDevicePath + StrLen(UnicodeDevicePath);
+      while ( p > UnicodeDevicePath  &&  *p != L'\\' ) {
+        p -= 1;
+      }
+      *p = 0;
+    }
+    // If not dirname is found (LoadedImage->FilePath can be NULL), use OPEN_CORE_ROOT_PATH as before.
```
vit9696, line 211: What if this is started from the driver root directory? It may not have a leading `\`. Also, I wonder if it is cleaner to introduce a new function similar to `strrchr` in OcStringLib instead of such a loop, though I will not force it.

Platform/OpenCore/OpenCore.c
```diff
@@ -175,13 +185,43 @@ OcBootstrapRerun (
...
+
     Status = OcStorageInitFromFs (
       &mOpenCoreStorage,
       FileSystem,
-      OPEN_CORE_ROOT_PATH,
+      UnicodeDevicePath,
       mOpenCoreVaultKey
       );
 
+    if ( UnicodeDevicePath ) FreePool (UnicodeDevicePath);
```
Download-Fritz, line 221: This may call FreePool() on a string literal (OPEN_CORE_ROOT_PATH).

jief666:
> Let's start with a simple question, what is the point?
Some PC (and VMWare) have bios(firmware) that can create boot entries to a specific file in a specific folder, not just /EFI/BOOT/Boot64.efi. I've also just made a "BootloaderChooser" (https://www.insanelymac.com/forum/topic/345336-bootloaderchooser/). I'm just so tired of having to rename OC folder to OCxxx, then OCyyy to OC, just to try something. And if it doesn't work, I have to swap it back.

I'm so surprised that you have hardcoded path when it's not necessary. Your OcStorage is made well to have a non hard-coded root and you just stopped like 15 lines away from being really path independent.

I'm leaving it to you. Feel free to take it and improving it to fit your coding conventions.

Platform/OpenCore/OpenCore.c
jief666: Oups, oups, just add that last minute... Very bad...
I'm so too used to not do Allocate and Free anymore.

vit9696:
Closing as will not fix.

Platform/OpenCore/OpenCore.c
```diff
@@ -157,6 +157,16 @@ OcMain (
...
 }
 
+/*
+ * Leave a signature that wouldn't be stripped in release version
+ */
+static const char* opencore_revision __attribute__((used)) = "OpenCore revision: " OPEN_CORE_VERSION;
+/*
+ * Leave a marker for BootloaderChooser, to prevent confusion (an efi file from a folder EFI\OCXXX that would use file from EFI\OC)
+ */
+static const char* path_independant __attribute__((used)) = "path_independant";
```
Download-Fritz, line 167: Are you going to bin-walk the entire driver for this string? Sorry, this sounds terrible. We have a custom DOS header signature for BOOTx64, I guess we could have one for OC too? It could include the version and you could check it against a minimum instead. CC @vit9696

vit9696: I intentionally did not include the version in OpenCore.efi, but in theory we could, as unlike `BOOTx64.efi` we do not need the exact match.

vit9696:
@dakanji if you need that feature, you could take over. I believe it will not be too hard to implement this properly, so I sort of give the green light.

Platform/OpenCore/OpenCore.c
```diff
@@ -175,13 +185,43 @@ OcBootstrapRerun (
...
+      DEBUG ((DEBUG_ERROR, "OC: Failed to locate loaded image - %r\n", Status));
+      return EFI_NOT_FOUND;
+    }
+
+    // Extract dirname from LoadedImage->FilePath
+    CHAR16* UnicodeDevicePath = NULL;
+    if ( LoadedImage->FilePath ) {
+      UnicodeDevicePath = ConvertDevicePathToText(LoadedImage->FilePath, FALSE, FALSE);
+      CHAR16* p = UnicodeDevicePath + StrLen(UnicodeDevicePath);
+      while ( p > UnicodeDevicePath  &&  *p != L'\\' ) {
```
Download-Fritz, line 206: *p is always '\0', this is dead code.

jief666:
It's too bad you don't see the value of being able to have more than one OC folder, being able to name them properly and being able to boot on them without having to rename them, thus avoid any ambiguity.
Thanks that have made OcStorage easy to change it's root.

Download-Fritz:
@jief666 Again, you and anyone else is invited to contribute the necessary changes. We have different priorities at the moment.

jief666:
> @jief666 Again, you and anyone else is invited to contribute the necessary changes. We have different priorities at the moment.

You've deleted your comment about dead code. I'm assuming I wasn't wrong on this one. Not like the FreePool of litteral (what a shame 👎 ).
I knew the OpenCore wouldn't take it anyway. That's why I didn't make a big effort. I just felt I need you to know that I'm proposing these version with my BootloaderChooser.
I too have different priorities.
So all's good for me.

Download-Fritz:
@jief666 
> You've deleted your comment about dead code.
Correct, my comment was based on misreading (if vs while) and because nobody replied yet, I decided to delete it.
> (what a shame  ).
I don't know what you think to imply here, but this is a PR and PRs are reviewed for bugs. If you are upset we do not pull bugged code and fix it ourselves, I'm sorry, but no OSS project works this way.
> I knew the OpenCore wouldn't take it anyway. That's why I didn't make a big effort.
Well, is it not a self-fulfilling prophecy if you intentionally put up no effort just to see your initial assumption validated? Afterall you got a detailed review, I don't know what else you expect us to do.

Good luck with your project.

jief666:
"What a shame" was for **me** having made that obvious bug of freeing a litteral. 
If you quote, put the whole sentence : "Not like the FreePool of litteral (what a shame 👎 )." And that nasty bug was me. So no, not upset you saw that bug. Quite the opposite.

I understand and respect the benefit of coding standards and rules. But because you know them well, maybe you can help me ? I did 75% of the work. Maybe you could add the comment and the test and the coding style you want... How would that takes you ? 30 minutes ? Isn't that collaborative work ?

Some will do it one day. Because it's obvious it's bad to hard-code. Don't agree ?

Download-Fritz:
@jief666 Sorry, the first part was a misunderstanding.

I don't think it's *obvious* hard-coding is bad, it's a pros vs cons game. For instance, it is a pro that multiple folders can be supported, yes.
However a drawback is that if UEFI is bugged (and we do know of old code such as Aptio IV and 32-bit HFSPlus to have bugs in this direction), this solution will not work. Of course one could say there is no loss in this or blame the FW, however OC is designed to be very clear and explicit in every aspect (code, docs, behaviour) - the behaviour is simply not well predictable or formalisable. "Looks for files in launch directory, except when the firmware is bugged, it then looks in a hardcoded directory" sounds quite awful to me. If we did not have Vault, I could also imagine attacks were possible this way.

This is not to say that it will not be accepted for this reason, I just want to outline the benefit vs cost of this change is not as clear as it seems at first glance.

Help in the OSS community generally does not work by taking unrefined stuff and finishing it (usually this would happen when forking), but by instructing on how to improve the work. This both gives the submiter a chance to improve their skill and experience, as well as takes load from the maintainers.

For this specific case, I'm happy to help with instructions because it may happen to help you out improving your code (style). However, I personally do not care whether or not this specific change lands, so I will not finish it.

dakanji:
> if you need that feature, you could take over.

Seems beyond my capabilities and don't need it any longer as I have a perfectly fine workaround as my [build script](https://gist.github.com/dakanji/be49c4e591f20ae5ef0701ca1392e621) patches the source to create up to three variants in one go.

While it might make you choke on your Ryazhenka, I now run four variants chainloaded from an [enhanced version of Refind](https://github.com/dakanji/RefindPlus) coincidentally based on initial improvements to that tool by @jief666 as follows:
1. Standard Variant that runs out of `/EFi/OC`
2. Alternative Variant that runs out of `/EFi/OC_ALT`
3. Development Variant that runs out of `/EFi/OC_Dev` for testing the current Git Head before release
4. Known Working  Variant that runs out of `/EFi/OC_WRK` which I maintain as a known working instance from v0.5.7

In summary, fine with how things are but would not complain if someone is able to implement it without making things worse.

matis-schotte:
Interesting discussion! For me I just use a USB stick to try different OC configurations, and if it breaks no problem just unplug it and then am back to the internal-storage based oc config.

eljojo:
very interesting thread indeed. I have my whole EFI folder under git version control, switch branches in a USB stick and local boot drive, and cherry pick commits. It's been quite handy to test upgrades.

usr-sse2:
> very interesting thread indeed. I have my whole EFI folder under git version control, switch branches in a USB stick and local boot drive, and cherry pick commits. It's been quite handy to test upgrades.

Time to port Git to UEFI :D
```
fs0:\EFI\OC\tools\git.efi reset --hard HEAD~1
```

PR #135, opened 2020-10-10
Docs: Fix 'Firmware' Pluralisation

dakanji:
Firmware, as with software and hardware, should not be pluralised.
This PR fixes instances in the docs (strays into fixing some other items as well)

Instances in code comments and similar to be considered later

### DIFF
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -132,7 +132,7 @@ and UEFI functionality. For these reasons, this document is available exclusivel
 and all other sources or translations of this document are unofficial and may
 contain errors.
 
-Third-party articles, utilities, books, and alike, may be more useful for a wider audience as
+Third-party articles, utilities, books, and similar, may be more useful for a wider audience as
 they could provide guide-like material. However, they are subject to their authors' preferences,
 tastes, misinterpretations of this document, and unavoidable obsolescence.
 In cases of using such sources, such as \href{https://dortania.github.io}{Dortania}'s
@@ -500,7 +500,7 @@ entries include:
   \texttt{BootProtect} for more details.
 \item
   \texttt{boot} \\
-  Duet bootstrap loader, which initialises UEFI environment on legacy BIOS firmwares
+  Duet bootstrap loader, which initialises UEFI environment on legacy BIOS firmware
   and loads \texttt{OpenCore.efi} similarly to other bootstrap loaders. Modern Duet
   bootstrap loader will default to \texttt{OpenCore.efi} on the same partition when
   present.
@@ -561,7 +561,7 @@ To install OpenCore reflect the
 \hyperref[configuration-structure]{Configuration Structure} described
 in the previous section on a EFI volume of a GPT partition. While
 corresponding sections of this document do provide some information
-in regards to external resources like ACPI tables, UEFI drivers,
+regarding external resources such as ACPI tables, UEFI drivers,
 or kernel extensions (kexts), completeness of the matter is out of
 the scope of this document. Information about kernel extensions may
 be found in a separate
@@ -671,13 +671,14 @@ all possible means.
 
 \subsection{Coding conventions}\label{configuration-conv}
 
-Just like any other project we have conventions that we follow during the development.
-All third-party contributors are highly recommended to read and follow the conventions
-listed below before submitting their patches. In general it is also recommended to firstly
-discuss the issue in \href{https://github.com/acidanthera/bugtracker}{Acidanthera Bugtracker}
-before sending the patch to ensure no double work and to avoid the patch being rejected.
+As with any other project, we have conventions that we follow during development.
+All third-party contributors are advised to adhere to the conventions listed below
+before submitting patches. To minimise abortive work and the potential rejection of
+submissions, third-party contributors should initially raise issues to the
+\href{https://github.com/acidanthera/bugtracker}{Acidanthera Bugtracker}
+for feedback before submitting patches.
 
-\textbf{Organisation}. The codebase is contained in \texttt{OpenCorePkg} repository,
+\textbf{Organisation}. The codebase is contained in the \texttt{OpenCorePkg} repository,
 which is the primary EDK II package.
 \begin{itemize}
 \tightlist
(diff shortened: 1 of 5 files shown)


Docs/Configuration.tex
```diff
@@ -1522,33 +1522,33 @@ \subsection{Quirks Properties}\label{booterpropsquirks}
...
-  call, which results in early boot crashes. This quirk workarounds the problem by
+  Some firmware access memory by virtual addresses after a \texttt{SetVirtualAddresses}
+  call, resulting in early boot crashes. This quirk workarounds the problem by
   performing early boot identity mapping of assigned virtual addresses to physical
   memory.
 
-  \emph{Note}: The necessity of this quirk is determined by early boot failures. Currently
-  new firmwares with memory protection support (like OVMF) do not support this quirk due to
+  \emph{Note}: The necessity of this quirk is determined by early boot failures. Currently,
+  new firmware with memory protection support (such as OVMF) do not support this quirk. See
```
vit9696, line 1551: does not I guess

Docs/Configuration.tex
```diff
@@ -5752,27 +5752,27 @@ \subsection{Output Properties}\label{uefioutputprops}
...
+  \emph{Note}: This option only applies to the \texttt{System} renderer.
 
 \item
   \texttt{ReplaceTabWithSpace}\\
   \textbf{Type}: \texttt{plist\ boolean}\\
   \textbf{Failsafe}: \texttt{false}\\
-  \textbf{Description}: Some firmwares do not print tab characters or even everything
-  that follows them, causing difficulties or inability to use the UEFI Shell builtin
+  \textbf{Description}: Some types of firmware do not print tab characters or everything
+  that follows them, causing difficulties in using the UEFI Shell's builtin
```
vit9696, line 5775: It is not a living being, is `'s` actually allowed?

Docs/Configuration.tex
```diff
@@ -6054,31 +6054,31 @@ \subsection{Quirks Properties}\label{uefiquirkprops}
   \textbf{Description}: Adds delay in microseconds after \texttt{EXIT\_BOOT\_SERVICES}
   event.
 
-  This is a very ugly quirk to circumvent "Still waiting for root device" message
-  on select APTIO IV firmwares, namely ASUS Z87-Pro, when using FileVault 2 in particular.
-  It seems that for some reason they execute code in parallel to \texttt{EXIT\_BOOT\_SERVICES},
-  which results in SATA controller being inaccessible from macOS. A better approach should be
-  found in some future. Expect 3-5 seconds to be enough in case the quirk is needed.
+  This is a very rough workaround to circumvent the "Still waiting for root device" message
```
vit9696, line 6057: Really need to fix that ``Still waiting for root device''.

Docs/Configuration.tex
```diff
@@ -6108,23 +6108,23 @@ \subsection{Quirks Properties}\label{uefiquirkprops}
...
   registers.
 
 \item
   \texttt{UnblockFsConnect}\\
   \textbf{Type}: \texttt{plist\ boolean}\\
   \textbf{Failsafe}: \texttt{false}\\
-  \textbf{Description}: Some firmwares block partition handles by opening them
-  in By Driver mode, which results in File System protocols being unable to install.
+  \textbf{Description}: Some types of firmware block partition handles by opening them
+  in \texttt{By\ Driver} mode, resulting in being unable to install File System Protocols.
```
vit9696, line 6127: Protocols should probably not be capitalised as it is not part of the name.

vit9696:
Thanks for that actually. I believed that firmware allows for plural form under some circumstances, but it still does not it seems. I marked some lines inline where I felt uncertain, but overall the changes look good.

Docs/Configuration.tex
Download-Fritz: There is no consistent rule, some people tend to avoid it, but it works.

Docs/Configuration.tex
Download-Fritz: The UEFI spec is a bit inconsistent, but generally uses capital case for UEFI Protocols. Lower case is usually used for network protocols etc., but not always.

vit9696:
Besides, you seem to consistently(?) use `some firmware`/`most firmware` as plural. I am not sure whether it is common as my natural choice with similar words like `software` will always be singular, i.e. `most software is affected` not `most software are affected`. Could you please confirm that this is intended?

dakanji:
> Besides, you seem to consistently(?) use some firmware/most firmware as plural. I am not sure whether it is common as my natural choice with similar words like software will always be singular, i.e. most software is affected not most software are affected. Could you please confirm that this is intended?

These words are a bit tricky but they do have a "plural" form which is inferred from the context when not properly "pluralised" by assigning groupings. The best way to consider these is like a liquid which is not countable but groupings of such are.  You can keep 30 barrels of beer in a warehouse but not 30 beers.

In summary, it is intended and the key is consistency. 

Your `Most software are affected` is "correct" in the same way as you can drink 30 beers on a night out. This is because it comes with an implied grouping and is understood to mean 30 pints/cans/bottles/litres of beer ... the unit grouping actually doesn't have to be specific defined.

Similarly, there is an implied "SomethingOrTheOther of" between "Most" and "Software" that doesn't strictly need to be defined.

In these situations though, I have tried to use "types of" as a consistent grouping but there are situations were i was not 100% sure whether the intent was the plural form or not.

Having said all that, I don't recollect specifically introducing such without adding "types of".

Docs/Configuration.tex
dakanji: Probably can go either way. Will drop to lowercase. 
I used uppercase because of "File System" ... Was the actual intention "Filesystem"

Docs/Configuration.tex
dakanji: Presume you mean to change to \texttt{Still\ waiting\ for\ root\ device}

Docs/Configuration.tex
dakanji: I think "do not" as it seems not talking about only one specific new firmware.

Docs/Configuration.tex
vit9696: Well, could be both. TeX smart quotes are inserted with ``sequences''.

Docs/Configuration.tex
```diff
@@ -3020,16 +3020,16 @@ \subsection{Boot Properties}\label{miscbootprops}
   \end{itemize}
 
   \emph{Note 1}: Activated \texttt{KeySupport}, \texttt{OpenUsbKbDxe}, or similar driver is required
-  for key handling to work. On many firmwares it is not possible to get all the keys function.
+  for key handling to work. On sevaral types of firmware, it is not possible to get all the key functions.
```
telepati, line 3023: Needs fix typo: "sevaral" to "several".

dakanji: Thanks. Will fix presently.

PR #136, opened 2020-10-12
Extend Firmware Pluralisation Fixes

dakanji:
Extends #135

### DIFF
--- a/Application/PavpProvision/PavpProvision.c
+++ b/Application/PavpProvision/PavpProvision.c
@@ -116,7 +116,7 @@ ReadProvisioningDataFile (
   }
 
   //
-  // Implement fallback for our firmwares.
+  // Implement fallback for our firmware.
   //
   if (EFI_ERROR (Status)) {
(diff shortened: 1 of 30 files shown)


Patches/0001-ShellPkg-Add-support-for-input-with-separately-repor.patch
```diff
@@ -6,12 +6,12 @@ Subject: [PATCH 1/1] ShellPkg: Add support for input with separately reported
 
 REF: https://bugzilla.tianocore.org/show_bug.cgi?id=2510
 
-Some firmwares:
+Some types of firmware:
 - Report Shift modifier even when they report upper-case unicode letter.
```
vit9696, line 10: You cannot edit patch files as they affect the files we do not own.

dakanji: Yes, Should not ... at least not to items not specifically being added at most.

PR #138, opened 2020-10-14
Convert to Darwin Version number

1alessandro1:


Andrey1970AppleLife:
The version macOS is more convenient for perception.

Andrey1970AppleLife:
I added Darwin Version and macOS Version  together.

1alessandro1:
> I added Darwin Version and macOS Version together.

That seems reasonable. Thank you

PR #141, opened 2020-10-19
add darwin to new ssdt UNC for x99

Lorys89:


Lorys89:
I would delete this line, but I await your opinion
        // yet on select pre-Windows 10 it may cause issues.

PMheart:
Thanks

dreamwhite:
May I gently ask you where does `PRBM` comes from? And why should we set it to `0`? 
May you provide more infos about this SSDT and how does it work?  It's for the sake of the community .-.
TIA

PR #143, opened 2020-10-22
OcConsoleLib: Improve Error Specificity in TextoutputBuiltin

dakanji:
Closes acidanthera/bugtracker#1238

Three separate conditions can feed `EFI_DEVICE_ERROR` back to `OcUseBuiltinTextOutput` which leads to some difficulty discerning which one was triggered when this error is triggered. This is more so when OpenCore is used as a 3rd party library. 
This PR adds specificity to the conditions.

TODO:
- [x] Rollback Initial Changes
- [x] Change RenderResync function to return different values for all cases
- [x] Change functions calling RenderResync to always return device error on its failure
- [x] Make a separate AsciiTextResetEx function that will take a Verbose param to log on failure
- [x] Include logging with Status printing for RenderResync.
- [x] Call AsciiTextResetEx from AsciiTextReset with FALSE.
- [x] Call AsciiTextResetEx from OcUseBuiltinTextOutput with TRUE.
- [x] Permit status specificity return to 3rd parties.

### DIFF
--- a/Library/OcConsoleLib/TextOutputBuiltin.c
+++ b/Library/OcConsoleLib/TextOutputBuiltin.c
@@ -367,7 +367,7 @@ RenderResync (
 
   if (Info->HorizontalResolution < TGT_CHAR_WIDTH * 3
     || Info->VerticalResolution < TGT_CHAR_HEIGHT * 3) {
-    return EFI_DEVICE_ERROR;
+    return EFI_LOAD_ERROR;
   }
 
   if (mCharacterBuffer != NULL) {
@@ -376,7 +376,7 @@ RenderResync (
 
   mCharacterBuffer = AllocatePool (TGT_CHAR_AREA * sizeof (mCharacterBuffer[0]));
   if (mCharacterBuffer == NULL) {
-    return EFI_DEVICE_ERROR;
+    return EFI_OUT_OF_RESOURCES;
   }
 
   mConsoleGopMode          = mGraphicsOutput->Mode->Mode;
@@ -406,9 +406,10 @@ RenderResync (
 STATIC
 EFI_STATUS
 EFIAPI
-AsciiTextReset (
+AsciiTextResetEx (
   IN EFI_SIMPLE_TEXT_OUTPUT_PROTOCOL *This,
-  IN BOOLEAN                         ExtendedVerification
+  IN BOOLEAN                         ExtendedVerification,
+  IN BOOLEAN                         Debug
   )
 {
   EFI_STATUS                            Status;
@@ -424,6 +425,11 @@ AsciiTextReset (
 
   if (EFI_ERROR (Status)) {
     gBS->RestoreTPL (OldTpl);
+    if (Debug) {
+      DEBUG ((DEBUG_INFO, "OCC: ASCII Text Reset [HandleProtocolFallback] - %r\n", Status));
+      return Status;
+    }
+
     return EFI_DEVICE_ERROR;
   }
 
@@ -433,13 +439,31 @@ AsciiTextReset (
   mForegroundColor.Raw     = mGraphicsEfiColors[ARRAY_SIZE (mGraphicsEfiColors) / 2 - 1];
 
   Status = RenderResync (This);
+  gBS->RestoreTPL (OldTpl);
   if (EFI_ERROR (Status)) {
-    gBS->RestoreTPL (OldTpl);
-    return Status;
+    if (Debug) {
+      DEBUG ((DEBUG_INFO, "OCC: ASCII Text Reset [RenderResync] - %r\n", Status));
+      return Status;
+    }
+
+    return EFI_DEVICE_ERROR;
   }
 
-  gBS->RestoreTPL (OldTpl);
   return EFI_SUCCESS;
+}
+
+STATIC
+EFI_STATUS
+EFIAPI
+AsciiTextReset (
+  IN EFI_SIMPLE_TEXT_OUTPUT_PROTOCOL *This,
+  IN BOOLEAN                         ExtendedVerification
+  )
+{
+  EFI_STATUS                            Status;
+
+  Status = AsciiTextResetEx (This, ExtendedVerification, FALSE);
+  return Status;
 }
 
 STATIC
@@ -470,7 +494,7 @@ AsciiTextOutputString (
     Status = RenderResync (This);
     if (EFI_ERROR (Status)) {
       gBS->RestoreTPL (OldTpl);
-      return Status;
+      return EFI_DEVICE_ERROR;
     }
   }
 
@@ -591,7 +615,7 @@ AsciiTextQueryMode (
     Status = RenderResync (This);
     if (EFI_ERROR (Status)) {
       gBS->RestoreTPL (OldTpl);
-      return Status;
+      return EFI_DEVICE_ERROR;
     }
   }
 
@@ -628,7 +652,7 @@ AsciiTextSetMode (
     Status = RenderResync (This);
     gBS->RestoreTPL (OldTpl);
     if (EFI_ERROR (Status)) {
-      return Status;
+      return EFI_DEVICE_ERROR;
     }
   }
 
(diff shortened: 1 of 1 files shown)


vit9696:
None of these error codes are specified in UEFI spec for Simple Text Out protocol. Furthermore allocation failures should usually report out of resources. Not sure there are enough reasons to merge this. CC @Download-Fritz

dakanji:
Can't say I fully understood what you wrote but I hope the intent/issue I laid out was clear.

Currently, if calling OcUseBuiltinTextOutput externally and you get EFI_DEVICE_ERROR back, it could be any one of three things. With this, it would only be one thing and it doesn't impact on OpenCorePkg so not sure what the issue is from OpenCore's side. 

Perhaps there is a belief it is somehow "wrong/not by the book" but this is debatable since it is not changing the return status of any internal Simple_Text_In/Out function but statuses from "procedural" code.

This would help a caller to decide on what next steps to take by knowing exactly what the failure is. I do concede there is no compelling reason from your side to merge this since the benefit may appear to you as solely accruing to 3rd party utilities.

vit9696:
The change modifies `RenderResync` function return codes, which are used in many Simple Text Output protocol functions in `TextOutputBuiltin.c`. For example, `AsciiTextReset` implementing `Reset`, `AsciiTextOutputString` implementing `OutputString`, and so on.

If you check the latest [UEFI Spec](https://uefi.org/sites/default/files/resources/UEFI%20Spec%202.8B%20May%202020.pdf), namely section `12.4 Simple Text Output Protocol`, which starts at page 443, you will see that `Reset` function can only return `EFI_DEVICE_ERROR` or `EFI_SUCCESS`, and `OutputString` can return the same codes and additionally `EFI_UNSUPPORTED` or `EFI_WARN_UNKNOWN_GLYPH`.

The suggest changes will make it possible for these functions to return `EFI_BAD_BUFFER_SIZE` or `EFI_PROTOCOL_ERROR`, however, none of these codes are allowed by the UEFI specification, i.e. they essentially violate it. Currently I am unconvinced we should violate the specification for an unclear reason as this function is not supposed to fail in the first place and upon receiving such an error we will unlikely proceed much further in the first place.

dakanji:
Fair enough. Best to just patch this as needed.

Bmju:
Reading this, +1 to more specific error codes when using OpenCore as a library - _if_ it's possible without violating the spec. @dakanji - I am assuming you must have real world cases where this was failing and it was hard to find out exactly why? Is it possible to describe briefly? (I'm just asking as an interested party, also hoping to use OpenCore as a library.)

Library/OcConsoleLib/TextOutputBuiltin.c
```diff
@@ -376,7 +376,7 @@ RenderResync (
 
   mCharacterBuffer = AllocatePool (TGT_CHAR_AREA * sizeof (mCharacterBuffer[0]));
   if (mCharacterBuffer == NULL) {
-    return EFI_DEVICE_ERROR;
+    return EFI_BAD_BUFFER_SIZE;
```
vit9696, line 379: Better EFI_OUT_OF_RESOURCES.

Library/OcConsoleLib/TextOutputBuiltin.c
```diff
@@ -423,6 +424,9 @@ AsciiTextReset (
     );
 
   if (EFI_ERROR (Status)) {
+    if (Verbose == TRUE) {
```
vit9696, line 427: For boolean types just use if (Verbose)

Library/OcConsoleLib/TextOutputBuiltin.c
dakanji: Noted

Library/OcConsoleLib/TextOutputBuiltin.c
dakanji: Noted

Library/OcConsoleLib/TextOutputBuiltin.c
```diff
@@ -424,6 +425,12 @@ AsciiTextReset (
 
   if (EFI_ERROR (Status)) {
     gBS->RestoreTPL (OldTpl);
+    if (Verbose) {
+      DEBUG ((DEBUG_INFO, "OCC: ASCII Text Reset HandleProtocolFallback - %r\n", Status));
+
+      return Status
```
vit9696, line 431: Why return Status here?

vit9696:
Looks generally good otherwise. Could you test it?

Library/OcConsoleLib/TextOutputBuiltin.c
dakanji: Related to Task 8 that I added.

As it stands, this should log things in OpenCore clearly but need a way for the 'raw' statuses to get to a 3rd party calling OcUseBuiltinTextOutput. I also need to add the same to the other verbose if block to achieve this.

OpenCore does not handle the status from OcUseBuiltinTextOutput so should not be an issue on that front.

OK with you?

dakanji: Changes approach to leave internal OpenCorePkg calls unchanged. External calls to can be made to new OcUseBuiltinTextOutputEx Function if raw status is required.

Will build and test.

Library/OcConsoleLib/TextOutputBuiltin.c
```diff
@@ -919,8 +948,10 @@ ConsoleControlInstall (
 }
 
 EFI_STATUS
-OcUseBuiltinTextOutput (
```
vit9696, line 922: I do not really think we need two functions, just call AsciiTextResetEx with TRUE, TRUE in any case. Also, since it is basically always called with TRUE+TRUE or FALSE+FALSE the argument could be one, e.g. Debug.

dakanji: Sorry, a bit lost on that.

In AsciiTextReset we call:
`Status = AsciiTextResetEx (This, ExtendedVerification, false, FALSE);`

OcUseBuiltinTextOutputEx, when called from OcUseBuiltinTextOutput, calls:
`Status = AsciiTextResetEx (&mAsciiTextOutputProtocol, true, true, FALSE);`

The last, `RawStatus`, param is always `FALSE` on the normal internal calls using `AsciiTextReset` and `OcUseBuiltinTextOutput` to keep everything working as you wanted (logs the real status on error but returns `EFI_DEVICE_ERROR`). 

This works fine for doing everything in OpenCore but needed a bit more.

OcUseBuiltinTextOutputEx, when called directly, can call:
`Status = AsciiTextResetEx (&mAsciiTextOutputProtocol, true, true, TRUE);`

The `RawStatus` param can be set to `TRUE` to have the raw/real status returned instead of `EFI_DEVICE_ERROR`.

Not sure how this can be achieved in another way without adding an additional param to `OcUseBuiltinTextOutput` and editing all instances where it is called.

dakanji: > Not sure how this can be achieved in another way without adding an additional param to OcUseBuiltinTextOutput and editing all instances where it is called.

Actually the complication is the optional verbose logging. If you are happy for calls to `AsciiTextResetEx` from both `AsciiTextReset` and `OcUseBuiltinTextOutput` to have verbose logging, then this end can be achieved without an additional function.

vit9696: It does not matter what the AsciiTextResetEx and OcUseBuiltinTextOutput functions return, since they are for private use. AsciiTextReset on the other side must adhere to the protocol interface AND it cannot print due to a possibility of a recursion. For this reason I suggested to:

* drop OcUseBuiltinTextOutputEx and edit OcUseBuiltinTextOutput instead making it behave as OcUseBuiltinTextOutputEx(Mode, TRUE, TRUE) currently.
* edit AsciiTextResetEx to take only one extra parameter (Debug) that will enable both verbose logging and raw error code.
* edit OcUseBuiltinTextOutput to call AsciiTextResetEx with Debug = TRUE and let the other functions call it with FALSE.

This looks best to me, sorry for being unclear earlier.

PR #146, opened 2020-10-27
Use this to Fix IRQ Flags in HPET, TIMR, IPIC and RTC

andreszerocross:
Use this to Fix IRQ Flags in HPET, TIMR, IPIC and RTC

This SSDT will set your HPET to Enable / Disable (Enable by default), and remove IRQFlag from TIMR, IPIC and RTC and add "0,8, 11, 15" as IRQFlags to HPET.

This will Fix Audio and other problem for Broadwell and below.

### DIFF
--- a/Docs/AcpiSamples/SSDT-IRQFix.dsl
+++ b/Docs/AcpiSamples/SSDT-IRQFix.dsl
@@ -1,0 +1,203 @@
+/*
+ * Intel ACPI Component Architecture
+ * AML/ASL+ Disassembler version 20180427 (64-bit version)(RM)
+ * Copyright (c) 2000 - 2018 Intel Corporation
+ * 
+ * Disassembling to non-symbolic legacy ASL operators
+ *
+ * Disassembly of iASLpzP6ix.aml, Tue Oct 27 14:27:40 2020
+ *
+ * Original Table Header:
+ *     Signature        "SSDT"
+ *     Length           0x0000028D (653)
+ *     Revision         0x02
+ *     Checksum         0xFD
+ *     OEM ID           "ACDT "
+ *     OEM Table ID     "IRQFix"
+ *     OEM Revision     0x00001000 (4096)
+ *     Compiler ID      "INTL"
+ *     Compiler Version 0x20180427 (538444839)
+ */
+DefinitionBlock ("", "SSDT", 2, "ACDT ", "IRQFix", 0x00001000)
+{
+    External (_SB_.PCI0.LPCB.HPET._CRS, UnknownObj)    // (from opcode)
+    External (_SB_.PCI0.LPCB.HPET._STA, UnknownObj)    // (from opcode)
+    External (_SB_.PCI0.LPCB.IPIC._CRS, UnknownObj)    // (from opcode)
+    External (_SB_.PCI0.LPCB.RTC_._CRS, UnknownObj)    // (from opcode)
+    External (_SB_.PCI0.LPCB.TIMR._CRS, UnknownObj)    // (from opcode)
+
+    If (_OSI ("Darwin"))
+    {
+        Device (DEAX)
+        {
+            Name (_HID, "DEA00000")  // _HID: Hardware ID
+            Method (_STA, 0, NotSerialized)  // _STA: Status
+            {
+                If (_OSI ("Darwin"))
+                {
+                    Return (0x0F)
+                }
+                Else
+                {
+                    Return (Zero)
+                }
+            }
+
+            Method (_INI, 0, NotSerialized)  // _INI: Initialize
+            {
+                /* 
+                    * You can choose "Store (Zero, \_SB.PCI0.LPCB.HPET._STA) or Store (0x0F, \_SB.PCI0.LPCB.HPET._STA)"
+                    * to Disable or Enable HPET, "Store (0x0F, \_SB.PCI0.LPCB.HPET._STA)' is used as default (enabled HPET)
+                */
+                /* 
+                Store (Zero, \_SB.PCI0.LPCB.HPET._STA)
+                */
+                Store (0x0F, \_SB.PCI0.LPCB.HPET._STA)
+                Store (DEA1, \_SB.PCI0.LPCB.HPET._CRS)
+                Store (DEA2, \_SB.PCI0.LPCB.RTC._CRS)
+                Store (DEA3, \_SB.PCI0.LPCB.TIMR._CRS)
+                Store (DEA4, \_SB.PCI0.LPCB.IPIC._CRS)
+            }
+
+            Name (DEA1, ResourceTemplate ()
+            {
+                IRQNoFlags ()
+                    {0,8, 11, 15}
+                Memory32Fixed (ReadWrite,
+                    0xFED00000,         // Address Base
+                    0x00000400,         // Address Length
+                    )
+            })
+            Name (DEA2, ResourceTemplate ()
+            {
+                IO (Decode16,
+                    0x0070,             // Range Minimum
+                    0x0070,             // Range Maximum
+                    0x01,               // Alignment
+                    0x08,               // Length
+                    )
+            })
+            Name (DEA3, ResourceTemplate ()
+            {
+                IO (Decode16,
+                    0x0040,             // Range Minimum
+                    0x0040,             // Range Maximum
+                    0x01,               // Alignment
+                    0x04,               // Length
+                    )
+                IO (Decode16,
+                    0x0050,             // Range Minimum
+                    0x0050,             // Range Maximum
+                    0x10,               // Alignment
+                    0x04,               // Length
+                    )
+            })
+            Name (DEA4, ResourceTemplate ()
+            {
+                IO (Decode16,
+                    0x0020,             // Range Minimum
+                    0x0020,             // Range Maximum
+                    0x01,               // Alignment
+                    0x02,               // Length
+                    )
+                IO (Decode16,
+                    0x0024,             // Range Minimum
+                    0x0024,             // Range Maximum
+                    0x01,               // Alignment
+                    0x02,               // Length
+                    )
+                IO (Decode16,
+                    0x0028,             // Range Minimum
+                    0x0028,             // Range Maximum
+                    0x01,               // Alignment
+                    0x02,               // Length
+                    )
+                IO (Decode16,
+                    0x002C,             // Range Minimum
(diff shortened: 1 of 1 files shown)


andreszerocross:
Some ACPI use TMR instead TIMR, PIC instead IPIC, and RTC0 instead RTC. You only need to modify it

Ukr55:
What about different name something like LPC, LPC0, SBRG, maybe immediately do this.

Docs/AcpiSamples/SSDT-IRQFix.dsl
```diff
@@ -0,0 +1,203 @@
...
+            {
+                /* 
+                    * You can choose "Store (Zero, \_SB.PCI0.LPCB.HPET._STA) or Store (0x0F, \_SB.PCI0.LPCB.HPET._STA)"
+                    * to Disable or Enable HPET, "Store (0x0F, \_SB.PCI0.LPCB.HPET._STA)' is used as default (enabled HPET)
+                */
+                /* 
+                Store (Zero, \_SB.PCI0.LPCB.HPET._STA)
+                */
+                Store (0x0F, \_SB.PCI0.LPCB.HPET._STA)
+                Store (DEA1, \_SB.PCI0.LPCB.HPET._CRS)
```
vit9696, line 56: Are you sure you can assign any object like this?

Docs/AcpiSamples/SSDT-IRQFix.dsl
```diff
@@ -0,0 +1,203 @@
...
+            Method (_INI, 0, NotSerialized)  // _INI: Initialize
+            {
+                /* 
+                    * You can choose "Store (Zero, \_SB.PCI0.LPCB.HPET._STA) or Store (0x0F, \_SB.PCI0.LPCB.HPET._STA)"
+                    * to Disable or Enable HPET, "Store (0x0F, \_SB.PCI0.LPCB.HPET._STA)' is used as default (enabled HPET)
+                */
+                /* 
+                Store (Zero, \_SB.PCI0.LPCB.HPET._STA)
+                */
+                Store (0x0F, \_SB.PCI0.LPCB.HPET._STA)
```
vit9696, line 55: What if `_STA` is a method?

Docs/AcpiSamples/SSDT-IRQFix.dsl
```diff
@@ -0,0 +1,203 @@
...
+                IRQNoFlags ()
+                    {0,8, 11, 15}
+                Memory32Fixed (ReadWrite,
+                    0xFED00000,         // Address Base
+                    0x00000400,         // Address Length
+                    )
+            })
+            Name (DEA2, ResourceTemplate ()
+            {
+                IO (Decode16,
```
vit9696, line 73: Why do you delete IRQ from RTC? This may break RTC wake for instance? Same for other devices.

roddy20: i tried it with IRQ inside RTC and without IRQ, no difference found in RTC Wake, macOS assigns IRQ when needed

roddy20: Device (RTC)
                {
                    Name (_HID, EisaId ("PNP0B00") /* AT Real-Time Clock */)  // _HID: Hardware ID
                    Name (_CRS, ResourceTemplate ()  // _CRS: Current Resource Settings
                    {
                        IO (Decode16,
                            0x0070,             // Range Minimum
                            0x0070,             // Range Maximum
                            0x00,               // Alignment
                            0x02,               // Length
                            )
                    })
                }

Andrey1970AppleLife:
> Use this to Fix IRQ Flags in HPET, TIMR, IPIC and RTC

Don't touch the devices TIMR, IPIC and RTC.
Need only to add IRQFlags in HPET, there is enough of one 11.

roddy20:
> Need only to add IRQFlags in HPET, there is enough of one 11.

any chipset? or a specific one?

Andrey1970AppleLife:
> any chipset? or a specific one?

In most cases it isn't required even one. In most cases macOS assigns IRQ for HPET.
But for IVY and older it can to be demanded to add at least one 11.

roddy20:
> In most cases it isn't required even one.

macOS will start without it, but some problems with EHCI/XHCI and audio are possible
IRQNoFlags ()
                            {0,8,11,12} or {0,8,11,15} works Ok, but I dont know if it is the best solution (?)

Docs/AcpiSamples/SSDT-IRQFix.dsl
andreszerocross: No problem, in my Z170X Gaming 7 _STA of HPET is Method. And patch work normally,,

Andrey1970AppleLife:
> > In most cases it isn't required even one.
> 
> macOS will start without it, but some problems with EHCI/XHCI and audio are possible
> IRQNoFlags ()
> {0,8,11,12} or {0,8,11,15} works Ok, but I dont know if it is the best solution (?)

It is enough to assign of one, and macOS will assign still.

Docs/AcpiSamples/SSDT-IRQFix.dsl
andreszerocross: Yes, it work normally,

I have checked through ioreg in Windows and patch doesn't effect in Windows too.

andreszerocross:
> What about different name something like LPC, LPC0, SBRG, maybe immediately do this.

You just need to edit / modify LPCB to LPC/LPC0 or SBRG. It's easy

andreszerocross:
This patch is based on this one https://github.com/RehabMan/Laptop-DSDT-Patch/blob/master/system/system_IRQ.txt.

Long ago, for broadwell and below, we need this patch to got audio in Broadwell and below.

Most people with Broadwell and below system with Opencore need to patch DSDT.aml to patch IRQ or Disable HPET to get audio work with OpenCore, so i just upload my method with OpenCore.

Andrey1970AppleLife:
> Long ago, for broadwell and below, we need this patch to got audio in Broadwell and below.
> 
> Most people with Broadwell and below system with Opencore need to patch DSDT.aml to patch IRQ or Disable HPET to get audio work with OpenCore, so i just upload my method with OpenCore.

It isn't required to me on Haswell, but on IVY I need to add 11 that the sound worked.

andreszerocross:
> > Long ago, for broadwell and below, we need this patch to got audio in Broadwell and below.
> > Most people with Broadwell and below system with Opencore need to patch DSDT.aml to patch IRQ or Disable HPET to get audio work with OpenCore, so i just upload my method with OpenCore.
> 
> It isn't required to me on Haswell, but on IVY I need to add 11 that the sound worked.

I have installed in Asus A455LA haswell based. Need IRQ patch to make Audio work.

Honestly i always use 0, 8, 11, 15 for IRQFlag of HPET. I didn't make any experiment with other combination.

No IRQPatch = No audio in my Haswell Asus A455LA

Sniki:
For me patched HPET is required on Haswell ThinkPad Laptops like (L440, X240, T440, T440S, T440P) to have working audio. (alc-delay doesn't solve it in these laptops).

Also on Ivy Bridge on HP Elite 8300/6300 series, (Towers, Small Form Factors and USDT model)
For working audio as well.

roddy20:
> I need to add 11 that the sound worked.

changed to {11}
system assigned 11 12 14
seems to be working, will see ))

andreszerocross:
And for SkyLake notebook and above (Mobile chipset), i always disable HPET,, because i have seen in MacbookPro 13,1 and above there is no HPET in their IOREG.

And for Kabylake and above desktop PC, i  set disable for HPET in ACPI. Because i have seen there is no HPET device in iMac17,1 and newer platform.

And system work fine

Docs/AcpiSamples/SSDT-IRQFix.dsl
vit9696: But does it follow the ACPI specification? Especially if `_STA` is a method.

Docs/AcpiSamples/SSDT-IRQFix.dsl
vit9696: We tried it here, and it did not work for us, and we believe it violates the spec.

vit9696:
These changes look very interesting but we strongly lack hardware, tests, and understanding of whether we actually can apply them. We also believe that the changes will be different on many firmwares. For this reason I suggest this to be reviewed as a part of @dortania guides and not be bundled with OpenCore itself. In this case there at least can be written the caveats and nuances of these changes. CC @khronokernel

andreszerocross:
> These changes look very interesting but we strongly lack hardware, tests, and understanding of whether we actually can apply them. We also believe that the changes will be different on many firmwares. For this reason I suggest this to be reviewed as a part of @dortania guides and not be bundled with OpenCore itself. In this case there at least can be written the caveats and nuances of these changes. CC @khronokernel

Yes, it sound good. 
Currently in Dortania Guide ask user of OpenCore to edit their DSDT through SSDTtime to Patch IRQ and HPET,, we should avoid patch DSDT with OpenCore.

1Revenger1:
SSDTTime creates a set of ACPI patches and an SSDT - unless your saying that we shouldn't be recommending patches which go under ACPI->Patches in the SSDT?

andreszerocross:
Yes, i mean we doesn't need to rename Acpi method / device for HPET and other with SSDT above.

> Edit: As far as the SSDT presented here goes - I don't think that assigning 0 to _STA even if it's originally a method is part of the spec. I haven't seen it cause issues though.

I don't understand, what do you mean??

TonyC5:
I set HPET._STA=0 on my Kabylake and Coffeelake platforms and ignore IRQs (as mentioned earlier, Apple started setting HPET._STA=0 with Skylake).  I agree that setting HPET._STA=0 violates ACPI spec if HPET._STA is already defined as a method.  Many newer ACPI implementations offer a mechanism to enable / disable HPET (e.g. HPTE, where HPTE can be set to 0 or 1 to disable or enable HPET respectively).  If HPET._STA is already defined as a method, an alternative might be an ACPI rename patch that renames HPET._STA to HPET.XSTA (similar to the HPET._CRS rename patch) and then to define a new HPET._STA.

I am finding that setting HPET._STA=0 on my Kabylake and Coffeelake platforms (by setting HPTE = 0) and ignoring IRQs (since I am no longer concerned about IRQ conflicts with HPET) works perfectly.  This would probably need to be tested on each platform to confirm proper operation.

PR #150, opened 2020-11-04
Update SampleCustom.plist

1alessandro1:
https://github.com/acidanthera/bugtracker/issues/1255#issuecomment-721471164


vit9696:
That not what we want to do. The changes should be in the code.

PR #157, opened 2020-11-23
OcStorageLib: Uninstall EfiDevicePathProtocol on OcStorageFree

MikeBeaton:
`EfiDevicePathProtocol` is installed by `OcStorageInitFromFs` but never uninstalled. As it stands, this prevents any application which calls `OcStorageInitFromFs` (even when correctly followed later by calling `OcStorageFree` to clean up) from being able to exit cleanly, as per the comment below.

This PR uninstalls the protocol in what seems to be the obvious correct place.

### DIFF
--- a/Library/OcStorageLib/OcStorageLib.c
+++ b/Library/OcStorageLib/OcStorageLib.c
@@ -317,6 +317,15 @@ OcStorageFree (
   IN OUT OC_STORAGE_CONTEXT            *Context
   )
 {
+  if (Context->DummyStorageHandle != NULL) {
+    gBS->UninstallProtocolInterface (
+      Context->DummyStorageHandle,
+      &gEfiDevicePathProtocolGuid,
+      &mDummyBootDevicePath
+      );
+    Context->DummyStorageHandle = NULL;
+  }
+
   if (Context->Storage != NULL) {
     Context->Storage->Close (Context->Storage);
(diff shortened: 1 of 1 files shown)


MikeBeaton:
By 'exit cleanly' I'm referring to a specific real problem, not just something ideological, as follows:

Without this change, any app which has called `OcStorageInitFromFs` then `OcStorageFree`, and then exits from `UefiMain`, will not successfully return control either to OpenCore or to the Apple  boot manager, whereas without making those calls or with this fix, control is returned successfully.

Library/OcStorageLib/OcStorageLib.c
```diff
@@ -317,6 +317,14 @@ OcStorageFree (
   IN OUT OC_STORAGE_CONTEXT            *Context
   )
 {
+  if (Context->DummyStorageHandle != NULL) {
+    gBS->UninstallProtocolInterface (
```
vit9696, line 321: Please null the storage handle in case anybody calls this twice.

MikeBeaton: Changed. Squashed with previous to keep the whole change as a single commit.

MikeBeaton:
Coverity CI check script is aborting with `ERROR: Failed to download Coverity build tool with code 22!`. I think this is nothing to do with the content of my PR.

PR #159, opened 2020-11-26
[OpenDuet] Fix undefined behavior due to uninited memory (issue 1326)

nms42:
Garbage from memory was poped to CPU flags on return from legacy BIOS int86 calls.


vit9696:
Looks good to me, was that tested?

vit9696:
references https://github.com/acidanthera/bugtracker/issues/1326

nms42:
On my computers was tested.

PR #160, opened 2020-11-27
Build: Switch CI to GitHub Actions

dhinakg:


dakanji:
This is resulting in emails on cloned repos about CI runs.

PR #162, opened 2020-11-29
Added KernelQuirkExternalDiskIcons Kernel Patch

startergo:
Apply icon type patches to IONVMeFamily kext to force internal disk icons.


Docs/Differences/Differences.tex
```diff
@@ -2454,6 +2454,17 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
   \emph{Note}: This option is a substitution for BT4LEContinuityFixup.kext,
   which does not function properly due to late patching progress.
 
+\item
```
vit9696, line 2457: This file is generated automatically when BuildDocs.tool is executed. Do not edit it. We can build the documentation ourselves, so please limit your changes to Configuration.tex.

Docs/Configuration.tex
```diff
@@ -2400,6 +2400,17 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
 
   \emph{Note}: This option should be avoided whenever possible. Modern firmware
   usually have compatible AHCI controllers.
+  
+  \item
+  \texttt{ExternalDiskIconsNVME}\\
+  \textbf{Type}: \texttt{plist\ boolean}\\
+  \textbf{Failsafe}: \texttt{false}\\
+  \textbf{Requirement}: 10.4\\
```
vit9696, line 2408: 10.4 is definitely wrong, NVMe support was added in 10.13 or so in macOS. Please check where your patch actually applied and correct the minimal version.

Docs/Configuration.tex
```diff
@@ -2400,6 +2400,17 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
 
   \emph{Note}: This option should be avoided whenever possible. Modern firmware
   usually have compatible AHCI controllers.
+  
+  \item
+  \texttt{ExternalDiskIconsNVME}\\
```
vit9696, line 2405: I suggest NVMe instead of NVME. In both the quirk name and all the occurrences of the name. This is the correct spelling.

Docs/Configuration.tex
```diff
@@ -2400,6 +2400,17 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
 
   \emph{Note}: This option should be avoided whenever possible. Modern firmware
   usually have compatible AHCI controllers.
+  
+  \item
```
vit9696, line 2404: The indentation is a bit different compared to the upper item, please fix, since other changes are needed anyway.

Docs/Configuration.tex
```diff
@@ -2400,6 +2400,17 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
...
+  \texttt{ExternalDiskIconsNVME}\\
+  \textbf{Type}: \texttt{plist\ boolean}\\
+  \textbf{Failsafe}: \texttt{false}\\
+  \textbf{Requirement}: 10.4\\
+  \textbf{Description}: Apply icon type patches to IONVMeFamily.kext to force
+  internal disk icons for all NVME disks.
+
+  \emph{Note}: This option should be avoided whenever possible. Modern firmware
```
vit9696, line 2412: I do not like the description. To get an internal icon displayed you could add an SSDT with a device name and inject a `built-in` device property. This is the correct way to do it. When for some reason (e.g. testing or Thunderbolt) it is difficult, a quirk can be used. This should be explained here.

Docs/Differences/PreviousConfiguration.tex
```diff
@@ -2365,6 +2365,17 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
 
   \emph{Note}: This option should be avoided whenever possible. Modern firmware
   usually have compatible AHCI controllers.
+  
+  \item
```
vit9696, line 2369: This file contains previous version of the configuration, i.e. 0.6.3. It obviously did not have that quirk. Do not edit it.

Library/OcAppleKernelLib/CommonPatches.c
```diff
@@ -884,6 +884,55 @@ PatchForceInternalDiskIcons (
...
+STATIC
+EFI_STATUS
+PatchForceInternalDiskIconsNVME (
+  IN OUT PATCHER_CONTEXT    *Patcher,
+  IN     UINT32             KernelVersion
+  )
+{
+  EFI_STATUS          Status;
+
+  if (Patcher == NULL) {
```
vit9696, line 922: You must add version checking here similar to PatchUsbXhciPortLimit3. This patch should not return failure on systems where NVMe support is not present. This will break our tests for all the kernel versions.

vit9696 (commented):
Marked some technical issues that need to be addressed before merging. Are you sure the quirk is needed? For normal SATA disks there is no other option to make them internal, however, for NVMe disks it is easy to fix that via ACPI & DeviceProperties.

Docs/Differences/Differences.tex
startergo: Done

Docs/Configuration.tex
startergo: Done

Andrey1970AppleLife:
Don't edit Differences.tex

Docs/Configuration.tex
startergo: Done

Docs/Configuration.tex
startergo: Done

Docs/Differences/PreviousConfiguration.tex
startergo: Done

Library/OcAppleKernelLib/CommonPatches.c
startergo: Done. Please check if it is ok.

Docs/Configuration.tex
startergo: done

startergo:
> Don't edit Differences.tex

I removed my edits from there.

startergo:
> Marked some technical issues that need to be addressed before merging. Are you sure the quirk is needed? For normal SATA disks there is no other option to make them internal, however, for NVMe disks it is easy to fix that via ACPI & DeviceProperties.

Yes device properties's built-in property works, but it applies changes only a single PCIE. This is more universal. Added warning in the description as you suggested.

Docs/Differences/PreviousConfiguration.tex
```diff
@@ -2365,6 +2365,7 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
 
   \emph{Note}: This option should be avoided whenever possible. Modern firmware
   usually have compatible AHCI controllers.
+
```
vit9696, line 2368: The change is still here.

Docs/Differences/Differences.tex
```diff
@@ -2464,7 +2477,7 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
 
   \emph{Note}: This option should be avoided whenever possible. Modern firmware
   usually have compatible AHCI controllers.
-
+
```
vit9696, line 2480: The change is still here.

Docs/Differences/Differences.tex
```diff
@@ -2454,6 +2454,19 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
   \emph{Note}: This option is a substitution for BT4LEContinuityFixup.kext,
   which does not function properly due to late patching progress.
 
+\item
```
vit9696, line 2457: The change is still here.

Library/OcAppleKernelLib/CommonPatches.c
```diff
@@ -884,6 +884,69 @@ PatchForceInternalDiskIcons (
...
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_INFO, "OCAK: Failed to apply patch com.apple.iokit.IONVMeFamily - %r\n", Status));
+  } else {
+    DEBUG ((DEBUG_INFO, "OCAK: Patch success com.apple.iokit.IONVMeFamily\n"));
+  }
+
+ //
+  // TODO: Check when the patch changed actually.
+  //
+  if (EFI_ERROR (Status)
```
vit9696, line 941: Why is that needed?

Utilities/TestKextInject/KextInject.c
```diff
@@ -300,6 +300,14 @@ ApplyKextPatches (
     DEBUG ((DEBUG_WARN, "[OK] Success KernelQuirkExternalDiskIcons\n"));
   }
 
+Status = PrelinkedContextApplyQuirk (Context, KernelQuirkExternalDiskIconsNVMe, KernelVersion);
```
vit9696, line 303: Indentation is broken here.

vit9696:
> > Marked some technical issues that need to be addressed before merging. Are you sure the quirk is needed? For normal SATA disks there is no other option to make them internal, however, for NVMe disks it is easy to fix that via ACPI & DeviceProperties.
> 
> Yes device properties's built-in property works, but it applies changes only to a single PCIE. This is more universal. Added warning in the description as you suggested.

Is it really all about lazy users? Because if so, I would rather not merge this.

Docs/Differences/Differences.tex
```diff
@@ -2454,6 +2454,19 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
...
+  \texttt{ExternalDiskIconsNVMe}\\
```
vit9696, line 2458: The change is still here.

Docs/Differences/PreviousConfiguration.tex
```diff
@@ -2364,7 +2364,7 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
   internal disk icons for all AHCI disks.
 
   \emph{Note}: This option should be avoided whenever possible. Modern firmware
-  usually have compatible AHCI controllers.
+  usually have compatible AHCI controllers.
```
vit9696, line 2367: The change is still here.

startergo (commented):
Please elaborate. The AHCI patch does the same thing and also can be substituted by the built-in PCIE property in device properties and as a Kernel patch. What is the difference?
image

vit9696:
> Please elaborate. The AHCI patch does the same thing and also can be substituted by the built-in PCIE property in device properties and as a Kernel patch. What is the difference? Plus the built-in property implies permanent setup. If the devices change their slots it has to be reset.

I am not positive you can always do that with AHCI as they disks are not PCIe devices. Unsure, cc @Andrey1970AppleLife

startergo:
> > Please elaborate. The AHCI patch does the same thing and also can be substituted by the built-in PCIE property in device properties and as a Kernel patch. What is the difference? Plus the built-in property implies permanent setup. If the devices change their slots it has to be reset.
> 
> I am not positive you can always do that with AHCI as they disks are not PCIe devices. Unsure, cc @Andrey1970AppleLife

I tested the built-in PCIE property in device properties and the  Kernel patch for the AHCI  on my cMP and they work.

vit9696:
So, shall we remove the other one instead of adding an NVMe one instead?

startergo:
> So, shall we remove the other one instead of adding an NVMe one instead?

I have MacBookPro15,1 and there are only NVME controllers even on my external disk enclosure (thunderbolt) so I can't test the AHCI property there, someone on a PC may be able to check. Can you point me to a sample for the SSDT patch?

PR #163, opened 2020-12-01
Clean up keys after done with signing vault

junzhli:
Hi,

I notice the attached sign.command that it leaves Keys folder in EFI/OC/
To my understanding, those keys no longer need during the boot process. Its public key was compiled into opencore.efi
I think it would be safer to remove those keys after done with this script
Also, I put the folder Keys under temp folder "/tmp" to prevent create_vault from processing them.

### DIFF
--- a/Utilities/CreateVault/sign.command
+++ b/Utilities/CreateVault/sign.command
@@ -5,7 +5,12 @@ abort() {
   exit 1
 }
 
-if [ ! -x /usr/bin/dirname ] || [ ! -x /bin/chmod ] || [ ! -x /bin/mkdir ] || [ ! -x /usr/bin/openssl ] || [ ! -x /bin/rm ] || [ ! -x /usr/bin/strings ] || [ ! -x /usr/bin/grep ] || [ ! -x /usr/bin/cut ] || [ ! -x /bin/dd ] ; then
+cleanup() {
+  echo "Cleaning up keys"
+  rm -rf "${KeyPath}"
+}
+
+if [ ! -x /usr/bin/dirname ] || [ ! -x /bin/chmod ] || [ ! -x /bin/mkdir ] || [ ! -x /usr/bin/openssl ] || [ ! -x /bin/rm ] || [ ! -x /usr/bin/strings ] || [ ! -x /usr/bin/grep ] || [ ! -x /usr/bin/cut ] || [ ! -x /bin/dd ] || [ ! -x /usr/bin/uuidgen ] ; then
   abort "Unix environment is broken!"
 fi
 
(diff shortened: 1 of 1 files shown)


Utilities/CreateVault/sign.command
```diff
@@ -77,5 +77,10 @@ fi
 
 /bin/dd of="${OCBin}" if="${PubKey}" bs=1 seek="${off}" count=528 conv=notrunc || abort "Failed to bin-patch ${OCBin}"
 
+echo "Cleaning up keys"
+if [ -d "${KeyPath}" ]; then
+  rm -rf "${KeyPath}" | abort "Failed to clean up keys ${KeyPath}!"
```
vit9696, line 82: It is not `|` but `||`.

Utilities/CreateVault/sign.command
```diff
@@ -77,5 +77,10 @@ fi
 
 /bin/dd of="${OCBin}" if="${PubKey}" bs=1 seek="${off}" count=528 conv=notrunc || abort "Failed to bin-patch ${OCBin}"
 
+echo "Cleaning up keys"
+if [ -d "${KeyPath}" ]; then
```
vit9696, line 81: `-d` is not necessary here, as you always delete.

vit9696:
The implementation is not correct for two reasons:

1. `/tmp/Keys` directory may exist due to other software. Please make it `/tmp/Keys-$(/usr/bin/uuidgen)` to make sure it is unique. Furthermore directory existence prior to creation should now cause a fatal error, as `$(/usr/bin/uuidgen)` is pretty much guaranteed to produce a unique value. Also make sure that `/usr/bin/uuidgen` exists prior to executing it in the top of the script like other tools are.
2. Remove the directory not only in the normal workflow but also in abort to ensure that all the keys are dead regardless of script success or failure.

I also left a couple of notes inline.

junzhli:
Hi @vit9696 ,
Thanks for the review. I push another revised version!

Utilities/CreateVault/sign.command
```diff
@@ -77,5 +84,7 @@ fi
 
 /bin/dd of="${OCBin}" if="${PubKey}" bs=1 seek="${off}" count=528 conv=notrunc || abort "Failed to bin-patch ${OCBin}"
 
+
```
vit9696, line 87: Why is that necessary?

Utilities/CreateVault/sign.command
```diff
@@ -45,6 +50,8 @@ if [ ! -x ./RsaTool ] || [ ! -x ./create_vault.sh ]; then
   fi
 fi
 
+trap cleanup EXIT ERR INT TERM
```
vit9696, line 53: Do you believe it is better to directly calling `cleanup` in e.g. `abort`? I personally do not mind.

vit9696:
You may want to add an exception for that particular line with trap as currently CI fails.

junzhli:
@vit9696,
I remove the unnecessary blank line and unsupported signals for adapting to POSIX-compilant
My concern about putting cleanup() in abort() function is that all of code above line 53 invoking abort() won't work properly as var KeyPath is not ready at that moment. Also it provides guarantee that cleanup gets called when users terminate the script by pressing Ctrl+C. Thanks!

PR #164, opened 2020-12-01
Support reordering Bootstrap by the FW

Download-Fritz:
The firmware may move our boot option away from bootstrap-index. Scan BootOrder for the option by DP suffix. Implicitly fixes issues with boot entry duplication.

### DIFF
--- a/Include/Acidanthera/Library/OcDevicePathLib.h
+++ b/Include/Acidanthera/Library/OcDevicePathLib.h
@@ -154,6 +154,21 @@ OcFileDevicePathNameLen (
   );
 
 /**
+  Retrieve the length of the full file path described by DevicePath.
+
+  @param[in] DevicePath  The Device Path to inspect.
+
+  @returns   The length of the full file path.
+  @retval 0  DevicePath does not start with a File Path node or contains
+             non-terminating nodes that are not File Path nodes.
+
+**/
+UINTN
+OcFileDevicePathFullNameLen (
+  IN CONST EFI_DEVICE_PATH_PROTOCOL  *DevicePath
+  );
+
+/**
   Retrieve the size of the full file path described by DevicePath.
 
   @param[in] DevicePath  The Device Path to inspect.
@@ -377,11 +392,20 @@ OcGetNextLoadOptionDevicePath (
   IN EFI_DEVICE_PATH_PROTOCOL  *FullPath
   );
 
+/*
+  Checks DevicePath for whether it ends with file path Suffix.
+
+  @param[in] DevicePath    The Device Path to check.
+  @param[in] Suffix        The suffix to check for.
+  @param[in] SuffixLen  Must be equal to StrLen(Suffix).
+
+  @returns  Whether DevicePath ends with Suffix.
+*/
 BOOLEAN
 OcDevicePathHasFilePathSuffix (
   IN EFI_DEVICE_PATH_PROTOCOL  *DevicePath,
   IN CHAR16                    *Suffix,
-  IN UINTN                     SuffixSize
+  IN UINTN                     SuffixLen
   );
 
 #endif // OC_DEVICE_PATH_LIB_H
--- a/Include/Acidanthera/Library/OcDevicePathLib.h
+++ b/Include/Acidanthera/Library/OcDevicePathLib.h
@@ -377,4 +377,11 @@ OcGetNextLoadOptionDevicePath (
   IN EFI_DEVICE_PATH_PROTOCOL  *FullPath
   );
 
+BOOLEAN
+OcDevicePathHasFilePathSuffix (
+  IN EFI_DEVICE_PATH_PROTOCOL  *DevicePath,
+  IN CHAR16                    *Suffix,
+  IN UINTN                     SuffixSize
+  );
+
 #endif // OC_DEVICE_PATH_LIB_H
--- a/Library/OcBootManagementLib/DefaultEntryChoice.c
+++ b/Library/OcBootManagementLib/DefaultEntryChoice.c
@@ -12,6 +12,8 @@
   WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 **/
 
+#include <Uefi.h>
+#include <OpenCore.h>
 #include "BootManagementInternal.h"
 
 #include <Guid/AppleFile.h>
@@ -19,6 +21,7 @@
 #include <Guid/GlobalVariable.h>
 #include <Guid/OcVariable.h>
 
+#include <Protocol/DevicePath.h>
 #include <Protocol/LoadedImage.h>
 #include <Protocol/OcFirmwareRuntime.h>
 #include <Protocol/SimpleFileSystem.h>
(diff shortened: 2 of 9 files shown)


Include/Acidanthera/Library/OcDevicePathLib.h
```diff
@@ -377,4 +377,11 @@ OcGetNextLoadOptionDevicePath (
   IN EFI_DEVICE_PATH_PROTOCOL  *FullPath
   );
 
+BOOLEAN
+OcDevicePathHasFilePathSuffix (
+  IN EFI_DEVICE_PATH_PROTOCOL  *DevicePath,
+  IN CHAR16                    *Suffix,
```
vit9696, line 383: - Needs docs?
- Why not CONST for the Suffix?

Library/OcBootManagementLib/DefaultEntryChoice.c
```diff
@@ -12,13 +12,16 @@
   WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 **/
 
+#include <Uefi.h>
+#include <OpenCore.h>
```
vit9696, line 16: Ugh, it is not good that libraries include that header. Do we ever do it anywhere else? I designed it to be local to Platform/OpenCore originally. Most likely the things we need from there are to be moved elsewhere if at all.

Library/OcBootManagementLib/DefaultEntryChoice.c
```diff
@@ -197,20 +169,24 @@ InternalDebugBootEnvironment (
...
       if (UefiDevicePath == NULL) {
         DEBUG ((
           DEBUG_INFO,
           "OCB: %u -> Boot%04x - failed to read\n",
           (UINT32) Index,
           BootOrder[Index]
           ));
+        FreePool (LoadOption);
         continue;
       }
```
vit9696, line 191: LoadOption needs to be freed somewhere, no?

Library/OcBootManagementLib/DefaultEntryChoice.c
```diff
@@ -850,6 +844,65 @@ OcSetDefaultBootEntry (
...
+
+  if (BootIndex == NumBootOptions) {
+    return NULL;
+  }
+
+  if (LoadPath != NULL) {
+    *LoadPath   = CurrDevicePath;
+  }
+
+  *BootOption = BootIndex;
```
vit9696, line 902: Rename to BootOptionIndex. Too confusing otherwise.

Library/OcBootManagementLib/DefaultEntryChoice.c
```diff
@@ -850,6 +844,65 @@ OcSetDefaultBootEntry (
...
+  return CurrLoadOption;
+}
+
 STATIC
 EFI_STATUS
 InternalRegisterBootOption (
```
vit9696, line 908: BootstrapBootOption? Or pass some prefix as a value. The function is not really generic.

Library/OcBootManagementLib/DefaultEntryChoice.c
```diff
@@ -970,26 +999,51 @@ InternalRegisterBootOption (
       FreePool (BootOrder);
       return EFI_OUT_OF_RESOURCES;
     }
+
+    Option = InternalGetBoostrapOptionData (
+      &OptionSize,
+      &BootOptionIndex,
+      &CurrDevicePath,
+      &BootOrder[1],
```
vit9696, line 1007: Why [1] and not [0]? Did I miss anything

Library/OcBootManagementLib/VariableManagement.c
```diff
@@ -244,6 +246,69 @@ DeleteVariables (
...
+{
+  EFI_STATUS               Status;
+  UINTN                    BootOrderSize;
+  UINT16                   *BootOrder;
+  UINT16                   OptionIndex;
+  VOID                     *OptionData;
+
+  BootOrderSize = 0;
+  Status = gRT->GetVariable (
+    EFI_BOOT_ORDER_VARIABLE_NAME,
```
vit9696, line 262: Do not we really have functions for boot order reading?

Library/OcBootManagementLib/VariableManagement.c
```diff
@@ -293,40 +357,16 @@ OcDeleteVariables (
...
-          "OCB: Found %g:%s for preservation of %u bytes\n",
-          &gEfiGlobalVariableGuid,
-          BootOptionVariable,
-          (UINT32) BootOptionSize
-          ));
-      }
-    } 
-    
-    if (EFI_ERROR (Status)) {
+        BootOptionVariable,
```
vit9696, line 366: Hrm, why do we still have OC_BOOTSTRAP_INDEX_VARIABLE_NAME?

Library/OcDevicePathLib/OcDevicePathLib.c
```diff
@@ -1231,3 +1231,49 @@ OcGetNumDevicePathInstances (
 
   return NumInstances;
 }
+
+BOOLEAN
+OcDevicePathHasFilePathSuffix (
+  IN EFI_DEVICE_PATH_PROTOCOL  *DevicePath,
+  IN CHAR16                    *Suffix,
+  IN UINTN                     SuffixSize
```
vit9696, line 1239: A bit unpleasant that can access one char paste the Suffix at least. Should document…

Include/Acidanthera/Library/OcDevicePathLib.h
Download-Fritz: Because the subfunctions like OcStriCmp do not take CONST, could drop another commit at some point.

Library/OcBootManagementLib/DefaultEntryChoice.c
Download-Fritz: Sorry, I didn't realise this, it was for the Bootstrap path macro. How to solve it, move the macro or take a parameter from somewhere?

Library/OcBootManagementLib/DefaultEntryChoice.c
Download-Fritz: Ouch, overlooked

Library/OcBootManagementLib/DefaultEntryChoice.c
Download-Fritz: OK

Library/OcBootManagementLib/DefaultEntryChoice.c
Download-Fritz: OK

Library/OcBootManagementLib/DefaultEntryChoice.c
Download-Fritz: We read to "(VOID *) (BootOrder + 1)" to "pre-reserve" the first index to move Bootstrap there.

Library/OcBootManagementLib/VariableManagement.c
Download-Fritz: No, we do it in two places it seems, and above case pre-reserved the first index to avoid re-alloc (in case Bootstrap is not already in there) and the second does not. I don't know whether they could be nicely unified without sacrificing the no-realloc.

Library/OcBootManagementLib/VariableManagement.c
Download-Fritz: Missed, thanks

PR #168, opened 2020-12-18
update readme

wy414012:
Add simplified Chinese introduction document


vit9696:
Thank you for your effort, but we do not need translations to any other languages at the moment.

PR #173, opened 2020-12-27
OC: auto-generate config files from template .plist file

MikeBeaton:
@vit9696 

I have got somewhere with https://github.com/acidanthera/bugtracker/issues/1337 .

I took a little longer than I hoped, but I think the code is okay. I do understand that you said you would not necessarily merge it, but would be interested to see the results. Nevertheless, I've tried to polish it up to show how it can work and to make it (close to?) potentially mergeable.

`PlistToConfig.py` - which does the work - is considerably longer than the original 30-odd line file which convinced me that the approach could work; but after a couple of rounds of refactoring I think the complexity of the code does match the complexity of the problem it is solving.

The basic idea is to automatically derive the path of config items from the `<key>` path to the item in a template plist file. This basically works as is - because that is basically how the config files were already written - but then a fair amount of complexity is added to exactly match the current layout, including the fact that the parent or children at a given node often use a slight variant on the exact `<key>` path name, in either the .c or the .h file or both. (We also derive the .c/.h data format of `<data>` and `<integer>` items from the plist sample data, again adding attributes to override this only when needed.)

I realised about half-way through that this approach could also auto-generate `SampleCustom.plist` as well as `OcConfigurationLib.c` and `OcConfigurationLib.h` and it seemed silly not to finish off and do so. Importantly, I believe that doing this has picked on up a couple of remaining sorting errors in the hand-coded `SampleCustom.plist` file (and was picking up more similar errors that have just recently been fixed; see below). Note that `Template.plist` from which everything is generated in this approach *is* a complete copy of `SampleCustom.plist`, with minor additional annotation, hence the possibility (and helpfully sense!) of doing this.

There are three commits in the PR.

I normalize `OcConfigurationLib.c` and `OcConfigurationLib.h` in commit #1 4486556dddd55a4a3c4c687c770e4c170173d357. The changes there are explained in the commit message; they are minor normalisation to spacing, use of commas, etc., to make it possible to accurately compare the hand-coded and generated files, without noise from unimportant differences; but also note that even though, e.g., spacing is not *exactly* matched, I have made quite a lot of effort to make the generated files look quite like the hand-coded ones, as can be seen in the final generated files in commit #3 6c9777aa8b88f60f93308a9f3bef066f595b6aa4.

Commit #2 adds the template plist file and the autogeneration script and associated files; commit #3 replaces the normalized original files with the generated files.

After the normalization step (and then, obviously, after all the work I've put in to try to make this work properly!) the generated files show only the following differences from the originals (I think these are all actually _fixes_ to minor errors in the original, hand-coded files):

 - Two nested keys in `SampleCustom.plist` (`run-efi-updater` and `Type`) get moved into correct alphabetical order (see 6c9777aa8b88f60f93308a9f3bef066f595b6aa4)
   - While working on this I was also picking up several other out-of-order keys in `SampleCustom.plist`; but these were all recently fixed in 1b46dfca22ee42c7fb2d6fd2bdf5120e69729ece (hence are correctly no longer showing up as differences from the auto-generated file)
 - After stripping spaces (see below for how and why) the generated `OcConfigurationLib.c` has *no* changes from the normalized original file
 - In `OcConfigurationLib.h` we see the following changes only:

```
364c364
<  _(BOOLEAN , ExposeSensitiveData , , OCS_EXPOSE_VERSION , ()) \
---
>  _(UINT32 , ExposeSensitiveData , , OCS_EXPOSE_VERSION , ()) \
512c512
<  _(OC_STRING , ChassisAssetTag , , OC_STRING_CONSTR ("", _, __) , ()) \
---
>  _(OC_STRING , ChassisAssetTag , , OC_STRING_CONSTR ("", _, __) , OC_DESTR (OC_STRING) ) \
514c514
<  _(OC_STRING , ChassisSerialNumber , , OC_STRING_CONSTR ("", _, __) , ()) \
---
>  _(OC_STRING , ChassisSerialNumber , , OC_STRING_CONSTR ("", _, __) , OC_DESTR (OC_STRING) ) \
516c516
<  _(OC_STRING , ChassisVersion , , OC_STRING_CONSTR ("", _, __) , ()) \
---
>  _(OC_STRING , ChassisVersion , , OC_STRING_CONSTR ("", _, __) , OC_DESTR (OC_STRING) ) \
571c571
<  _(UINTN , SetupDelay , , 0 , ()) \
---
>  _(UINT32 , SetupDelay , , 0 , ()) \
```

I *think* all five of those changes are fixes to minor bugs in the original - i.e. I think they all 'should' be as in the generated version, not as in the current, hand-coded version?

To regenerate the three live files from the template file, run `make live` in the `Library/OcConfigurationLib`. (This was done to generate commit #3 6c9777aa8b88f60f93308a9f3bef066f595b6aa4).

To compare the original and generated `SampleCustom.plist` just look at any differences after generation of the live files (i.e. as seen in the third commit 6c9777aa8b88f60f93308a9f3bef066f595b6aa4); no other normalization or space-stripping is required for this file, the original and generated files are almost identical.

To compare the original and generated .c and .h files without being overwhelmed by minor noise, go to the second commit (i.e. b85081708a4a81a88f38a55251bc85f819ecd689 before the normalized files are over-written with the generated files in the final commit), then run `make debug`, then diff `_OcConfigurationLib.c` with `_test.c` and diff `_OcConfigurationLib.h` with `_test.h` (this works around the fact that spacing is somewhat different between the generated and original files - even after normalization - by stripping excess spacing from both of them; the `debug` target in the makefile and `StripSpace.py` are used only for this step).

I could add some more info about which additional markup tags I added in `Template.plist` and why, but this post is already pretty long, and hopefully they are somewhat self-explanatory?

The reason I was interested in working on this (as already mentioned in the related issue) is shown here https://github.com/MikeBeaton/macos-boot-helper/tree/master/Application/BootHelper where the config files in that directory are now generated using this new tool, and then loaded and parsed using the relevant parts of OpenCore. (Note that as of now the config there is far from final: it is basically a very stripped down OC config with a few minor changes just to test that this works - I wanted to post up something here before going back and finishing that.)


vit9696:
Hi Mike, thanks for working on this.

I made a closer look on the implementation, and while it looks nice, I believe this is not quite what we might have needed for a switch. Evaluating the background, I am afraid we are not quite ready for the switch in the nearest future at all. Even so, I believe this is a very nice thing to have, and it will be great to have it hand in your own repository as an inspiration at least.

For everyone involved let me try to list some points I have in mind:

- The macro hell we introduced in OcConfigurationLib.c/.h is mainly due to the attempts to avoid an external code generation tool. It does not feel like a great idea to keep it if an external code generation tool exists and can be used. It will not have the C preprocessor limitations to deal with.
- In most cases the fields in OcConfigurationLib.h and OcConfigurationLib.c are sorted differently. This is due to the alphabetic sorting requirement for the schema, and attempt to reduce the in-memory configuration size by reducing padding in the structure definitions. I believe for a generator it would be best to take this into account.
- The reason to use plist format is due to wide editor availability in macOS. Extending plist within the xml format might be helpful for the generation, but not much helpful for the user. Plist format is not really plain-text readable or edible due to base64 data encoding at least. Thus I do not think it is a good idea to use xml for the schema.
- Sample and failsafe values are quite different things. Sample contains a configuration close to the one you can find on non-Apple hardware, while failsafe values try to avoid any changes to the system. Storing them in one place is not just uneasy thing due to non-empty array and dictionary contents in particular, but is also a confusing matter.

Thanks!
P.S. Perhaps others want to add something as well.

vit9696:
By the way, considering your project, I wonder if you would benefit from an ability to allow to pass OpenCore configuration through a protocol. I.e. instead of reading config.plist OpenCore could use an in-memory file supplied by the first-stage bootloader.

MikeBeaton:
Thanks for the feedback.

#1. Fair point!

#2. I thought perhaps there was something I wasn't understanding about the reason for the sorting in the .h file. Obviously it would not be that hard to change the generator to deal with that too, but I understand there are the other points as well.

#3. I still think I perhaps disagree about using a plist as the template, at least as follows: 1/2) I'm __not__ trying to change what __end users__ use to set up OC __at all__; this is just a starting specification for the available config, only for use within the project (though to the extent that people are using plist editors wiin the project, then fair point). Also, 2/2) I do not think DTD can work (or not at all easily), because what makes a valid OC plist depends on key __values__ - so, at the least, I think you'd need a DTD for a initial non-plist XML format with the right logical structure, followed by a non-standard transformation step from that to a plist file.

#4. Also to some extent fair enough; doing it this way I agree is something of a hack - it just followed from the logic of noting that the sample plist file does (almost!) specify what the config .c and .h file must be, already!

If I can ask a question, I was still wondering if the (alleged; minor) bugs that I thought this approach identified in `OcConfigurationLib.h` are really bugs? i.e. `ExposeSensitiveData` and `SetupDelay` should really be `UINT32`, and the three strings are missing their destructors? (And the same but less important about the remaining two out of order changes that I thought I identified in `SampleCustom.plist`?)

MikeBeaton:
I'm not quite sure that I fully followed your last comment, but if I can come back with something that I think is relevant: yes, I had to copy quite a lot (only tens of lines) of setup biolerplate code from OC into my own project, before I could then use the OC file access and plist deserialization libraries which are what I really wanted to get at. So for me the issue (as I'm currently understanding it) would be how to separate the OC-specific and non-OC-specific parts of the OC setup, and make the non-OC-specific parts available to users of OC as a library, without having to copy and paste anything. (Perhaps through an entry point with callbacks, or something?) But I'm not sure that that really addresses what you were saying, nor even whether that's a use-case you're particularly concerned to support! (Either way being up to you, of course!)

vit9696:
> If I can ask a question, I was still wondering if the (alleged; minor) bugs that I thought this approach identified in OcConfigurationLib.h are really bugs? i.e. ExposeSensitiveData and SetupDelay should really be UINT32, and the three strings are missing their destructors? (And the same but less important about the remaining two out of order changes that I thought I identified in SampleCustom.plist?)

Sorry, forgot to commit these actually. Thank you. The ones in Sample.plist look wrong to me though, T follows S, so Type must follow Size. Similarly with p and r.

> I'm not quite sure that I fully followed your additional comment, but if I can come back with something that I think is relevant: yes, I had to copy quite a lot (only tens of lines) of setup biolerplate code from OC into my own project, before I could then use the OC file access and plist deserialization libraries which are what I really wanted to get at. So for me the issue (as I'm currently understanding it) would be how to separate the OC-specific and non-OC-specific parts of the OC setup, and make the non-OC-specific parts available to users of OC as a library, without having to copy and paste anything. (Perhaps through an entry point with callbacks, or something?) But I'm not sure that that really addresses what you were saying, nor even whether that's a use-case you're particularly concerned to support! (Either way being up to you, of course!)

I thought that you might want to edit the configuration in your UI, so I wondered whether you might need a functionality to feed custom configuration to OpenCore.

As for integrating custom OC serialisation libs, I don't know, should be fairly easy. See e.g. OcStorageLib for non-OcConfigurationLib example. But I guess your tools make it even easier for you ^^

MikeBeaton:
> Sorry, forgot to commit these actually. Thank you.

I glad those worked out - sorry, I know it's extra work for you, but it would be nice to have the chance to put together a little commit of changes like that (I think there was a similar, and similarly tiny, change that I pointed out before, elsewhere in OC - it would be nice to have the chance to make commits in OC for changes like those, since that's kind of the 'currency' of open source). Not that I'm asking for any change at this point though. I hope it's okay to say! (And thanks for the credit that you did add!)

> The ones in Sample.plist look wrong to me though, T follows S, so Type must follow Size. Similarly with p and r.

Oh dear, that's embarrassing! I'm not at my Mac now to figure out why those came in the wrong order (as you can tell, I didn't engage my brain long enough to realise that these were wrong, even though auto-generated - doh, sorry). It's true that the ordering of that .plist file *is* generating the correct .c file, so there might still be something going on, but there might not - I'll check!

MikeBeaton:
> the ordering of that .plist file is generating the correct .c file, so there might still be something going on, but there might not - I'll check

I was reading my own diffs backwards - assuming that since auto-generated the final result must be right; and I think I was probably asking you to re-import an minor mis-ordering error from 0.6.3 or so (in a bit of the .plist file that my generator wasn't using; because it only uses the first sample struct in an array or map) - apologies. (So I'm glad I at least spotted some real small fixes elsewhere! 🙃 )

> I thought that you might want to edit the configuration in your UI, so I wondered whether you might need a functionality to feed custom configuration to OpenCore.

Thanks for the suggestion - my current boot helper system can work stand-alone (i.e. without OC at all, e.g. to set/clear `csr-active-config` on a standard Mac), or it can be configured as a tool within OC, so I think it already works for what I need it for without that. But thank you.

MikeBeaton:
Hi Vitaly,

I'll stop pinging you with messages now! But just coming back with some final thoughts, unless/until/in case you want to pick up on it again at any point:

 - My generator has the correct template structure in memory, except for the correct .h ordering to reduce padding
 - I'd be happy to implement some padding logic, to auto-generate the correct .h order too
 - I could then output from my current generator into whatever format you'd prefer, for a new-style template (which I guess would only contain failsafe values, not sample values); i.e. I'm suggesting this could perhaps be used generate an initial template, in some specified format, more quickly/efficiently than doing it by hand
 - Then thinking about what the new template format would be and how it might be used: I guess if the preferred format was a DTD then it is possible that the .c and .h files could be generated from this, with transforms, using standard tools? That *would* be new tooling in the build chain, but then if a custom Python script is the wrong tool for the job, then it just is.
 - Also just trying to think through what the preferred .c and .h output would be, given your first comment about the macros: I guess it would be something more like the .c and .h code *after* the macros have been processed?
 - Finally, just noting that, rather than (re)generating a plist file to compare with the original, as I'm currently trying to do, the correct template format with the correct tooling could, I guess, be used to verify that the two sample .plist files in Docs are correct and in order; putting it another way, based on my experiments so far, I guess a useful eventual goal of automatic tooling here should be not only to generate the .c and .h files, but also to to verify any sample .plist files

But I am aware that there'll only be so much time you'll want to spend discussing this now(!), and that you've basically said you're probably not ready for this switch yet, period.

So as I say, just throwing in some final thoughts - and still keen to contribute if the route I've gone down looks at some point like it could still be adapted to be useful.

Mike

PR #174, opened 2020-12-31
[OpenDuet] Remove custom UsbLegacySupport

nms42:
Employ functionality from edk2.


roddy20:
USB keyboard does not work anymore

nms42:
> USB keyboard does not work anymore

Details, logs, ... please

roddy20:
it works OK when built to new empty folder, the problem was when updating an older version

PR #177, opened 2021-01-07
Replace HFS+ driver with compatible opensource one

shchuko:
Potentially closes https://github.com/acidanthera/bugtracker/issues/659 

Used partially-updated driver from here: https://github.com/gsomlo/edk2/tree/gls-miscopt/OvmfPkg/FswHfsPlus

Updated version is here: https://github.com/shchuko/OvmfDarwinPkg/tree/master/FswHfsPlus

Tasks todo:

- [x]  fix building

- [x]  check is AppleBless requests handling works correctly

- [x]  find a reason why volume label can't be shown, fix (was not implemented at all)

- [ ]  update documentation


roddy20:
open source driver is 2..3 times slower when reading BaseSeystem.dmg from USB device

shchuko:
> open source driver is 2..3 times slower when reading BaseSystem.dmg from USB device

Reasons to use open-source driver were listed in the [linked issue](https://github.com/acidanthera/bugtracker/issues/659). This driver replace helps to resolve only one but big problem - replace GPL-distributed VBoxHfs driver

vit9696:
CC @nms42

shchuko:
Seems to be working fine now. Tested to boot BaseSystem images:

- 10.13.6  macOS High Sierra
- 10.14.6  macOS Mojave
- 10.15.7  macOS Catalina
- 11.1  macOS Big Sur
image
image

Staging/FswHfsPlus/fsw_hfsplus.c
```diff
@@ -471,6 +474,24 @@ fsw_hfsplus_vol_mount(struct fsw_hfsplus_volume *v)
...
+        label.len = fsw_u16_be_swap(thread->nodeName.length);
+        label.size = sizeof(fsw_u16) * label.len;
+        label.data = thread->nodeName.unicode;
+        label.type = FSW_STRING_TYPE_UTF16_BE;
+        fsw_strdup_coerce(&v->g.label, FSW_STRING_TYPE_UTF16, &label);
+    }
+
+    fsw_free(btnode);
+done:
     return FSW_SUCCESS;
```
vit9696, line 495: Hrm, is it intended to return success always? Makes little sense to me.

shchuko: Upward this `return FSW_SUCCESS` there're several  `if (status) return status;` for fatal cases when the volume wasn't mounted. 
But if the volume was mounted, all operations were successful (reading volume header, creating root, etc.) except label reading finished with an error, we're continuing with FSW_SUCCESS. 
Maybe it's a wrong concept and we should return error on volume label reading fails (when non-readed label is fatal failure)

vit9696: Well, if the label is not read, what will the caller get? Empty or garbage?

shchuko: The caller will get an empty string, as it was initialized from `fsw_mount()`, fsw_core.c

Just noticed that when `fsw_strdup_coerce()` fails the label type can be set not  to empty (utf16), my mistake. I'll fix it

shchuko: Changes applied, now returning empty string on label reading failure

vit9696:
* We tested the driver and can confirm that it indeed bots both macOS and macOS DMG recoveries without issues.
* The performance is about the same compared to VBoxHfs (around 5 mins for 10.12 recovery compared to around 2 minutes for the same recovery with the proprietary HfsPlus driver from OcBinaryData).
* Security evaluation of the driver is unclear, but this is the same level of trust we have for VBoxHfs, so I will not request security review to be performed right now.
* Similarly codestyle does not match EDK II/UEFI, but this is no different from VBoxHfs.

I suggest:

- to merge this driver as is
- update the docs (I guess I can handle that)
- drop VBoxHfs from the codebase
- explore performance, security, and codestyle in subsequent commits

Do we agree on that? CC @Goldfish64 @mhaeuser @nms42 @PMheart

nms42:
No objections.

/nms

roddy20:
is there any possibility to use open source PartitionDxe.efi or do we still need the version from Apple firmware to boot from Lion/Maverick images?

telepati:
Does the performance vary from system to system? Because with this new driver, I can reach Recovery with Command + R key in about 30 seconds. Even the boot time of my system is 56 seconds longer than this time.

PR #181, opened 2021-01-19
Clarify Function Names in OcApfsLib

dakanji:
OcApfsLib has two functions with very similar naming
- OcApfsConnectDevice
- ApfsConnectDevice

This introduces some potential confusion especially when using as a 3rd party external library. 

This PR tries to remove such with the following changes that also seems to slightly better reflect what they do:
- OcApfsConnectDevice   ->    OcApfsConnectHandle
- ApfsConnectDevice       ->    OcApfsConnectDevice

PS: There are some other functions in `Library/OcApfsLib/OcApfsConnect.c` that are not canonically named that I haven't changed and one of these, `ApfsCheckOpenCoreScanPolicy`, does not appear to be used anywhere.

### DIFF
--- a/Library/OcApfsLib/OcApfsConnect.c
+++ b/Library/OcApfsLib/OcApfsConnect.c
@@ -476,7 +476,7 @@ OcApfsConfigure (
 }
 
 EFI_STATUS
-OcApfsConnectDevice (
+OcApfsConnectHandle (
   IN EFI_HANDLE  Handle,
   IN BOOLEAN     VerifyPolicy
   )
--- a/Include/Acidanthera/Library/OcApfsLib.h
+++ b/Include/Acidanthera/Library/OcApfsLib.h
@@ -94,7 +94,7 @@ OcApfsConnectParentDevice (
   @retval EFI_SUCCESS if the device was connected.
 **/
 EFI_STATUS
-OcApfsConnectDevice (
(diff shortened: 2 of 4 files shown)


Library/OcApfsLib/OcApfsConnect.c
```diff
@@ -392,7 +392,7 @@ ApfsStartDriver (
 
 STATIC
 EFI_STATUS
-ApfsConnectDevice (
+OcApfsConnectDevice (
```
vit9696, line 395: This is a private unexported (STATIC) function. Usually we do not give private functions an Oc prefix. In my opinion, leave it just Apfs. Similarly we give Internal prefix to private exported (non-STATIC) functions.

dakanji: Got it on the static function naming although not consistently applied throughout the project.

dakanji: Any steer on the unused `ApfsCheckOpenCoreScanPolicy` function?

vit9696: It is used, at least in master.

dakanji: Found the usage. Cheers.
Wil push update later.

PR #182, opened 2021-01-27
Updated docs about minimum kernel requirements

1alessandro1:
For `AtherosE2200Ethernet.kext` you can find it in the description:

*Requires macOS Sierra or newer.*

For `RealtekRTL8111.kext` instead:

*The driver has been successfully tested under 10.8.2 - 10.15.6*


vit9696:
Prebuilt kexts published there are very outdated (2018 vs 2020), and Mieze kexts do not use MacKernelSDK. The actual minimal supported version, I am afraid, cannot be guaranteed upstream. A standalone repository with a properly configured build system will be needed to provide this. Similarly to how we did with IntelMausi.

PR #184, opened 2021-01-29
SmbiosPatch: Swap SystemUuid byte order when copying from original structure

Bmju:
This fixes https://github.com/acidanthera/bugtracker/issues/1457 - after the fix the bytes in my OEM specified SystemUUID, as reflected by OC, are in the correct order. On the other hand, if I specify a UUID value in `config.plist` then this code change does not affect that path, and those bytes still appear in the correct order as before.

I would definitely appreciate some code review, but I believe I've identified the correct general location for the fix and the reason for the error.


vit9696:
This is not how it should be fixed, thus I am closing this, but I agree that the issue does exist, and will provide some comments in the original issue.

PR #185, opened 2021-02-01
Add new driver to docs

1alessandro1:


vit9696:
We do not advice this driver, thus it is intentionally not included.

PR #189, opened 2021-02-02
Misc Doc Tweaks

dakanji:
**"One can do XYZ"**, **"One may try XYZ"** and similar, are technically correct but slightly archaic and sound odd in context. Changed a few lingering examples to be more natural.

### DIFF
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -1427,7 +1427,7 @@ To view their current state use \texttt{pmset -g} command in Terminal.
 
   \emph{Note}: While this quirk is required to run older macOS versions
   on platforms with used lower memory it is not compatible with some
-  hardware and macOS 11. In this case one may try to use
+  hardware and macOS 11. In such cases, consider using
   \texttt{EnableSafeModeSlide} instead.
 
 \item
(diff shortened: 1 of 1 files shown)


vit9696:
From my tech literature experience this is more common actually, as "try using" is very informal. I do not mind the changes, but they look even more odd to me.

dakanji:
Let me take another look

Docs/Configuration.tex
```diff
@@ -6799,12 +6799,11 @@ \subsection{Legacy Apple OS}\label{legacyapple}
...
 While newer operating systems can be downloaded over the internet,
 older operating systems did not have installation media for every minor
-release, so to get a compatible distribution one may have to download
-a device-specific image and mod it if necessary. To get the list of
-the bundled device-specific builds for legacy operating systems
-one can visit this archived Apple Support
-\href{https://web.archive.org/web/20170705003629/https://support.apple.com/en-us/HT204319}{article}.
-Since it is not always accurate, the latest versions are listed below.
+release. For compatible distributions on such, download a device-specific
+image and modify this if necessary. Visit this archived Apple Support
```
vit9696, line 6803: modify this sounds awkward. Modify *it*?

PR #191, opened 2021-02-03
DOCS: Document limitations of writing via system Data Hub firmware

Bmju:
Addresses https://github.com/acidanthera/bugtracker/issues/1468


vit9696:
- This is true for essentially any DataHub protocol implementation, thus the legacy Apple hardware part is not quite correct. Many firmwares have this protocol. I suggest you to give Apple hardware as an example.
- Noting that overriding the DataHub protocol will result in the previously stored entries lost feels important to me.

Bmju:
@vit9696 Actually, first point, definitely, changing it. Second point, this is already documented at ProtocolOverrides/DataHub, which presumably someone should look at having been pointed there. Is it needed here too?

Bmju:
As a minor additional point, while I'm in this bit of the documentation, the warning says "This will delete all previous properties if the protocol was already installed.", whereas it is more correct to say something like "If a previous protocol was installed, no properties will be copied over, so all required properties must be specified." (I'm drawing the distinction that if you are dual-booting between OC and non-OC, then using this setting does not actually delete settings in the existing firmware, which one might think was the case or be unsure about on reading the existing text.)

vit9696:
> Second point, this is already documented at ProtocolOverrides/DataHub, which presumably someone should look at having been pointed there. Is it needed here too?

Maybe should explicitly remind then? Like "Refer to the DataHub protocol override description for more details"?

> "If a previous protocol was installed, no Data Hub entries will be copied over and all required entries must be specified in your configuration." (I'm drawing the distinction that if you are dual-booting between OC and non-OC, then using this setting does not actually delete settings in the existing firmware, which one might think was the case or be unsure about on reading the existing warning text.)

Do not mind.

Bmju:
Updated - I did not make the second change, as you've used the same "will be deleted" wording elsewhere, so seemed best to just leave alone as I'm not sure of the details elsewhere!

Bmju:
Should "most" say "essentially all"? I will re-do and re-push with that change.

Bmju:
Done. Ready.

Bmju:
Addressed the second issue also, please check. Many, many thanks, Vitaly. /Mike

Docs/Configuration.tex
```diff
@@ -6559,15 +6566,21 @@ \subsection{ProtocolOverrides Properties}\label{uefiprotoprops}
   \textbf{Type}: \texttt{plist\ boolean}\\
   \textbf{Failsafe}: \texttt{false}\\
   \textbf{Description}: Reinstalls Data Hub protocol with a builtin version.
-  This will delete all previous properties if the protocol was already installed.
+
+  \emph{Note}:  This will discard all previous entries if the protocol was already
```
vit9696, line 6570: Extra space here and below? Sorry.

Bmju: Of course.

Bmju:
Re-pushed with the space mistake fixed. Many thanks.

PR #195, opened 2021-02-06
Some code changes, need review.

cecekpawon:


Include/Acidanthera/Library/OcStringLib.h
```diff
@@ -71,7 +71,7 @@ IsAsciiAlpha (
 
   @retval  TRUE, if character is a white space character.
 **/
-INTN
+BOOLEAN
```
PMheart, line 74: These are to match `isalpha` and `isspace` in `ctype.h`, aren't they?

Library/OcStringLib/OcAsciiLib.c
```diff
@@ -57,7 +57,7 @@ IsAsciiAlpha (
 
   @retval  TRUE, if character is a white space character.
 **/
-INTN
+BOOLEAN
```
PMheart, line 60: And same thing here.

Include/Acidanthera/Library/OcStringLib.h
cecekpawon: Hello, sorry, im not sure here whether this function should match standard ctype or not. Im just refer to function header comment that retval shoud return TRUE if matched.

vit9696: Yeah, better fix the header then. Non-zero if character is a white space character, Zero otherwise.

vit9696:
I cherry-picked the commits that make sense.

Include/Acidanthera/Library/OcStringLib.h
cecekpawon: Thanks for review & merging @PMheart @vit9696 
I will leave this commit & close this pull req to let you guys do for the best regarding this commit.
- peace -

PR #197, opened 2021-02-09
Disable AWAC with this SSDT-AWAC-D

Lorys89:
Hello, I wanted to propose this SSDT as current ssdt awac with _INI method does not disable awac device in my ice lake laptop, with this version it disables it.
https://github.com/Lorys89/DELL_VOSTRO_5401-ICE-LAKE
cc @vit9696 @PMheart


vit9696:
We cannot cover all the hardware variants. While sample SSDTs do cover a wide range of models, they are not meant to be used as is blindly. You may provide a PR adding a comment to the existing SSDT which would mention that updating STAS from a different place may be needed on the machines that already have `_INI` implemented.

PR #198, opened 2021-02-09
Update SSDT-AWAC

Lorys89:
if _ini is present to device awac 
If you want, implement comments to remove _INI method, for newbie users

cc @vit9696 @Andrey1970AppleLife


Lorys89:
@Andrey1970AppleLife @vit9696  the comment needs a change, specifically that should be used if in the dsdt origin there is already Exists a method _INI under SB ok?

Lorys89:
@Andrey1970AppleLife i've seen the change (/) is the universal optimal solution. good jobs

PR #199, opened 2021-02-09
update usbx for laptop

Lorys89:
cc @Andrey1970AppleLife @vit9696

### DIFF
--- a/Docs/AcpiSamples/Source/SSDT-EC-USBX.dsl
+++ b/Docs/AcpiSamples/Source/SSDT-EC-USBX.dsl
@@ -54,7 +54,45 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "SsdtEC", 0x00001000)
     **/
 
     Scope (\_SB)
-    {
+    {    
+       /* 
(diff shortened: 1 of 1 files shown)


Lorys89:
@Andrey1970AppleLife  please expose your opinion, so I understand the reason for the closure of pr.
Thank you

Lorys89:
@vit9696 I studied the USBX device well, looking at ACPI of real pro/air macbooks and imacs SKL and later, the device on laptops is different! does this is born the PR with modification inside the comment. 
Thank you.

vit9696:
The fact that they are "different" does not mean we should follow. Not specifying two fields explicitly by Apple can be convenient for them (Apple) to handle the defaults, but they are not convenient for us, since we need to calculate the values ourselves per each model individually. While the suggested values in USBX often work, you should not rely on that and must be ready to adjust things.

Also, please do not make PRs without prior notice. Firstly raise a discussion in acidanthera/bugtracker, and better group the related issues (e.g. ACPI sample enhancements — list of suggested changes, reasons behind).

Lorys89:
ok, Sorry.

PR #201, opened 2021-02-11
Fix some incorrect spellings of possessive "its" throughout the project

justinkb:
This fixes some wrong language in the documentation (source code comments and Markdown files), but the purpose of the pull request is really to ask whether you'd accept a pull request for a feature I intend to implement, namely a way to blacklist (via config.plist) certain entries, for example via specifying the partition UUID, so that AddBootEntryFromBootOption will not add them to the boot choices. If it is not an upstreamable change, I wouldn't bother creating this feature, since then the maintenance burden for the change is greater than the gain I'd get (my use case is a multiboot situation with Chromium OS, where I prefer a custom Entries entry for booting its grubx64.efi directly, rather than relying on the bcfg/efibootmgr nvram BootOption) as opposed to just manually deleting the bootx64.efi from the Chromium OS drive's ESP after each CROS update (my motherboard's firmware seemingly autogenerates partition style boot options when it detects an ESP with EFI/BOOT/BOOTx64.efi present).

### DIFF
--- a/Include/Acidanthera/Library/OcDeviceTreeLib.h
+++ b/Include/Acidanthera/Library/OcDeviceTreeLib.h
@@ -125,7 +125,7 @@ DTLookupEntry (
 /**
   An Entry Iterator maintains three variables that are of interest to clients.
   First is an "OutermostScope" which defines the outer boundry of the iteration.
-  This is defined by the starting entry and includes that entry plus all of it's
+  This is defined by the starting entry and includes that entry plus all of its
   embedded entries. Second is a "currentScope" which is the entry the iterator is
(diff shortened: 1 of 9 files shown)


vit9696:
Hi, this feature has been requested several times earlier, and we kind of opposed it. Given that you would like not only to use it, but also contribute the code, I think it may well be about the time to push our egos a little, and actually land a feature.

May I ask you to create an issue in the bugtracker (https://github.com/acidanthera/bugtracker/issues), where you could explain the suggested change design-wise (i.e. how one would specify it in the `plist` and how would it work) and code-wise (which changes would you like to make). I think we should be able to agree.

PR #203, opened 2021-02-11
DOCS: Volume icons

Bmju:
https://github.com/acidanthera/bugtracker/issues/1493

- Clarify file locations
- Explain the position regarding Apple's broken method
- Reword existing note about internal/external drives.

As above, in addition to as briefly as I could clearly explaining the issue around `Data` volume icons, I saw and made some small additional edits in exactly the same area that I thought were useful or added to clarity (at least for me) without being significant changes - I hope you will think so too.


Docs/Configuration.tex
```diff
@@ -3092,16 +3092,30 @@ \subsection{Boot Properties}\label{miscbootprops}
...
 
-    For all other entries OpenCore will try to load a volume icon and fallback to the default icon:
+    For all other entries OpenCore will try to load a volume icon by searching
+    as follows, and will fallback to the default icon otherwise:
     \begin{itemize}
     \tightlist
-      \item \texttt{.VolumeIcon.icns} file at \texttt{Preboot} volume directory for APFS (if present).
-      \item \texttt{.VolumeIcon.icns} file at \texttt{Preboot} root for APFS (otherwise).
+      \item \texttt{.VolumeIcon.icns} file at \texttt{Preboot} per-volume directory
+      (\texttt{/System/Volumes/Preboot/-GUID-/} within macOS) for APFS (if present).
```
vit9696, line 3100: Hmm, I slightly do not like the `/System/Volumes/Preboot/-GUID-/` part. `Preboot` is the role of a volume inside an APFS container. How It matters little on how it is mounted in macOS, and in fact you can mount it anywhere, not just here. Maybe better to word it like this?

```
      \item \texttt{.VolumeIcon.icns} file at \texttt{Preboot} volume in per-volume directory
       (\texttt{/System/Volumes/Preboot/\{GUID\}/} when mounted at default location within
       macOS) for APFS (if present).
```

Docs/Configuration.tex
```diff
@@ -3092,16 +3092,30 @@ \subsection{Boot Properties}\label{miscbootprops}
...
+      \item \texttt{.VolumeIcon.icns} file at \texttt{Preboot} root
+      (\texttt{/System/Volumes/Preboot/} within macOS) for APFS (otherwise).
       \item \texttt{.VolumeIcon.icns} file at volume root for other filesystems.
     \end{itemize} \medskip
+    
+    \emph{Note 1}: Apple's boot picker partially supports placing a volume icon file
+    at \texttt{Data} root (\texttt{/System/Volumes/Data/} within macOS). This approach
```
vit9696, line 3107: Here it also makes think that `Data` volume, just like `Preboot`, is a single volume within the container. But it is rather: Preboot, Recovery, System1, Data1, System2, Data2, System3, Data3. Maybe something like this:
```
    \emph{Note 1}: Apple's boot picker partially supports placing a volume icon file
     at operating system's \texttt{Data} volume root (\texttt{/System/Volumes/Data/}
     when mounted at default location within macOS). This approach
```

Docs/Configuration.tex
Bmju: Applied

Docs/Configuration.tex
Bmju: Added word "the" ("the operating system's") and applied 👍

Bmju:
Re-done as a new single commit and force-pushed. Ready to check again. Thanks.

PR #204, opened 2021-02-11
Docs: Fix Itemisation

dakanji:
Addresses acidanthera/bugtracker#1494


Bmju:
Hi @dakanji - I appreciate as per https://github.com/acidanthera/bugtracker/issues/1494#issuecomment-777622744 that the docs already say that the failsafe is an empty string. But with this change, nothing now says what empty string does!

Put another way, this change (or the earlier change, now with this fix) definitely seems to me to remove some information that I would like to know, and unfortunately (unless I'm still really missing something) cannot get from the docs as of now!

Thanks for your attention!

PR #207, opened 2021-02-14
Docs: Slightly reword Apple Inc. warning for clarity

Bmju:
This PR is good to go, if it looks right.

At the end of the para I would have liked to add:

    "It is recommended to continue to use \texttt{Apple\ Inc.} on Apple hardware."

which sounds right to me, and which from [your comment](https://github.com/acidanthera/bugtracker/issues/1497#issuecomment-778550054) you would completely agree with, except that this takes a side in an [ongoing](https://github.com/acidanthera/bugtracker/issues/1497#issuecomment-778742744) [debate](https://github.com/acidanthera/bugtracker/issues/1469#issuecomment-778325679) about what *is* correct to do on Apple hardware, and in which OCLP (who I think in most people's minds are very closely associated with the core OC team) are seemingly taking the opposite view to this, so I thought it best to leave that!


vit9696:
I do not think we should recommend anything. The change is good as it is.

PR #210, opened 2021-02-15
ocvalidate: allow OEM special string in Generic/SystemUUID

Bmju:
- tested!
 - have used "special string OEM" instead of just "OEM" in warning, to try to avoid same sort of ambiguity in wording addressed in https://github.com/acidanthera/OpenCorePkg/pull/208

### DIFF
--- a/Utilities/ocvalidate/ValidatePlatformInfo.c
+++ b/Utilities/ocvalidate/ValidatePlatformInfo.c
@@ -94,8 +94,10 @@ CheckPlatformInfoGeneric (
   }
 
   AsciiSystemUUID     = OC_BLOB_GET (&UserPlatformInfo->Generic.SystemUuid);
-  if (AsciiSystemUUID[0] != '\0' && !AsciiGuidIsLegal (AsciiSystemUUID)) {
-    DEBUG ((DEBUG_WARN, "PlatformInfo->Generic->SystemUUID is borked!\n"));
+  if (AsciiSystemUUID[0] != '\0'
+    && AsciiStrCmp (AsciiSystemUUID, "OEM") != 0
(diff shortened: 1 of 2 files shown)


Utilities/ocvalidate/ValidatePlatformInfo.c
```diff
@@ -94,8 +94,8 @@ CheckPlatformInfoGeneric (
   }
 
   AsciiSystemUUID     = OC_BLOB_GET (&UserPlatformInfo->Generic.SystemUuid);
-  if (AsciiSystemUUID[0] != '\0' && !AsciiGuidIsLegal (AsciiSystemUUID)) {
-    DEBUG ((DEBUG_WARN, "PlatformInfo->Generic->SystemUUID is borked!\n"));
+  if (AsciiSystemUUID[0] != '\0' && AsciiStrCmp (AsciiSystemUUID, "OEM") != 0 && !AsciiGuidIsLegal (AsciiSystemUUID)) {
```
PMheart, line 97: This line is becoming sort of too long. In my opinion, split it into multiple lines. That is to say:
```
if (cond1
  && cond2
  && ...) {
  // statements
}
```
Please make sure that logical operators (e.g. `&&`) are indented by two spaces.

Also, I believe now we should also list this `OEM` check in [ocvalidate README](https://github.com/acidanthera/OpenCorePkg/blob/master/Utilities/ocvalidate/README.md), at `PlatformInfo->Generic` part.

Once done, I think it will be the time to merge this PR. Thanks for your contribution!

Bmju:
Thank you for your feedback - I have made those changes, and squashed down into one new commit off the current master.

PR #216, opened 2021-02-20
Updates to Docs Section 8

dakanji:
Need steer on this bit where something seems to have been skipped:
Screen Shot 2021-02-20 at 21 00 18

The one seems odd: `Locate Device Path of resulting Device Path`
Screen Shot 2021-02-20 at 21 15 18

### DIFF
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -120,7 +120,7 @@ This document provides information on the
 configuration file format used to set up the correct functioning of the macOS
 operating system. It is to be read as the official clarification of expected
 OpenCore behaviour. All deviations, if found in published OpenCore releases,
-shall be considered to be documentation or implementation flaws which should be
+shall be considered to be documentation or implementation issues which should be
 reported via the \href{https://github.com/acidanthera/bugtracker}{Acidanthera Bugtracker}.
 An errata sheet is available in
 \href{https://github.com/acidanthera/OpenCorePkg/blob/master/Docs/Errata/Errata.pdf}{OpenCorePkg repository}.
@@ -206,7 +206,7 @@ with the full list available in
 \item
   \texttt{plist\ multidata} --- value cast to data by the implementation.
   Permits passing \texttt{plist\ string}, in which case the result is
-  represented by a null-terminated sequence of bytes (aka C string),
+  represented by a null-terminated sequence of bytes (C string),
   \texttt{plist\ integer}, in which case the result is represented by
   \emph{32-bit} little endian sequence of bytes in two's complement
   representation, \texttt{plist\ boolean}, in which case the value is
@@ -266,7 +266,7 @@ with the full list available in
 \item
   \texttt{fatal\ behaviour} --- behaviour leading to boot termination.
   Implementations shall prevent the boot process from continuing until
-  the next host system boot. It is permitted, but not required, to
+  the host system is restarted. It is permitted, but not required, to
   execute cold reboots or to show warning messages in such cases.
 \item
   \texttt{undefined\ behaviour} --- behaviour not prescribed by this
@@ -368,7 +368,7 @@ The file is structured to group related elements in subsections as follows:
 \item
   \texttt{Patch} provides support for data modification.
 \item
-  \texttt{Quirks} provides support for specific hacks.
+  \texttt{Quirks} provides support for specific workarounds.
 \end{itemize}
 
 Root configuration entries consist of the following:
@@ -506,10 +506,10 @@ Available entries include:
 \item
   \texttt{BOOTx64.efi} or \texttt{BOOTIa32.efi} \\
   Initial bootstrap loaders, which load \texttt{OpenCore.efi}. \texttt{BOOTx64.efi}
-  is loaded by the firmware by default according to the UEFI specification. However, it can
-  also be renamed and put in a custom location to allow OpenCore to coexist with operating
-  systems using \texttt{BOOTx64.efi} as their own loaders (e.g. Windows). See
-  the \texttt{LauncherOption} property for more details.
+  is loaded by the firmware by default consistent with the UEFI specification. However,
+  it may also be renamed and put in a custom location to allow OpenCore coexist alongside
+  operating systems, such as Windows, that use \texttt{BOOTx64.efi} files as their loaders.
+  Refer to the \texttt{LauncherOption} property for more details.
 \item
   \texttt{boot} \\
   Duet bootstrap loader, which initialises the UEFI environment on legacy BIOS firmware
@@ -821,10 +821,9 @@ ACPI changes apply globally (to every operating system) with the following effec
 \end{itemize}
 
 Applying the changes globally resolves the problems of incorrect operating system
-detection, which is not possible before the operating system boots according to
-the ACPI specification, operating system chainloading, and harder ACPI debugging.
-For this reason it may be required to carefully use \texttt{\_OSI} method when
-writing the changes.
+detection (consistent with the ACPI specification, not possible before the operating
+system boots), operating system chainloading, and difficult ACPI debugging. Hence,
+more attention may be required when writing changes to \texttt{\_OSI}.
 
 Applying the patches early makes it possible to write so called ``proxy'' patches,
 where the original method is patched in the original table and is implemented in
@@ -908,8 +907,8 @@ quality of the suggested solutions will be variable.
   Example values include \texttt{DSDT.aml}, \texttt{SubDir/SSDT-8.aml},
   \texttt{SSDT-USBX.aml}, etc.
 
-  ACPI table load order follows the item order in the array. All ACPI tables
-  load from \texttt{OC/ACPI} directory.
+  The ACPI table load order follows the item order in the array. ACPI tables
+  are loaded from the \texttt{OC/ACPI} directory.
 
   \textbf{Note}: All tables apart from tables with a \texttt{DSDT} table identifier
   (determined by parsing data, not by filename) insert new tables into the ACPI stack.
@@ -1060,7 +1059,7 @@ In most cases, ACPI patches are not useful and are harmful:
   \href{https://github.com/acidanthera/WhateverGreen}{WhateverGreen}.
 \item
   Avoid patching \texttt{\_OSI} to support a higher feature set level
-  whenever possible. While this enables a number of hacks on APTIO
+  whenever possible. While this enables a number of workarounds on APTIO
   firmware, it typically results in a need for additional patches. Modern
   firmware generally does not need this and smaller patches work well on
   firmware that does. However, laptop vendors often rely on this method to
@@ -1068,7 +1067,7 @@ In most cases, ACPI patches are not useful and are harmful:
   adjustment and custom feature additions.
 \item
   Avoid patching embedded controller event \texttt{\_Qxx} just to enable
-  brightness keys. The conventional process to find these keys usually involves
+  brightness keys. The conventional process to find these keys typically involves
   significant modifications to DSDT and SSDT files and in addition, the debug kext
   is not stable on newer systems. Please use the built-in brightness key discovery in
   \href{https://github.com/acidanthera/BrightnessKeys}{BrightnessKeys} instead.
@@ -1119,10 +1118,10 @@ patching or the padding of \texttt{NOP} to the remaining area could be considere
   \textbf{Type}: \texttt{plist\ boolean}\\
   \textbf{Failsafe}: \texttt{false}\\
   \textbf{Description}: Cleanup ACPI header fields to workaround macOS
-  ACPI implementation bug causing boot crashes. Reference:
+  ACPI implementation flaws that result in boot crashes. Reference:
   \href{https://alextjam.es/debugging-appleacpiplatform/}{Debugging
-  AppleACPIPlatform on 10.13} by Alex James aka theracermaster. The
-  issue is fixed in macOS Mojave (10.14).
+  AppleACPIPlatform on 10.13} by Alex James (also known as theracermaster).
+  The issue was fixed in macOS Mojave (10.14).
 
 \item
   \texttt{RebaseRegions}\\
@@ -1134,7 +1133,7 @@ patching or the padding of \texttt{NOP} to the remaining area could be considere
   ACPI tables are often generated dynamically by the underlying firmware
   implementation. Among the position-independent code, ACPI tables may
   contain the physical addresses of MMIO areas used for device
-  configuration, usually grouped by region (e.g.
+  configuration, typically grouped by region (e.g.
   \texttt{OperationRegion}). Changing firmware settings or hardware
   configuration, upgrading or patching the firmware inevitably leads to
   changes in dynamically generated ACPI code, which sometimes results in
@@ -1229,7 +1228,8 @@ should be met before starting:
 When debugging sleep issues, Power Nap and automatic power off (which appear to sometimes cause
 wake to black screen or boot loop issues on older platforms) may be temporarily disabled.
 The specific issues may vary, but generally ACPI tables should be looked at first.
-Here is an example of a bug found in some
+
+Here is an example of a defect found in some
 \href{http://www.insanelymac.com/forum/topic/329624-need-cmos-reset-after-sleep-only-after-login/#entry2534645}{Z68 motherboards}.
 To turn Power Nap and the others off, run the following commands in Terminal:
 \begin{lstlisting}[label=powernap, style=ocbash]
@@ -1404,10 +1404,10 @@ To view their current state, use the \texttt{pmset -g} command in Terminal.
     \item KASLR (slide) is unsupported (this is macOS 10.7 or older)
   \end{itemize}
 
-  This quirk requires \texttt{ProvideCustomSlide} to also be enabled
-  and typically needs \texttt{AvoidRuntimeDefrag} to be set for this to
-  work correctly. Hibernation is not supported when booting with a relocation
-  block (but relocation block is not always used when the quirk is enabled).
+  This quirk requires \texttt{ProvideCustomSlide} to be enabled and
+  typically also requires enabling \texttt{AvoidRuntimeDefrag} to function
+  correctly. Hibernation is not supported when booting with a relocation
+  block, which will only be used if required when the quirk is enabled.
 
   \emph{Note}: While this quirk is required to run older macOS versions
   on platforms with used lower memory, it is not compatible with some
@@ -1469,7 +1469,7 @@ To view their current state, use the \texttt{pmset -g} command in Terminal.
   This quirk requires \texttt{OC\_FIRMWARE\_RUNTIME} protocol implemented
   in \texttt{OpenRuntime.efi}.
 
-  \emph{Note}: This quirk can also be used as an ad hoc workaround for flawed UEFI
+  \emph{Note}: This quirk can also be used as an ad hoc workaround for defective UEFI
   runtime services implementations that are unable to write variables to NVRAM
   and results in operating system failures.
 
@@ -1485,7 +1485,7 @@ To view their current state, use the \texttt{pmset -g} command in Terminal.
   \href{https://docs.microsoft.com/en-us/windows-hardware/design/device-experiences/oem-uefi#hibernation-state-s4-transition-requirements}{preserving}
   runtime memory size and location after S4 wake.
 
-  \emph{Note}: This may be used to workaround flawed memory map implementations on older,
+  \emph{Note}: This may be used to workaround defective memory map implementations on older,
   rare legacy hardware. Examples of such hardware are Ivy Bridge laptops with Insyde firmware
   such as the Acer V3-571G. Do not use this option without a full understanding of the implications.
 
@@ -1498,9 +1498,14 @@ To view their current state, use the \texttt{pmset -g} command in Terminal.
   This option is relevant to users with issues booting to safe mode
   (e.g. by holding \texttt{shift} or with using the \texttt{-x} boot argument). By
   default, safe mode forces \texttt{0} slide as if the system was launched with the
-  \texttt{slide=0} boot argument. This quirk attempts to patch \texttt{boot.efi} to remove
-  this limitation and let some other value (from \texttt{1} to \texttt{255}) be used.
-  This quirk requires enabling \texttt{ProvideCustomSlide}.
+  \texttt{slide=0} boot argument.
+
+  \begin{itemize}
+    \tightlist
+    \item This quirk attempts to patch the \texttt{boot.efi} file to remove this limitation
+    and to allow using other values (from \texttt{1} to \texttt{255} inclusive).
+    \item This quirk requires enabling \texttt{ProvideCustomSlide}.
+    \end{itemize}
 
   \emph{Note}: The need for this option is dependent on the availability of safe mode.
   It can be enabled when booting to safe mode fails.
@@ -1544,12 +1549,12 @@ To view their current state, use the \texttt{pmset -g} command in Terminal.
   \begin{itemize}
     \tightlist
     \item The CSM region can be marked as boot services code, or data, which
-      leaves it as free memory for XNU kernel.
+      leaves it as free memory for the XNU kernel.
     \item MMIO regions can be marked as reserved memory and stay unmapped.
       They may however be required to be accessible at runtime for NVRAM support.
   \end{itemize}
 
-  This quirk attempts to fix various types of these regions, e.g. ACPI NVS for
+  This quirk attempts to fix the types of these regions, e.g. ACPI NVS for
   CSM or MMIO for MMIO.
 
   \emph{Note}: The need for this quirk is determined by artifacts, sleep
@@ -1567,7 +1572,7 @@ To view their current state, use the \texttt{pmset -g} command in Terminal.
 
   \emph{Note}: This quirk attempts to avoid issues with NVRAM implementations with
   fragmentation issues, such as on the \texttt{MacPro5,1} as well as on certain
-  Insyde firmware without garbage collection or with faulty garbage collection.
+  Insyde firmware without garbage collection or with defective garbage collection.
 
 \item
   \texttt{ProtectUefiServices}\\
@@ -1577,9 +1582,9 @@ To view their current state, use the \texttt{pmset -g} command in Terminal.
 
   Some modern firmware, including on virtual machines such as VMware,
   may update pointers to UEFI services during driver loading and related actions.
-  Consequently, this directly breaks other quirks that affect memory management,
+  Consequently, this directly obstructs other quirks that affect memory management,
   such as \texttt{DevirtualiseMmio}, \texttt{ProtectMemoryRegions}, or \texttt{RebuildAppleMemoryMap},
-  and may also break other quirks depending on the scope of such.
+  and may also obstruct other quirks depending on the scope of such.
 
   \emph{Note}: On VMware, the need for this quirk may be determined by the appearance of the
   ``Your Mac OS guest might run unreliably with more than one virtual core.'' message.
@@ -1615,8 +1620,8 @@ To view their current state, use the \texttt{pmset -g} command in Terminal.
 
   \emph{Note}: The need for this quirk is determined by random boot failures
   when \texttt{ProvideCustomSlide} is enabled and the randomized slide falls
-  into the unavailable range. When \texttt{AppleDebug} is enabled, usually the
-  debug log may contain messages such as \texttt{AAPL: [EB|`LD:LKC] \} Err(0x9)}.
+  into the unavailable range. When \texttt{AppleDebug} is enabled, the
+  debug log typically contains messages such as \texttt{AAPL: [EB|`LD:LKC] \} Err(0x9)}.
   To find the optimal value, append \texttt{slide=X}, where \texttt{X} is the slide value,
   to the \texttt{boot-args} and select the largest one that does not result in boot failures.
 
@@ -1651,7 +1656,7 @@ To view their current state, use the \texttt{pmset -g} command in Terminal.
 
   \emph{Note 2}: The need for this quirk is determined by early boot failures.
   This quirk replaces \texttt{EnableWriteUnprotector} on firmware supporting
-  Memory Attribute Tables (MAT). This quirk is usually unnecessary when using
+  Memory Attribute Tables (MAT). This quirk is typically unnecessary when using
   \texttt{OpenDuetPkg} but may be required to boot macOS 10.6, and earlier, for
   reasons that are as yet unclear.
 
@@ -1674,9 +1679,9 @@ To view their current state, use the \texttt{pmset -g} command in Terminal.
   \textbf{Failsafe}: \texttt{false}\\
   \textbf{Description}: Report macOS being loaded through OS Info for any OS.
 
-  This quirk is useful on Mac firmware, which behaves differently for different OS.
-  For example, it is supposed to enable Intel GPU in Windows and Linux in some
-  dual-GPU MacBook models.
+  This quirk is useful on Mac firmware, which loads different operating systems
+  with different hardware configurations. For example, it is supposed to enable
+  Intel GPU in Windows and Linux in some dual-GPU MacBook models.
 
 \item
   \texttt{SyncRuntimePermissions}\\
@@ -1868,7 +1873,7 @@ blocking.
   drivers, which are not cached otherwise. The issue normally affects older
   operating systems, where various dependency kexts, such as \texttt{IOAudioFamily}
   or \texttt{IONetworkingFamily} may not be present in the kernel cache by default.
-  Kernel driver load order follows the item order in the array, thus the dependencies
+  The kernel driver load order follows the item order in the array, thus the dependencies
   should be written prior to their consumers. \texttt{Force} happens before
   \texttt{Add}.
 
@@ -1907,8 +1912,8 @@ blocking.
 \item
   \texttt{Arch}\\
   \textbf{Type}: \texttt{plist\ string}\\
-  \textbf{Failsafe}: \texttt{Any}\\
-  \textbf{Description}: Kext architecture (\texttt{Any}, \texttt{i386}, \texttt{x86\_64}).
+  \textbf{Failsafe}: \texttt{Any} (Apply to any supported architecture)\\
+  \textbf{Description}: Kext architecture (\texttt{i386}, \texttt{x86\_64}).
 
 \item
   \texttt{BundlePath}\\
@@ -2004,8 +2009,8 @@ blocking.
 \item
   \texttt{Arch}\\
   \textbf{Type}: \texttt{plist\ string}\\
-  \textbf{Failsafe}: \texttt{Any}\\
-  \textbf{Description}: Kext block architecture (\texttt{Any}, \texttt{i386}, \texttt{x86\_64}).
+  \textbf{Failsafe}: \texttt{Any} (Apply to any supported architecture)\\
+  \textbf{Description}: Kext block architecture (\texttt{i386}, \texttt{x86\_64}).
 
 \item
   \texttt{Comment}\\
@@ -2056,12 +2061,12 @@ blocking.
   \textbf{Description}: Sequence of \texttt{EAX}, \texttt{EBX}, \texttt{ECX},
   \texttt{EDX} values to replace \texttt{CPUID (1)} call in XNU kernel.
 
-  This property primarily serves for three needs:
+  This property primarily meets three requirements:
 
   \begin{itemize}
     \tightlist
-    \item Enabling support of an unsupported CPU model (e.g. Intel Pentium).
-    \item Enabling support of a CPU model that is not yet supported by a specific version of macOS which usually is old.
+    \item Enabling support for an unsupported CPU model (e.g. Intel Pentium).
+    \item Enabling support for a CPU model not yet supported by a specific version of macOS (typically old versions).
     \item Enabling XCPM support for an unsupported CPU variant.
   \end{itemize}
 
@@ -2086,7 +2091,7 @@ blocking.
     \texttt{Cpuid1Mask}: \texttt{FF FF FF FF 00 00 00 00 00 00 00 00 00 00 00 00}
   \end{itemize}
 
-  \emph{Note 4}: Note that the following configurations are unsupported by XCPM (at least out of the box):
+  \emph{Note 4}: Be aware that the following configurations are unsupported by XCPM (at least out of the box):
 
   \begin{itemize}
     \tightlist
@@ -2094,7 +2099,7 @@ blocking.
       and recommends legacy power management for these CPUs. \texttt{\_xcpm\_bootstrap}
       should manually be patched to enforce XCPM on these CPUs instead of this option.
     \item Low-end CPUs (e.g. Haswell+ Pentium) as they are not supported properly by macOS.
-      Legacy hacks for older models can be found in the \texttt{Special NOTES} section of
+      Legacy workarounds for older models can be found in the \texttt{Special NOTES} section of
       \href{https://github.com/acidanthera/bugtracker/issues/365}{acidanthera/bugtracker\#365}.
   \end{itemize}
 
@@ -2118,7 +2123,7 @@ blocking.
   \texttt{NullCpuPowerManagement.kext} for CPUs without native power
   management driver in macOS.
 
-  \emph{Note 2}: While this option is usually needed to disable \texttt{AppleIntelCpuPowerManagement}
+  \emph{Note 2}: While this option is typically needed to disable \texttt{AppleIntelCpuPowerManagement}
   on unsupported platforms, it can also be used to disable this kext in other situations
   (e.g. with \texttt{Cpuid1Data} left blank).
 
@@ -2146,8 +2151,8 @@ blocking.
 \item
   \texttt{Arch}\\
   \textbf{Type}: \texttt{plist\ string}\\
-  \textbf{Failsafe}: \texttt{Any}\\
-  \textbf{Description}: Kext architecture (\texttt{Any}, \texttt{i386}, \texttt{x86\_64}).
+  \textbf{Failsafe}: \texttt{Any} (Apply to any supported architecture)\\
+  \textbf{Description}: Kext architecture (\texttt{i386}, \texttt{x86\_64}).
 
 \item
   \texttt{BundlePath}\\
@@ -2217,8 +2222,8 @@ blocking.
 \item
   \texttt{Arch}\\
   \textbf{Type}: \texttt{plist\ string}\\
-  \textbf{Failsafe}: \texttt{Any}\\
-  \textbf{Description}: Kext patch architecture (\texttt{Any}, \texttt{i386}, \texttt{x86\_64}).
+  \textbf{Failsafe}: \texttt{Any} (Apply to any supported architecture)\\
+  \textbf{Description}: Kext patch architecture (\texttt{i386}, \texttt{x86\_64}).
 
 \item
   \texttt{Base}\\
@@ -2398,7 +2403,7 @@ blocking.
 
   \emph{Note}: While this may increase the performance, this patch is strongly discouraged
   on all systems but those explicitly dedicated to scientific or media calculations.
-  In general only certain Xeon models benefit from the patch.
+  Only certain Xeon models typically benefit from the patch.
 
 \item
   \texttt{CustomSMBIOSGuid}\\
@@ -2417,8 +2422,8 @@ blocking.
   which may conflict with the firmware implementation.
 
   \emph{Note}: This option is a preferred alternative to deleting \texttt{DMAR}
-  ACPI table and disabling VT-d in firmware preferences, which does not break
-  VT-d support in other systems in case they need it.
+  ACPI table and disabling VT-d in firmware preferences, which does not obstruct
+  VT-d support in other systems in case they need this.
 
 \item
   \texttt{DisableLinkeditJettison}\\
@@ -2467,7 +2472,7 @@ blocking.
   internal disk icons for all AHCI disks.
 
   \emph{Note}: This option should be avoided whenever possible. Modern firmware
-  usually have compatible AHCI controllers.
+  typically have compatible AHCI controllers.
 
 \item
   \texttt{ForceSecureBootScheme}\\
@@ -2486,8 +2491,8 @@ blocking.
   \textbf{Requirement}: 10.10\\
   \textbf{Description}: Increases 32-bit PCI bar size in IOPCIFamily from 1 to 4 GBs.
 
-  \emph{Note}: This option should be avoided whenever possible. In general the necessity
-  of this option means misconfigured or broken firmware.
+  \emph{Note}: This option should be avoided whenever possible. A need for this option
+  indicates misconfigured or defective firmware.
 
 \item
   \texttt{LapicKernelPanic}\\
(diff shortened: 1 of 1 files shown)


dakanji:
Also picked up some requested changes to #213 that were missed.

Docs/Configuration.tex
```diff
@@ -2487,7 +2487,7 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
   \textbf{Description}: Increases 32-bit PCI bar size in IOPCIFamily from 1 to 4 GBs.
 
   \emph{Note}: This option should be avoided whenever possible. In general the necessity
-  of this option means misconfigured or broken firmware.
+  of this option means misconfigured or faulty firmware.
```
Bmju, line 2491: 'faulty' sounds too close  to gone wrong since purchase (&/or in many instances of the same thing it works, but in yours it doesn't); I suspect 'broken' (as informal technical usage for incorrectly programmed/designed) is more accurate to what was meant.

Docs/Configuration.tex
```diff
@@ -2765,126 +2762,129 @@ \section{Misc}\label{misc}
...
-This section contains miscellaneous configuration affecting OpenCore operating system loading behaviour
-as well as other entries, which do not go to any other section.
+This section contains miscellaneous configuration options that do not readily fit into other sections.
+These options affect the loading behaviour of the OpenCore operating system in addition to other items.
 
-OpenCore tries to follow ``bless'' model also known as ``Apple Boot Policy''. The primary specialty of
-``bless'' model is to allow embedding boot options within the file system (and be accessible through a
-specialised driver) as well as supporting a broader range of predefined boot paths compared to the
-removable media list found in the UEFI specification.
+OpenCore broadly follows a ``bless'' model also known as the ``Apple Boot Policy''. The primary purpose of
```
Bmju, line 2768: It should be '**the** bless model' at every occurrence, I think (it's Apple specific).

Docs/Configuration.tex
```diff
@@ -2765,126 +2762,129 @@ \section{Misc}\label{misc}
...
-may contain arguments compromising the operating system, which is undesired once secure boot
-is enabled.
+\emph{Note 2}: UEFI variable ``boot options'' boot arguments will be removed, if present, as
+they may contain arguments that can compromise the operating system, which is undesirable when
+secure boot is enabled.
 
-\emph{Note 3}: Some operating systems, namely Windows, will create their boot option and
-mark it as top most upon first boot or after NVRAM Reset. When this happens default boot
-entry choice will update till next manual reconfiguration.
+\emph{Note 3}: Some operating systems, such as Windows, may create a boot option and mark it as
```
Bmju, line 2862: Agree that it's 'such as', not namely. Ubuntu does the same!

Bmju: Btw, I hope it's helpful even if I only pick out the odd thing that I spot? As probably not going to be able to read this one and ongoing in as much detail as I volunteered myself to do on the last.

vit9696:
By creating a custom boot option… without in.

The second one is correct, but can be reworded. Obtain device handle by locating the device path with preciously fixed device path.

dakanji:
> the original is clearer

Left as is.

Andrey1970AppleLife:
Return back \texttt{OC/ACPI}

Docs/Configuration.tex
```diff
@@ -821,10 +821,9 @@ \subsection{Introduction}\label{acpiintro}
 \end{itemize}
 
 Applying the changes globally resolves the problems of incorrect operating system
-detection, which is not possible before the operating system boots according to
-the ACPI specification, operating system chainloading, and harder ACPI debugging.
-For this reason it may be required to carefully use \texttt{\_OSI} method when
-writing the changes.
+detection (consistent with the ACPI specification, not possible before the operating
+system boots), operating system chainloading, and difficult ACPI debugging. Hence,
+carefully using the \texttt{\_OSI} method may be required when writing changes.
```
PMheart, line 826: How about `Hence, more attention may be required when writing changes to \texttt{\_OSI}.`?

Docs/Configuration.tex
```diff
@@ -1498,9 +1498,14 @@ \subsection{Quirks Properties}\label{booterpropsquirks}
...
-  \texttt{slide=0} boot argument. This quirk attempts to patch \texttt{boot.efi} to remove
-  this limitation and let some other value (from \texttt{1} to \texttt{255}) be used.
-  This quirk requires enabling \texttt{ProvideCustomSlide}.
+  \texttt{slide=0} boot argument.
+
+  \begin{itemize}
+    \tightlist
+    \item This quirk attempts to patch the \texttt{boot.efi} file to remove this limitation and
+      to allow using other values (\texttt{1} to \texttt{255} inclusive).
+    \item This quirk requires enabling \texttt{ProvideCustomSlide}.
```
PMheart, line 1507: `This quirk requires \texttt{ProvideCustomSlide} to be enabled.` sounds better?

Docs/Configuration.tex
```diff
@@ -2534,39 +2539,40 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
...
-  once the fragmentation increases. This means that trimming on these SSDs
-  will be broken soon after the installation, causing extra wear to the flash.
+  on the drive is, the more commands are necessary to trim
+  all the free space.
+
+  Depending on the SSD controller and the level of drive fragmenation, the trim
+  procedure may take a considerable amount of time, causing noticeable boot slowdown.
+  The APFS driver explicitly ignores previously unmapped areas and repeatedly trims
+  them on boot. To mitigate against such boot slowdowns, the macOS driver introduced a
+  timeout (\texttt{9.999999} seconds) that stops the trim operation when not completed in
```
PMheart, line 2556: `when not finished in time.`?

Docs/Configuration.tex
```diff
@@ -2534,39 +2539,40 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
...
+  time.
+
+  On several controllers, such as from Samsung, where the deallocation process is relatively
```
PMheart, line 2559: There should not be `from` I believe?

Docs/Configuration.tex
```diff
@@ -2534,39 +2539,40 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
...
+  slow, this timeout can be reached very quickly. Essentially, it means that the level of
+  fragmentation is high, macOS will attempt to trim the same lower blocks that have
```
PMheart, line 2561: `thus macOS will...`

Docs/Configuration.tex
```diff
@@ -2534,39 +2539,40 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
...
+  previously been deallocated, but not never have enough time to deallocate higher
```
PMheart, line 2562: `but never`...

Docs/Configuration.tex
```diff
@@ -2534,39 +2539,40 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
...
+  blocks. The outcome is that trimming on such SSDs will be non-functional soon
+  after installation, resulting in additional wear on the flash.
 
   One way to workaround the problem is to increase the timeout to a very
```
PMheart, line 2566: `extremely high value` maybe? Very sounds like spoken English.

Docs/Configuration.tex
```diff
@@ -2766,126 +2775,135 @@ \section{Misc}\label{misc}
 
 \subsection{Introduction}\label{miscintro}
 
-This section contains miscellaneous configuration affecting OpenCore operating system loading behaviour
-as well as other entries, which do not go to any other section.
+This section contains miscellaneous configuration options that affect the loading behaviour of the OpenCore
+operating system in addition to other configuration options that do not readily fit into other sections.
```
PMheart, line 2779: Why repeatedly say `configuration options` here? Sorry, I do not get the point.

Docs/Configuration.tex
```diff
@@ -2766,126 +2775,135 @@ \section{Misc}\label{misc}
...
-Scan process starts with obtaining all the partitions filtered with ``Scan policy''. Each partition may
-produce multiple primary and alternate options. Primary options describe operating systems installed
-on this media. Alternate options describe recovery options for the operating systems on the media.
-It is possible for alternate options to exist without primary options and vice versa. Be warned
-that the options may not necessarily describe the operating systems on the same partition.
-Each primary and alternate option can be an auxiliary option or not. Refer to the \texttt{HideAuxiliary}
-section for more details.
+The scan process starts with enumerating all available partitions, filtered based on the \texttt{Scan policy}.
+Each partition may generate multiple primary and alternate options. Primary options represent operating systems
+installed on the media while alternate options represent recovery options for the operating systems on the media.
```
PMheart, line 2793: There should be a comma before `while alternate options...` I guess?

Docs/Configuration.tex
```diff
@@ -2766,126 +2775,135 @@ \section{Misc}\label{misc}
...
-  \item For file device paths check presence on the file system directly.
+  \item Obtain the device handle by locating the device path of the resulting device path (ignore it on failure).
+  \item Locate the device handle in the list of partition handles (ignore it if missing).
+  % To determine device path type we can use LocateDevicePath RemainingDevicePath argument.
+  % Just check whether it points to the END device path.
+  \item For disk device paths (not specifying a bootloader), execute ``bless'' (may return > 1 entry).
+  \item For file device paths, check for presence on the file system directly.
   % Just kill all \EFI\APPLE\ paths.
-  \item On OpenCore boot partition exclude all OpenCore bootstrap files by header checks.
+  \item On the OpenCore boot partition, exclude all OpenCore bootstrap files by header checks.
```
PMheart, line 2824: Sorry, I am going to have my breakfast now. Let me leave a mark here so that I would be back to it once finished. :P

Docs/Configuration.tex
```diff
@@ -2766,126 +2775,135 @@ \section{Misc}\label{misc}
...
   The option will become auxiliary for some types (e.g. Apple HFS recovery).
   \end{itemize}
 \item For each partition handle:
   \begin{itemize}
-  \item If partition handle is marked as \textit{unused} execute ``bless'' primary option list retrieval. \\
-    In case \texttt{BlessOverride} list is set, not only standard ``bless'' paths will be found but
-    also custom ones.
-  \item On OpenCore boot partition exclude all OpenCore bootstrap files by header checks.
+  \item If the partition handle is marked as \textit{unused}, execute ``bless'' primary option list retrieval. \\
+    In case \texttt{BlessOverride} list is set, both standard and custom ``bless'' paths will be found.
```
PMheart, line 2833: maybe `In case a \texttt{BlessOverride} list is set`?

Docs/Configuration.tex
```diff
@@ -2766,126 +2775,135 @@ \section{Misc}\label{misc}
...
+The display order as follows:
 
 \begin{itemize}
 \tightlist
-\item Alternate options follow corresponding primary options, i.e. Apple recovery will be following the
+\item Alternate options follow corresponding primary options. That is, Apple recovery options will follow the
   relevant macOS option whenever possible.
-\item Options will be listed in file system handle firmware order to maintain an established order across
-  the reboots regardless of the chosen operating system for loading.
+\item Options will be listed file system handle firmware order to maintain an established order across
```
PMheart, line 2855: Any reason why `in` has been deleted? I think it should remain there.

Docs/Configuration.tex
```diff
@@ -2766,126 +2775,135 @@ \section{Misc}\label{misc}
...
 \item Boot option through the picker or without it depending on the \texttt{ShowPicker} option.
 \item Show picker on failure otherwise.
 \end{itemize}
 
-\emph{Note 1}: This process is meant to work reliably only when \texttt{RequestBootVarRouting}
+\emph{Note 1}: This process will only work reliably when the \texttt{RequestBootVarRouting}
 option is enabled or the firmware does not control UEFI boot options (\texttt{OpenDuetPkg} or
-custom BDS). Without \texttt{LauncherOption} it also is possible that other operating systems
-overwrite OpenCore, make sure to enable it when planning to use them.
+custom BDS). When \texttt{LauncherOption} is not enabled, other operating systems can
```
PMheart, line 2873: Change `can` to `may` to emphasise probability?

Docs/Configuration.tex
```diff
@@ -2766,126 +2775,135 @@ \section{Misc}\label{misc}
...
+  \textbf{Description}: Add custom scanning paths through the bless model.
 
   Designed to be filled with \texttt{plist\ string} entries containing
-  absolute UEFI paths to customised bootloaders, for example,
+  absolute UEFI paths to customised bootloaders such as
   \texttt{\textbackslash EFI\textbackslash debian\textbackslash grubx64.efi}
-  for Debian bootloader. This allows unusual boot paths to be automatically
-  discovered by the boot picker. Designwise they are equivalent to predefined blessed path, such as
+  for the Debian bootloader. This allows non-standard boot paths to be automatically
+  discovered by the boot picker. Designwise, they are equivalent to predefined blessed paths, such as
```
PMheart, line 2903: Change `Designwise` to `In terms of design` maybe?

Docs/Configuration.tex
```diff
@@ -2986,9 +3006,9 @@ \subsection{Boot Properties}\label{miscbootprops}
   \texttt{HideAuxiliary}\\
   \textbf{Type}: \texttt{plist\ boolean}\\
   \textbf{Failsafe}: \texttt{false}\\
-  \textbf{Description}: Hides auxiliary entries from picker menu by default.
+  \textbf{Description}: Hides auxiliary entries from the picker menu unless overriden.
```
PMheart, line 3009: Some verbs of `Description` part have imperative forms. It should always be consistent I guess. So, change `Hides` to `Hide`.

Docs/Configuration.tex
```diff
@@ -2986,9 +3006,9 @@ \subsection{Boot Properties}\label{miscbootprops}
...
 
-  An entry is considered auxiliary when at least one of the following applies:
+  An entry is considered to be an auxiliary entry when one or more of the following apply:
```
PMheart, line 3011: The original statement seems tidier in my opinion.

Docs/Configuration.tex
```diff
@@ -2998,66 +3018,68 @@ \subsection{Boot Properties}\label{miscbootprops}
     \item Entry is system (e.g. \texttt{Reset NVRAM}).
   \end{itemize}
 
-  To see all entries picker menu needs to be reloaded in extended mode by pressing
-  \texttt{Spacebar} key. Hiding auxiliary entries may increase boot performance
-  for multidisk systems.
-
+  To display all entries, the picker menu can be reloaded into ``Extended Mode'' by pressing the
+  \texttt{Spacebar} key. Hiding auxiliary entries may increase boot performance on multidisk systems.
```
PMheart, line 3022: I am not a native English speaker, but I still think that `multi` and `disk` should be connected with a dash. i.e. `multi-disk`.

Docs/Configuration.tex
```diff
@@ -2998,66 +3018,68 @@ \subsection{Boot Properties}\label{miscbootprops}
...
-  \emph{Note 1}: Some types of firmware may have faulty NVRAM, no boot option support, or other
-  incompatibilities. While unlikely, the use of this option may even cause boot failures.
-  This option should be used without any warranty exclusively on the boards known to be compatible.
-  Check \href{https://github.com/acidanthera/bugtracker/issues/1222}{acidanthera/bugtracker\#1222}
+  This option allows integration with third-party operating system installation and upgrades
+  (which may overwrite the \texttt{\textbackslash EFI\textbackslash BOOT\textbackslash BOOTx64.efi}
+  file). The BOOTx64.efi file is no longer used for bootstrapping OpenCore if a custom option is created.
+  The custom path used for bootstrapping can be specified by using the \texttt{LauncherPath} option.
+
+  \emph{Note 1}: Some types of firmware may have defective NVRAM implementation, no boot option
```
PMheart, line 3054: I think that there should be a transition word. `hence` would make sense to me.

Docs/Configuration.tex
```diff
@@ -3138,16 +3160,15 @@ \subsection{Boot Properties}\label{miscbootprops}
...
 
-  For macOS bootloader screen reader preference is set in \texttt{preferences.efires}
-  archive in \texttt{isVOEnabled.int32} file and is controlled by the operating system.
-  For OpenCore screen reader support this option is an independent equivalent.
-  Toggling screen reader support in both OpenCore boot picker and macOS bootloader
-  FileVault 2 login window can also be done with \texttt{Command} + \texttt{F5} key
+  For the macOS bootloader, screen reader preference is set in the \texttt{preferences.efires}
+  archive in the \texttt{isVOEnabled.int32} file and is controlled by the operating system.
+  For OpenCore screen reader support, this option is an independent equivalent.
+  Toggling screen reader support in both the OpenCore boot picker and  the macOS bootloader
```
PMheart, line 3166: There is one more space in the second `the`. Drop it please.

Docs/Configuration.tex
```diff
@@ -3290,33 +3315,33 @@ \subsection{Debug Properties}\label{miscdebugprops}
...
   \textbf{Failsafe}: \texttt{false}\\
-  \textbf{Description}: Save macOS kernel panic to OpenCore root partition.
+  \textbf{Description}: Save macOS kernel panic output to the OpenCore root partition.
 
   The file is saved as \texttt{panic-YYYY-MM-DD-HHMMSS.txt}. It is strongly
-  recommended to have \texttt{keepsyms=1} boot argument to see debug symbols
-  in the panic log. In case it was not present \texttt{kpdescribe.sh} utility
-  (bundled with OpenCore) may be used to partially recover the stacktrace.
+  recommended to set the \texttt{keepsyms=1} boot argument to see debug symbols
+  in the panic log. In case where it is not present, the \texttt{kpdescribe.sh}
```
PMheart, line 3330: `In cases where...` maybe?

Docs/Configuration.tex
```diff
@@ -3290,33 +3315,33 @@ \subsection{Debug Properties}\label{miscdebugprops}
...
 
-  In case OpenCore kernel panic saving mechanism was not used, kernel panics may
-  still be found in \\
-  \texttt{/Library/Logs/DiagnosticReports} directory.
-  Starting with macOS Catalina kernel panics are stored in JSON format, so they
-  need to be preprocessed before passing to \texttt{kpdescribe.sh}:
+  In cases where the OpenCore kernel panic saving mechanism is not used, kernel panic
+  logs may still be found in the \texttt{/Library/Logs/DiagnosticReports} directory.
+
+  Starting with macOS Catalina, kernel panics are stored in JSON format and
```
PMheart, line 3343: `and thus` maybe?

Docs/Configuration.tex
```diff
@@ -3290,33 +3315,33 @@ \subsection{Debug Properties}\label{miscdebugprops}
...
+  need to be preprocessed before being passed to \texttt{kpdescribe.sh}:
```
PMheart, line 3344: I think `passing` is fine here. e.g. passing an argument to a script.

Docs/Configuration.tex
```diff
@@ -3412,46 +3439,49 @@ \subsection{Debug Properties}\label{miscdebugprops}
...
 
-  Console logging prints less than all the other variants.
+  Console logging prints less than the other variants.
   Depending on the build type (\texttt{RELEASE}, \texttt{DEBUG}, or
   \texttt{NOOPT}) different amount of logging may be read (from least to most).
 
-  Data Hub log will not log kernel and kext patches. To obtain Data Hub log use
-  the following command in macOS:
+  To obtain Data Hub logs, use the following command in macOS
+  (Data Hub logs do not log kernel and kext patches):
```
PMheart, line 3447: How about `Note that...`

Docs/Configuration.tex
```diff
@@ -3532,15 +3562,15 @@ \subsection{Security Properties}\label{miscsecurityprops}
   \href{https://github.com/acidanthera/bugtracker/issues/995}{acidanthera/bugtracker\#995}
   for more details.
 
-  \emph{Note 2}: Resetting NVRAM will also erase all the boot options
-  otherwise not backed up with bless (e.g. Linux).
+  \emph{Note 2}: Resetting NVRAM will also erase any boot options, such as
+  Linux boot options, not backed up using the bless command.
```
PMheart, line 3566: `which are not backed up` I guess?

Docs/Configuration.tex
```diff
@@ -3550,26 +3580,26 @@ \subsection{Security Properties}\label{miscsecurityprops}
...
   \href{https://support.apple.com/en-us/HT208330}{\texttt{Full Security}} of Apple
   Secure Boot.
 
-  To start using personalised Apple Secure Boot, the operating system will have
-  to be reinstalled or personalised. Unless the operating system is personalised,
-  macOS DMG recovery cannot be loaded. If DMG recovery is missing,
-  it can be downloaded with \texttt{macrecovery} utility and put to
-  \texttt{com.apple.recovery.boot} as explained in
+  To start using personalised Apple Secure Boot, the operating system must be
+  reinstalled or personalised. Unless the operating system is personalised,
```
PMheart, line 3594: Just `unless personalised`?

Docs/Configuration.tex
```diff
@@ -3578,16 +3608,15 @@ \subsection{Security Properties}\label{miscsecurityprops}
...
-  image an \texttt{Unable to verify macOS} error message will appear. To workaround
-  this issue allocate a dedicated RAM disk of 2 MBs for macOS personalisation
-  by entering the following commands in macOS recovery terminal before starting the
-  installation:
+  may not work as expected. When reinstalling the operating system, the macOS Installer
+  from macOS 10.15 and older will often run out of free memory on the \texttt{/var/tmp}
+  partition when trying to install macOS with the personalised Apple Secure Boot.
+  Soon after downloading the macOS installer image, an \texttt{Unable to verify macOS}
+  error message will appear. To workaround this issue, allocate a dedicated RAM disk of
+  2 MBs for macOS personalisation by entering the following commands in the macOS recovery
```
PMheart, line 3618: Mark here. Will go on later.

Docs/Configuration.tex
```diff
@@ -1404,10 +1404,10 @@ \subsection{Quirks Properties}\label{booterpropsquirks}
     \item KASLR (slide) is unsupported (this is macOS 10.7 or older)
   \end{itemize}
 
-  This quirk requires \texttt{ProvideCustomSlide} to also be enabled
-  and typically needs \texttt{AvoidRuntimeDefrag} to be set for this to
-  work correctly. Hibernation is not supported when booting with a relocation
-  block (but relocation block is not always used when the quirk is enabled).
+  This quirk requires \texttt{ProvideCustomSlide} to be enabled
+  and typically requires \texttt{AvoidRuntimeDefrag} to be set to work
```
vit9696, line 1408: `and typically requires \texttt{AvoidRuntimeDefrag} to be enabled as well for correct functioning.`

Docs/Configuration.tex
```diff
@@ -1404,10 +1404,10 @@ \subsection{Quirks Properties}\label{booterpropsquirks}
...
+  correctly. Hibernation is not supported when booting with a relocation
+  block (but the relocation block may not be used when the quirk is enabled).
```
vit9696, line 1410: `but the relocation block will only be used if required when the quirk is enabled`

Docs/Configuration.tex
```diff
@@ -1498,9 +1498,14 @@ \subsection{Quirks Properties}\label{booterpropsquirks}
...
   default, safe mode forces \texttt{0} slide as if the system was launched with the
-  \texttt{slide=0} boot argument. This quirk attempts to patch \texttt{boot.efi} to remove
-  this limitation and let some other value (from \texttt{1} to \texttt{255}) be used.
-  This quirk requires enabling \texttt{ProvideCustomSlide}.
+  \texttt{slide=0} boot argument.
+
+  \begin{itemize}
+    \tightlist
+    \item This quirk attempts to patch the \texttt{boot.efi} file to remove this limitation and
+      to allow using other values (\texttt{1} to \texttt{255} inclusive).
```
vit9696, line 1506: from missing?

Docs/Configuration.tex
```diff
@@ -1868,7 +1873,7 @@ \subsection{Properties}\label{kernelprops}
   drivers, which are not cached otherwise. The issue normally affects older
   operating systems, where various dependency kexts, such as \texttt{IOAudioFamily}
   or \texttt{IONetworkingFamily} may not be present in the kernel cache by default.
-  Kernel driver load order follows the item order in the array, thus the dependencies
+  The Kernel driver load order follows the item order in the array, thus the dependencies
```
vit9696, line 1876: Kernel → kernel

Docs/Configuration.tex
```diff
@@ -2626,31 +2632,34 @@ \subsection{Scheme Properties}\label{kernelpropsscheme}
...
-      corresponds to \texttt{-legacy} kernel boot argument. This option is unavailable
-      for 10.4 and 10.5 when running on 64-bit firmware due to an uninitialised 64-bit
-      segment in the XNU kernel, which causes AppleEFIRuntime to incorrectly execute
-      64-bit code as 16-bit code.
+      by the operating system.
+      \begin{itemize}
+        \tightlist
+        \item On macOS, 64-bit capable processors are assumed to support \texttt{SSSE3}.
+          This is not the case for older 64-bit capable Pentium processors, which cause some
+          applications to crash on macOS~10.6. This behaviour corresponds with the \texttt{-legacy}
```
vit9696, line 2653: it is not really with. Using `i386-user32` here does the same thing `-legacy` boot argument does. The change in the wording might make it less obvious.

Docs/Configuration.tex
vit9696: There is no such thing as OpenCore operating system as well. It was operating system loading behaviour.

Docs/Configuration.tex
vit9696, line 2824: by file header checks

Docs/Configuration.tex
```diff
@@ -2766,126 +2775,135 @@ \section{Misc}\label{misc}
...
-\item Alternate options follow corresponding primary options, i.e. Apple recovery will be following the
+\item Alternate options follow corresponding primary options. That is, Apple recovery options will follow the
   relevant macOS option whenever possible.
-\item Options will be listed in file system handle firmware order to maintain an established order across
-  the reboots regardless of the chosen operating system for loading.
+\item Options will be listed file system handle firmware order to maintain an established order across
+  reboots regardless of the operating system chosen for loading.
 \item Custom entries, tools, and system entries will be added after all other options.
-\item Auxiliary options will only show upon entering ``Advanced Mode'' in the picker (usually by pressing ``Space'').
+\item Auxiliary options will only be displayed upon entering ``Extended Mode'' in the boot picker (typically by pressing the \texttt{Space} key).
```
vit9696, line 2858: Hrm. I am not sure I like the Advanced → Extended mode change here. Advanced mode is common in BIOS prefs and other places. Better to keep it unless there is a good reason.

Docs/Configuration.tex
```diff
@@ -2766,126 +2775,135 @@ \section{Misc}\label{misc}
...
-custom BDS). Without \texttt{LauncherOption} it also is possible that other operating systems
-overwrite OpenCore, make sure to enable it when planning to use them.
+custom BDS). When \texttt{LauncherOption} is not enabled, other operating systems can
+overwrite OpenCore settings and this property should therefore be enabled when planning
+to use other operating systems.
 
-\emph{Note 2}: UEFI variable boot options' boot arguments will be removed if present as they
-may contain arguments compromising the operating system, which is undesired once secure boot
-is enabled.
+\emph{Note 2}: UEFI variable ``boot options'' boot arguments will be removed, if present, as
```
vit9696, line 2877: This was not a quote. UEFI boot option variables are structures, which contain boot options field that is used as boot arguments for most operating systems. Please revert..

Docs/Configuration.tex
```diff
@@ -2766,126 +2775,135 @@ \section{Misc}\label{misc}
...
+they may contain arguments that can compromise the operating system, which is undesirable when
+secure boot is enabled.
 
-\emph{Note 3}: Some operating systems, namely Windows, will create their boot option and
-mark it as top most upon first boot or after NVRAM Reset. When this happens default boot
-entry choice will update till next manual reconfiguration.
+\emph{Note 3}: Some operating systems, such as Windows, may create a boot option and mark it as
+the topmost option upon first boot or after NVRAM resets. When this happens, the default boot
```
vit9696, line 2882: NVRAM Reset refers to OpenCore functionality, not any kind of NVRAM reset.

Docs/Configuration.tex
vit9696: I like the former more.

Docs/Configuration.tex
vit9696: I do not understand what "unless overridden" (two d actually) refers to. You cannot override that in the plist. The only way to see them is by entering the Advanced mode.

Docs/Configuration.tex
vit9696: Agree

Docs/Configuration.tex
```diff
@@ -2998,66 +3018,68 @@ \subsection{Boot Properties}\label{miscbootprops}
     \item Entry is system (e.g. \texttt{Reset NVRAM}).
   \end{itemize}
 
-  To see all entries picker menu needs to be reloaded in extended mode by pressing
-  \texttt{Spacebar} key. Hiding auxiliary entries may increase boot performance
-  for multidisk systems.
-
+  To display all entries, the picker menu can be reloaded into ``Extended Mode'' by pressing the
```
vit9696, line 3021: Hmm, I see where it comes from… I agree that we should pick one, but maybe `Advanced` is really better.

Docs/Configuration.tex
```diff
@@ -2998,66 +3018,68 @@ \subsection{Boot Properties}\label{miscbootprops}
...
   \tightlist
   \item \texttt{Disabled} --- do nothing.
-  \item \texttt{Full} --- create or update top-priority
-  boot option in UEFI variable storage at bootloader startup. For this option
-  to work \texttt{RequestBootVarRouting} is required to be enabled.
+  \item \texttt{Full} --- create or update the top priority
+  boot option in UEFI variable storage at bootloader startup.
+  \begin{itemize}
+  \tightlist
+    \item Enable \texttt{RequestBootVarRouting} for this option to work.
```
vit9696, line 3039: Should not we keep passive here for consistency?

Docs/Configuration.tex
vit9696: No, this is correct, the NVRAM may work but there could be no boot option support.

Docs/Configuration.tex
```diff
@@ -3068,7 +3090,7 @@ \subsection{Boot Properties}\label{miscbootprops}
   \item \texttt{0x0001} --- \texttt{OC\_ATTR\_USE\_VOLUME\_ICON}, provides custom icons
     for boot entries:
 
-    For \texttt{Tools} OpenCore will try to load a custom icon and fallback to the default icon:
+    For \texttt{Tools}, OpenCore will attempt loading custom icons and fallback to a default icon on failure:
```
vit9696, line 3093: Is it still clear that each tool can have one icon at most?

Docs/Configuration.tex
```diff
@@ -3186,98 +3208,101 @@ \subsection{Boot Properties}\label{miscbootprops}
...
 
 \item
   \texttt{PickerMode}\\
   \textbf{Type}: \texttt{plist\ string}\\
   \textbf{Failsafe}: \texttt{Builtin}\\
   \textbf{Description}: Choose boot picker used for boot management.
 
-  Picker describes underlying boot management with an optional user interface
-  responsible for handling boot options. The following values are supported:
+  PickerMode describes the underlying boot management with an optional user
```
vit9696, line 3232: Use either Picker mode or `\texttt{PickerMode}`.

Docs/Configuration.tex
```diff
@@ -3186,98 +3208,101 @@ \subsection{Boot Properties}\label{miscbootprops}
...
   \item \texttt{Apple} --- Apple boot management is used if available.
-  Otherwise \texttt{Builtin} mode is used.
+  Otherwise, the \texttt{Builtin} mode is used.
   \end{itemize}
 
-  Upon success \texttt{External} mode will entirely disable all boot management
-  in OpenCore except policy enforcement. In \texttt{Apple} mode it may additionally
-  bypass policy enforcement. See \hyperref[ueficanopy]{OpenCanopy} plugin
-  for an example of a custom user interface.
+  Upon success, the \texttt{External} mode will entirely disable all boot management
```
vit9696, line 3247: will → may. OpenCanopy was changed to actually reuse most of that.

Docs/Configuration.tex
```diff
@@ -3186,98 +3208,101 @@ \subsection{Boot Properties}\label{miscbootprops}
...
   \begin{itemize}
   \tightlist
-  \item \texttt{Default} --- this is the default option, and it lets OpenCore built-in
-  boot picker to loads the default boot option as specified in
+  \item \texttt{Default} --- this is the default option, and it lets the built-in OpenCore
+  boot picker to load the default boot option as specified in the
   \href{https://support.apple.com/HT202796}{Startup Disk} preference pane.
-  \item \texttt{ShowPicker} --- this option forces picker to show. Normally it can be
-  achieved by holding \texttt{OPT} key during boot. Setting \texttt{ShowPicker} to
+  \item \texttt{ShowPicker} --- this option forces the boot picker to be displayed. This can
```
vit9696, line 3263: We try to use `picker` when we talk about OpenCore boot menu and `boot picker` when we talk about Apple's one. I'd opt to be consistent with this convention.

Docs/Configuration.tex
```diff
@@ -3186,98 +3208,101 @@ \subsection{Boot Properties}\label{miscbootprops}
...
+  \item \texttt{BootAppleRecovery} --- this option performs booting into the Apple operating
+  system recovery partition. This is either that related to the default chosen operating system,
+  or first one found when the chosen default operating system is not from Apple or does not have
+  a recovery partition. Hold the \texttt{CMD+R} key combination down to choose this option.
   \end{itemize}
 
-  \emph{Note 1}: Activated \texttt{KeySupport}, \texttt{OpenUsbKbDxe}, or similar driver is required
-  for key handling to work. On several types of firmware, it is not possible to get all the key functions.
+  \emph{Note 1}: The \texttt{KeySupport}, \texttt{OpenUsbKbDxe}, or similar drivers are required for key
+  handling. However, not all of the key handling functions are available on several types of firmware.
```
vit9696, line 3280: These functions may not be implemented in the firmware. E.g. `KeySupport` builds this on top firmware drivers. The original description is more accurate. `can be implemented on several types of firmware` perhaps.

Docs/Configuration.tex
```diff
@@ -3186,98 +3208,101 @@ \subsection{Boot Properties}\label{miscbootprops}
...
   \begin{itemize}
   \tightlist
-  \item \texttt{Auto} --- Automatically select one set of icons based on \texttt{DefaultBackground}
+  \item \texttt{Auto} --- Automatically select one set of icons based on the \texttt{DefaultBackground}
   colour.
   \item \texttt{Default} --- Normal icon set (without prefix).
   \item \texttt{Old} --- Vintage icon set (\texttt{Old} filename prefix).
   \item \texttt{Modern} --- Nouveau icon set (\texttt{Modern} filename prefix).
-  \item Other value --- Custom icon set if supported by the resources.
+  \item Other value --- Custom icon set if supported by available resources.
```
vit9696, line 3305: installed?

Docs/Configuration.tex
```diff
@@ -3412,46 +3439,49 @@ \subsection{Debug Properties}\label{miscdebugprops}
     \item \texttt{0x40} (bit \texttt{6}) --- Enable logging to file.
```
vit9696, line 3439: Should non-volatile be also be guarded with `\texttt{}` here? Given the text below.

Docs/Configuration.tex
```diff
@@ -3412,46 +3439,49 @@ \subsection{Debug Properties}\label{miscdebugprops}
...
-  even with boot log disabled.
+  While the OpenCore boot log already contains basic version information including build type
+  and date, this information may also be found in the \texttt{opencore-version} NVRAM variable
+  even when boot logging is disabled.
 
-  File logging will create a file named \texttt{opencore-YYYY-MM-DD-HHMMSS.txt} at EFI
-  volume root with log contents (the upper case letter sequence is replaced with date
+  File logging will create a file named \texttt{opencore-YYYY-MM-DD-HHMMSS.txt} under the
+  EFI volume root with log contents (the upper case letter sequence is replaced with date
   and time from the firmware). Please be warned that some file system drivers present
```
vit9696, line 3473: add in UTC

Docs/Configuration.tex
vit9696: This is wrong and gets even more wrong. Linux options generally follow bless model and thus do not die. Yet some users may create custom boot options, usually for Linux. These do die. In my opinion, revert.

Docs/Configuration.tex
vit9696: In my opinion, no.

Docs/Configuration.tex
```diff
@@ -3856,10 +3886,12 @@ \subsection{Security Properties}\label{miscsecurityprops}
     scanning devices directly connected to PCI bus (e.g. VIRTIO).
   \end{itemize}
 
-  \emph{Note}: Given the above description, \texttt{0xF0103} value is expected to allow
-  scanning of SATA, SAS, SCSI, and NVMe devices with APFS file system, and prevent scanning
+  \emph{Note}: Given the above description, a value of \texttt{0xF0103} can be expected to allow
```
vit9696, line 3889: There is no probability here. It should do this if working correctly. In my opinion, is expected is correct.

Docs/Configuration.tex
```diff
@@ -3915,39 +3949,41 @@ \subsection{Security Properties}\label{miscsecurityprops}
...
+    \item The list of cached drivers may be different, resulting in a need
       to change the list of \texttt{Added} or \texttt{Forced} kernel drivers.
       For example, \texttt{IO80211Family} cannot be injected in this case.
     \item System volume alterations on operating systems with sealing, such as
       macOS~11, may result in the operating system being unbootable. Do not
       try to disable system volume encryption unless Apple Secure Boot is disabled.
-    \item If the platform requires certain settings, but they were not enabled,
-      because the obvious issues did not trigger before, boot failure might occur.
+    \item Boot failures might occur when the platform requires certain settings,
+      but they have not been enabled because the associated issues were not moticed before.
```
vit9696, line 3968: noticed? maybe discovered earlier?

Docs/Configuration.tex
```diff
@@ -3915,39 +3949,41 @@ \subsection{Security Properties}\label{miscsecurityprops}
...
-      and are assumed to be handled by the firmware as Microsoft Windows is.
-    \item On older CPUs (e.g. before Sandy Bridge) enabling Apple Secure Boot
-      might cause slightly slower loading by up to 1 second.
-    \item Since \texttt{Default} value will increase with time to support the latest
-      major release operating system, it is not recommended to use \texttt{ApECID}
-      and \texttt{Default} value together.
+    \item Operating systems released before Apple Secure Boot was released (e.g.
+      macOS~10.12 or earlier), will still boot until UEFI Secure Boot is enabled.
+      This is so because Apple Secure Boot treats these as incompatible
+      and are assumed to be handled by the firmware (as Microsoft Windows is).
```
vit9696, line 3973: and they are then handled...

Docs/Configuration.tex
```diff
@@ -3632,34 +3661,34 @@ \subsection{Security Properties}\label{miscsecurityprops}
...
+    there are \texttt{boot.efi} files compatible with Apple Secure Boot.
     Manually downloaded DMG images stored in \texttt{com.apple.recovery.boot}
     directories will not load, however.
-  \item \texttt{Signed} --- only Apple-signed DMG images will load. Due to
-    Apple Secure Boot design \texttt{Signed} policy will let any Apple-signed
-    macOS Recovery to load regardless of Apple Secure Boot state, which may
+  \item \texttt{Signed} --- only Apple-signed DMG images will load. Due to the
+    design of Apple Secure Boot, the \texttt{Signed} policy will let any Apple-signed
+    macOS Recovery load regardless of the Apple Secure Boot state, which may
     not always be desired.
```
vit9696, line 3672: Add something like: While using signed DMG images is more desired, verifying image signature may slightly slowdown the boot time (by up to 1 second).

Docs/Configuration.tex
```diff
@@ -3632,34 +3661,34 @@ \subsection{Security Properties}\label{miscsecurityprops}
...
-    usually are \texttt{boot.efi} files compatible with Apple Secure Boot.
+  \item \texttt{Disabled} --- loading DMG images will fail. The \texttt{Disabled}
+    policy will still let the macOS Recovery load in most cases as typically,
+    there are \texttt{boot.efi} files compatible with Apple Secure Boot.
     Manually downloaded DMG images stored in \texttt{com.apple.recovery.boot}
     directories will not load, however.
-  \item \texttt{Signed} --- only Apple-signed DMG images will load. Due to
-    Apple Secure Boot design \texttt{Signed} policy will let any Apple-signed
-    macOS Recovery to load regardless of Apple Secure Boot state, which may
+  \item \texttt{Signed} --- only Apple-signed DMG images will load. Due to the
```
vit9696, line 3669: I saw dmg instead of DMG in several places, could you fix?

Docs/Configuration.tex
```diff
@@ -3915,39 +3949,41 @@ \subsection{Security Properties}\label{miscsecurityprops}
...
+      target volumes. This may include HFS+ formatted drives when no spare APFS drive is available.
   \end{enumerate}
 
-  Sometimes the already installed operating system may have outdated Apple Secure
-  Boot manifests on the \texttt{Preboot} partition causing boot failure. If there is
-  ``OCB: Apple Secure Boot prohibits this boot entry, enforcing!'' message,
-  it is likely the case. When this happens, either reinstall the operating
+  The installed operating system may have sometimes outdated Apple Secure Boot manifests
+  on the \texttt{Preboot} partition, resulting in boot failures. This is likely to be the
+  case when a ``OCB: Apple Secure Boot prohibits this boot entry, enforcing!'' message
```
vit9696, line 3985: when an?

Docs/Configuration.tex
```diff
@@ -3973,28 +4009,28 @@ \subsection{Entry Properties}\label{miscentryprops}
   \texttt{Auxiliary}\\
   \textbf{Type}: \texttt{plist\ boolean}\\
   \textbf{Failsafe}: \texttt{false}\\
-  \textbf{Description}: This entry will not be listed by default when
-  \texttt{HideAuxiliary} is set to \texttt{true}.
+  \textbf{Description}: Set to \texttt{true} to hide this entry
+  when \texttt{HideAuxiliary} is also set to \texttt{true}.
+  Press the \texttt{Spacebar} key to display the entry when hidden.
```
vit9696, line 4014: to enter `Advanced mode` and display…

Docs/Configuration.tex
```diff
@@ -3973,28 +4009,28 @@ \subsection{Entry Properties}\label{miscentryprops}
...
+  \textbf{Description}: Arbitrary ASCII string used to provide a human readable
   reference for the entry. Whether this value is used is implementation defined.
 
 \item
   \texttt{Enabled}\\
   \textbf{Type}: \texttt{plist\ boolean}\\
   \textbf{Failsafe}: \texttt{false}\\
-  \textbf{Description}: This entry will not be listed unless set to
-  \texttt{true}.
+  \textbf{Description}: Set to \texttt{true} deactivate this entry.
```
vit9696, line 4027: This is borked. Is it rather `to activate`?

Docs/Configuration.tex
```diff
@@ -3973,28 +4009,28 @@ \subsection{Entry Properties}\label{miscentryprops}
...
 
 \item
   \texttt{Name}\\
   \textbf{Type}: \texttt{plist\ string}\\
   \textbf{Failsafe}: Empty\\
-  \textbf{Description}: Human readable entry name displayed in boot picker.
+  \textbf{Description}: Human readable entry name displayed in the boot picker.
```
vit9696, line 4033: A human readable … in the picker?

Docs/Configuration.tex
```diff
@@ -4018,25 +4054,25 @@ \subsection{Entry Properties}\label{miscentryprops}
...
   \textbf{Type}: \texttt{plist\ boolean}\\
   \textbf{Failsafe}: \texttt{false}\\
   \textbf{Description}: Run the entry in text mode instead of graphics mode.
 
-  This setting may be benefitial to some older tools that require text output.
-  By default all the tools are launched in graphics mode. Read more about text
-  modes in the \hyperref[uefioutputprops]{Output Properties} section below.
+  This setting may be beneficial for some older tools that require text output
+  as all the tools are launched in graphics mode by default. Refer to the
+ \hyperref[uefioutputprops]{Output Properties} section below for information on text modes.
```
vit9696, line 4075: was a leading space dropped here?

Docs/Configuration.tex
```diff
@@ -6103,8 +6138,8 @@ \subsection{Audio Properties}\label{uefiaudioprops}
   \textbf{Description}: Audio codec reconfiguration delay in microseconds.
 
   Some codecs require a vendor-specific delay after the reconfiguration
-  (e.g. volume setting). This option makes it configurable. In general
-  the necessary delay may be as long as 0.5 seconds.
+  (e.g. volume setting). This option makes it configurable. The required
+  delay may be as long as 0.5 seconds.
```
vit9696, line 6142: It changes the meaning here. We observed the issue on systems that require 0.5 seconds. But it does not mean all the systems will have this delay. This value is rather a common-sense recommendation.

Docs/Configuration.tex
```diff
@@ -6293,20 +6328,20 @@ \subsection{Output Properties}\label{uefioutputprops}
...
   all platforms. \texttt{SanitiseClearScreen}, \texttt{IgnoreTextInGraphics},
   and \texttt{ClearScreenOnModeSwitch} are more specific, and their use
   depends on the firmware.
 
-  \emph{Note}: Some Macs, namely \texttt{MacPro5,1}, may have broken
-  console output with newer GPUs, and thus only \texttt{BuiltinGraphics}
-  may work for them.
+  \emph{Note}: Some Macs, such as the \texttt{MacPro5,1}, may have incompatible
+  console output when using modern GPUs, and thus only \texttt{BuiltinGraphics}
+  may work for them in such cases.
```
vit9696, line 6344: It is worth adding that NVIDIA GPUs need additional firmware upgrades and linking  to https://github.com/acidanthera/bugtracker/issues/1280.

Docs/Configuration.tex
dakanji: These are instructions/specifications (telling users what to do). An active voice is needed.

Docs/Configuration.tex
dakanji: Isn't "finished" just a synonym here?
Doesn't add any nuance missing from "completed".

Docs/Configuration.tex
dakanji: Perhaps even better not to mention Samsung in the first place

Docs/Configuration.tex
dakanji: > Why repeatedly say configuration options here? 

"Repeatedly", which suggests multiple times, is misleading here as the phrase is only repeated once.
Overall aim has been to be as explicit as possible so that the reader never has any doubt what is being talked about at any point. A third and fourth repetition would clearly be too much but a single one helps make things clear. 

If you feel strongly about it, can change to just "options".

dakanji: > There is no such thing as OpenCore operating system as well. 

Note that was always there. Seemed odd but just left alone. Will amend.

Docs/Configuration.tex
dakanji: My understanding is:

Other OS may attempt to overwrite OC Settings.
When LauncherOption is disabled, they can actually do this but when it is enabled, they cannot do this.

Docs/Configuration.tex
dakanji: > change Hides to Hide

The description refers to what the subject, HideAuxiliary, does. It hides xyz.

dakanji: >  do not understand what "unless overridden" (two d actually) refers to

Removed. Aim is met in the notes

Docs/Configuration.tex
dakanji: As you wish!

Docs/Configuration.tex
dakanji: Correct that it could be like that but "being passed to" makes it explicitly clear we are talking about passing the preprocessed item to the script.

That is, with just "passing", you could do the preprocessing, Task 1, before passing (on to running) the script, a separate and not necessarily directly related Task 2.

Docs/Configuration.tex
dakanji: Changed to the same intent

Docs/Configuration.tex
dakanji: This looks like this: `... to allow using other values (1 to 255 inclusive).`
Does not need "from" ... but added anyway.

Docs/Configuration.tex
dakanji: Maintained "Extended"

Docs/Configuration.tex
dakanji: Changed to `after NVRAM resets from within OpenCore`

Docs/Configuration.tex
dakanji: Maintained "Extended"

Docs/Configuration.tex
dakanji: Aim is to go active across the board:
- Set to true to do XYZ
- Enable this for that to happen
- You should do this
- You must not do that
etc

Docs/Configuration.tex
dakanji: Amended to singular

Docs/Configuration.tex
dakanji: Used `\texttt{PickerMode}`

Docs/Configuration.tex
dakanji: Changed. Use of 'may' might be reserved later.

Docs/Configuration.tex
dakanji: Used `OpenCore picker` and `Apple picker` respectively

Docs/Configuration.tex
dakanji: Changed but note that "is expected" implies there is a probability, maybe small. but still present, that it may not happen.

If indeed there is no such probability, then should be `a value of \texttt{0xF0103} will allow`

Docs/Configuration.tex
dakanji: Seems there was only one or two relevant ones

Docs/Configuration.tex
dakanji: OpenCore picker

Docs/Configuration.tex
vit9696: No, this is important, as basically all Samsung SSDs are unusable (from the APFS point of view) for macOS despite being widely common.

Docs/Configuration.tex
vit9696: don't see the difference either, could leave as is, unless @PMheart has a point

Docs/Configuration.tex
vit9696: > Note that was always there. Seemed odd but just left alone. Will amend.

Sure, but it changed the meaning. Previously it was `[OpenCore operating system] loading behaviour` now it is `the loading behaviour of the OpenCore operating system`. I think you did not catch the fact it is rather "OpenCore's loading behaviour of the operating systems it discovers".

Docs/Configuration.tex
vit9696: Not really, they still can if they try hard. It rather means that there a higher plausibility that they do if it is off.

Docs/Configuration.tex
vit9696: Hrm. I do not believe I saw stuff like "Enable X for Y" in the specs.

Docs/Configuration.tex
dakanji: `OpenCore operating system loading behaviour` is invalid gramatically which caused the confusion.
How about  `how OpenCore behaves while loading operating systems` or something to that effect?

Docs/Configuration.tex
dakanji: which do you want? "from" or no "from"?

Docs/Configuration.tex
dakanji: So what would you prefer?

Docs/Configuration.tex
dakanji: Removed the "from" to close out.

Docs/Configuration.tex
dakanji: Restored original phrase to close out.

Docs/Configuration.tex
dakanji: Reverted to "passing" to close out

Docs/Configuration.tex
dakanji: Found "For this option to work, \texttt{RequestBootVarRouting} is required to be enabled."

Docs/Configuration.tex
dakanji: Changed to "finished" to close out

PR #217, opened 2021-02-23
Down keys - WIP

ghost:
So, I finished the rough proof of concept of ‘new down key’ detection.

Basic method:
1. ~Keep up/down state of all Apple hid keys with an original USB code of below 256 (which should be all of them?) in a 256 CHAR8 buffer (values 0 and 1 only)~ **EDIT:** Major improvements to state buffer approach after input from @vit9696 and @mhauser.
2. Call new method OcGetDownKeys which returns newly-down keys only, in main kb loop, instead of calling KeyMap->GetKeyStrokes directly.

To a first approximation, OcKeyMapFlush is no longer needed and is commented out to do nothing in this PoC (it is not obvious if/whether the specific additional flush code which was in there for some legacy machines, even *can* be applied anywhere truly sensible in the new logic).

It works.

The UI feel is honestly considerably nicer; if I could put it this way, you don’t really notice that much when switching to it (well, except for ESC not hanging the UI, which is a major plus anyway), but unfortunately you definitely do notice, for the worse, when switching back again.

With this method on:

On my Apple Mac (which I suspect may be quite typical?), and on OVMF + OpenUsbKbDxe.efi: keys respond on key down now; there is no hanging (e.g. waiting for control key to be released before actioning CTRL+Enter); ESC does not hang the boot menu; ~however, there is no kb repeat (e.g. for left right arrow), but noting that, for obvious reasons (i.e. nothing at all happened until you released a key), there never was on these setups~. **EDIT:** Key repeat handling (and optional non-repeating keys where required) is now an essential part of the updated PR.

In OVMF + KeySupport: keys still respond to key down as before; there is (still) kb repeat.

(Additional advantage in my mind, if you could accept this method, would be that it would now be possible to implement circular-arrow-while-ctrl-key-depressed in OpenCanopy, as in Apple bootpicker, since the anim is now continuing while keys, incl CTRL, are held down; and wasn’t before.)

### DIFF
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -6216,6 +6216,24 @@ functioning. Feature highlights:
 
 \begin{enumerate}
 
+  \item
+  \texttt{DownkeysHandler}\\
+  \textbf{Type}: \texttt{plist\ string}\\
+  \textbf{Failsafe}: \texttt{Auto}\\
+  \textbf{Description}: Configure additional keyboard handling to enable more fluent re-rendering within
+  bootpicker. When applied, affects both \texttt{Builtin} bootpicker and OpenCanopy. For best UI experience
+  this setting should always be enabled when \texttt{KeySupport} is disabled,
+  and should normally be disabled when \texttt{KeySupport} is enabled. On
+  some less common configurations (particularly some PS/2 keyboards) enabling this setting together with
+  \texttt{KeySupport} is required to obtain correct bootpicker key repeat behaviour.
+
+  \begin{itemize}
+  \tightlist
+  \item \texttt{Auto} --- Performs automatic choice based on \texttt{KeySupport} setting.
+  \item \texttt{Enabled} --- Add additional downkey handling.
+  \item \texttt{Disabled} --- Do not add additional downkey handling.
+  \end{itemize}
+
 \item
   \texttt{KeyFiltering}\\
   \textbf{Type}: \texttt{plist\ boolean}\\
--- a/Include/Acidanthera/Library/OcAppleKeyMapLib.h
+++ b/Include/Acidanthera/Library/OcAppleKeyMapLib.h
@@ -1,5 +1,6 @@
 /** @file
   Copyright (C) 2019, Download-Fritz. All rights reserved.
+  Additions (downkeys support) copyright (C) 2021 Mike Beaton. All rights reserved.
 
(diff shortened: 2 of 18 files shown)


Bmju:
https://github.com/acidanthera/bugtracker/issues/1526

Bmju:
Revised WIP ideas based on feedback from @mhaeuser - still very much PoC Do Not Merge - thanks

Bmju:
@vit9696 - Just to flag you that this is ready for your own testing, nothing currently urgent, when you have time, with Auto/Enabled/Disabled flag as requested.

I still plan to do more work, before I consider it nearly mergeable:

1. Passing flags out to the two pickers (not least, that they should be flags rather than two BOOLEANs as I have it now)
2. I would plan to incorporate changing to '+' instead of '>' for CTRL-held indicator in Builtin picker as part of validating 1.
3. Allowing SPACE to toggle (as opposed to set) the '_MORE' options is yet another PoC, and is not planned for live, I was just testing it, not least as there are no voice prompts to support it - as noted in comments where this PoC code is
4. Causing non-repeat on certain keys is needed/helpful regardless of 3. It should be applied to any keys which refresh the entire list of available options. I have currently applied this to SPACE and ESC behaviour, and have not yet looked carefully as to whether there are any other equivalents.

MikeBeaton:
@vit9696 - Just to flag you that this is ready for your own testing, nothing currently urgent, when you have time, with Auto/Enabled/Disabled config setting as requested.

I still need to do more work before nearly mergeable:

- [x] Improve code passing flags out to the two pickers (not least, they should be flags rather than two BOOLEANs as now)
- [ ] Incorporate changing to '+' instead of '>' as CTRL-held indicator in Builtin picker as part of validating above
   - In progress
- [x] Allowing SPACE to toggle (as opposed to set) the '_MORE' options is currently PoC; rethink carefully. But I think it is okay just in Canopy (are voice prompts Bultin only?)
   - Left in as PoC in test version, but comment-out-able on `#define`
   - Will need feedback
- [x] Non-repeat on certain keys is needed/helpful (as mentioned by @mhaeuser). Currently applied to SPACE and ESC (and directly equivalent behaviour keys); need to confirm whether others required.
   - Leave on just these on test version
   - Will need feedback

MikeBeaton:
Still working - re-doing via AppleEvents, as discussed - brb...!

PR #222, opened 2021-02-28
OCB: make builtin picker start at default OS even when no timeout

Bmju:
Minor change, implements preferred behaviours indicated at https://github.com/acidanthera/bugtracker/issues/1536 .


vit9696:
Hmm, will it keep showing `*` when timeout is enabled and `>` when it is not?

Bmju:
Yes

PR #223, opened 2021-02-28
OcAfterBootCompatLib: Store BootCompat CpuInfo

cecekpawon:
Hi,

Assumed here you forgot to store CpuInfo into BootCompat->CpuInfo to be used [here](https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcAfterBootCompatLib/CustomSlide.c#L812).


vit9696:
Thank you for that, it indeed is quite a terrible issue.

PR #225, opened 2021-03-01
OcSmbiosLib: Report actual MemoryFormFactor in SMBIOS

PMheart:
closes https://github.com/acidanthera/bugtracker/issues/1537


Docs/Configuration.tex
```diff
@@ -5100,6 +5100,13 @@ \subsection{Memory Properties}\label{platforminfomemory}
   When \texttt{CustomMemory} is \texttt{false}, this value is automatically set
   based on Mac product name.
 
+  When \texttt{Automatic} is \texttt{true}, the original value from the firmware
+  will be set if valid (i.e. \texttt{non-zero}). Otherwise, the value from \texttt{OcMacInfoLib}
```
vit9696, line 5104: from the corresponding Mac model*

Docs/Configuration.tex
```diff
@@ -5100,6 +5100,13 @@ \subsection{Memory Properties}\label{platforminfomemory}
...
+  will be set. Under non-Automatic mode (i.e. when \texttt{Automatic} is \texttt{false}),
+  a user-specified value will be set if available. Otherwise, the original value from 
+  the firmware will be set if valid (i.e. \texttt{non-zero}).
```
vit9696, line 5107: Whether or not the value is valid, it essentially is used, because invalid firmware value equals 0. Please drop the `if valid` part.

Library/OcSmbiosLib/SmbiosPatch.c
```diff
@@ -105,6 +105,60 @@ SmbiosGetOriginalStructureCount (
...
+
+  IsAutomatic = !Data->ForceMemoryFormFactor;
+
+  if (IsAutomatic) {
+    if (SmbiosHasValidOemFormFactor (Original)) {
+      //
+      // Try to use the original value if valid first.
+      //
+      return Original.Standard.Type17->FormFactor;
+    } else if (Data->MemoryFormFactor != NULL) {
```
vit9696, line 136: Please do not use else if after return. Always use just if. Here and below. This improves code readability.

PR #227, opened 2021-03-01
Various OpenCanopy performance and responsiveness improvements

mhaeuser:
Properly clip bounds from the main viewport down, propagating conditions to remove many unnecessary branches. Simplify the "draw request" and pointer redraw logic to provide an easier interface and more consistent drawing performance, forcing draws (excluding the cursor) to go through the entire stack. Separate blending approaches (solid, opaque) to decrease branches in the main blending code, as blending with additional opacity is not used frequently. FPS improvements are clearly mensurable in QEMU, albeit providing little noticeable improvement. Remove legacy pointer support and update stack to slim the stack. Poll latest pointer position right before drawing to reduce input lag.


Platform/OpenCanopy/BitmapFont.c
```diff
@@ -618,11 +618,13 @@ BlendMem (
   UINTN  Index;
 
   for (Index = 0; Index < PixelCount; ++Index) {
-    //
-    // We assume that the font is generated by dpFontBaker
-    // and has only gray channel, which should be interpreted as alpha.
-    //
-    GuiBlendPixel(Dst, Color, AlphaSrc->Red);
+    if (AlphaSrc->Red != 0) {
```
vit9696, line 621: Are you sure this is a speedup? Conditionals noticeable slowdown code in general.

vit9696:
Looks solid otherwise at a glance. Should merge as long as it works with both UIScale 1 and 2.

Platform/OpenCanopy/BitmapFont.c
mhaeuser: This is a tradeoff, this is code that is called once per label on initialisation, and it saves branches in the main draw loop in return.

vit9696: Isn't it per-pixel?

mhaeuser: @vit9696 Yes, the point is it slows down only initialisation (and only slightly, as labels are small), while the other paths, that are improved, are in the main drawing loop.

mhaeuser: @vit9696 Hmm, actually I realised the code works for 0 opacity (I thought there was a division by 0 with the new code), do you want me to drop the branch?

vit9696: I'd rather profile.

mhaeuser: I don't think I really have time for that, could leave it as-is and add a TODO

PR #232, opened 2021-03-07
Update SSDT-SBUS-MCHC.dsl

leungbu:


khronokernel:
What's the purpose of this PR? The SSDTs we provide in OpenCorePkg are merely an example, not for specific machines

PR #233, opened 2021-03-08
OcCpuLib: Implement TSC frequency reading from Apple Platform Info

PMheart:
closes https://github.com/acidanthera/bugtracker/issues/1484


Library/OcCpuLib/FrequencyDetect.c
```diff
@@ -225,6 +228,57 @@ InternalCalculateTSCFromPMTimer (
...
+    PlatformInfo,
+    &gAppleFsbFrequencyPlatformInfoGuid,
+    &Size
+    );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_INFO, "OCCPU: Failed to get first data size - %r\n", Status));
+    return 0;
+  }
+
+  Status = PlatformInfo->GetFirstData (
```
vit9696, line 261: You should check that `Size` is not bigger than `sizeof (UINT64)` before calling `GetFirstData` here. Otherwise it will overwrite wrong memory. For our values it is an error, so it is safe to just return 0 if that actually happens.

Library/OcCpuLib/FrequencyDetect.c
```diff
@@ -225,6 +228,57 @@ InternalCalculateTSCFromPMTimer (
...
+    PlatformInfo,
+    &gAppleFsbFrequencyPlatformInfoIndexHobGuid,
+    &FsbFreq,
```
vit9696, line 264: Before reading to `FsbFreq` please initialise it with 0. As if we have `UINT32` (4 byte) FsbFrequency, then higher bits will not be set to initialised values.

Library/OcCpuLib/OcCpuLib.c
```diff
@@ -273,33 +273,81 @@ ScanThreadCount (
...
+  UINT8                               CpuModel;
 
-  //
-  // TODO: this may not be accurate on some older processors.
-  //
-  if (CpuInfo->CpuGeneration >= OcCpuGenerationNehalem) {
-    PlatformInfo.Uint64 = AsmReadMsr64 (MSR_NEHALEM_PLATFORM_INFO);
-    MaxBusRatio = (UINT8) PlatformInfo.Bits.MaximumNonTurboRatio;
-    MaxBusRatioDiv = 0;
+  if (MaxBusRatio == NULL || MaxBusRatioDiv == NULL) {
```
vit9696, line 287: It is a requirement for this function to get pointers to `MaxBusRatio` and `MaxBusRatioDiv` (they are OUT, and there is no OPTIONAL). Thus these checks should become `ASSERT`s, because it is wrong to call the function with these pointers `NULL`.

Library/OcCpuLib/OcCpuLib.c
```diff
@@ -273,33 +273,81 @@ ScanThreadCount (
...
+    return;
+  }
+
+  if (CpuInfo != NULL) {
```
vit9696, line 291: Rather introduce a `BOOLEAN` — `NehalemOrNewer`, and assign it to `CpuInfo->CpuGeneration >= OcCpuGenerationNehalem` in the if branch and to `CpuModel >= CPU_MODEL_NEHALEM` in the else branch. Do not copy the whole assignment logic.

Library/OcCpuLib/OcCpuLib.c
```diff
@@ -273,33 +273,81 @@ ScanThreadCount (
...
+    if (CpuModel >= CPU_MODEL_NEHALEM) {
+      PlatformInfo.Uint64 = AsmReadMsr64 (MSR_NEHALEM_PLATFORM_INFO);
+      *MaxBusRatio = (UINT8) PlatformInfo.Bits.MaximumNonTurboRatio;
+      *MaxBusRatioDiv = 0;
+    } else {
+      PerfStatus.Uint64 = AsmReadMsr64 (MSR_IA32_PERF_STATUS);
+      *MaxBusRatio = (UINT8) (RShiftU64 (PerfStatus.Uint64, 8) & 0x1FU);
+      *MaxBusRatioDiv = (UINT8) (RShiftU64 (PerfStatus.Uint64, 46) & BIT0);
+    }
+  }
```
vit9696, line 333: Add a check, if `*MaxBusRatio` is 0, assign it 1, and then drop the `MaxBusRatio > 0` check. Sounds like a safe failsafe in my opinion.

Library/OcCpuLib/OcCpuLib.c
```diff
@@ -416,10 +483,15 @@ ScanIntelProcessor (
       Recalculate
       );
 
+    //
+    // Determine the TSC frequency from Apple Platform Info.
+    //
+    Cpu->CPUFrequencyFromApple = InternalCalculateTSCFromApplePlatformInfo (NULL);
```
vit9696, line 489: Please do it only if `Cpu->CPUFrequencyFromART == 0 || Recalculate` is TRUE, in my opinion.

Library/OcCpuLib/FrequencyDetect.c
```diff
@@ -225,6 +228,57 @@ InternalCalculateTSCFromPMTimer (
...
+    return 0;
+  }
+
+  Status = PlatformInfo->GetFirstDataSize (
+    PlatformInfo,
+    &gAppleFsbFrequencyPlatformInfoGuid,
+    &Size
+    );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_INFO, "OCCPU: Failed to get first data size - %r\n", Status));
```
vit9696, line 257: Actually we should use a fallback mechanism when reading via automatic FirstData discovery fails. It is [here](https://github.com/acidanthera/mac-efi-firmware/blob/759669b625c7d4b6d1d09afd06dd20a0bce8b353/EfiPkg/Universal/AppleSmbiosDxe/AppleSmbios.c#L614) in the ref.

I.e. get the HOB and then use it manually. GetFirstGuidHob and GET_GUID_HOB_DATA are implemented by [HobLib](https://github.com/acidanthera/audk/blob/master/MdePkg/Include/Library/HobLib.h).

Library/OcCpuLib/OcCpuLib.c
```diff
@@ -424,16 +487,31 @@ ScanIntelProcessor (
       TimerAddr = InternalGetPmTimerAddr (&TimerSourceType);
       DEBUG ((DEBUG_INFO, "OCCPU: Timer address is %Lx from %a\n", (UINT64) TimerAddr, TimerSourceType));
       DEBUG_CODE_END ();
-      Cpu->CPUFrequencyFromTSC = InternalCalculateTSCFromPMTimer (Recalculate);
+      Cpu->CPUFrequencyFromApple = InternalCalculateTSCFromApplePlatformInfo (NULL);
+      Cpu->CPUFrequencyFromTSC   = InternalCalculateTSCFromPMTimer (Recalculate);
```
vit9696, line 491: THis is broken in another way now. We do not want to call `InternalCalculateTSCFromPMTimer` if Apple succeeds.

Library/OcCpuLib/FrequencyDetect.c
```diff
@@ -225,6 +231,94 @@ InternalCalculateTSCFromPMTimer (
...
+      return 0;
+    }
+
+    Status = PlatformInfo->GetFirstData (
+      PlatformInfo,
+      &gAppleFsbFrequencyPlatformInfoIndexHobGuid,
+      &FsbFreq,
+      &Size
+      );
+    if (EFI_ERROR (Status)) {
```
vit9696, line 274: On failure this should fall back to GetFirstGuidHob (NULL) path.

Library/OcCpuLib/FrequencyDetect.c
```diff
@@ -225,6 +231,94 @@ InternalCalculateTSCFromPMTimer (
...
+    return 0;
+  }
+
+  Status = PlatformInfo->GetFirstDataSize (
+    PlatformInfo,
+    &gAppleFsbFrequencyPlatformInfoGuid,
+    &Size
+    );
+  if (!EFI_ERROR (Status)) {
+    if (Size > sizeof (UINT64) || Size < sizeof (UINT32)) {
```
vit9696, line 263: On failure this should fallback to GetFirstGuidHob (NULL) path. Perhaps you need to introduce two functions that would just read values from different places?

```C
EFI_STATUS
ReadApplePlatformFirstData (
  IN APPLE_PLATFORM_INFO  *PlatformInfo,
  IN OUT UINT32 *Size,
  OUT VOID *Data
  );
```

And another one for ReadApplePlatformData.

Library/OcCpuLib/FrequencyDetect.c
```diff
@@ -225,6 +231,94 @@ InternalCalculateTSCFromPMTimer (
...
+  OUT  UINT64  *FSBFrequency  OPTIONAL
+  )
+{
+  EFI_STATUS                             Status;
+  APPLE_PLATFORM_INFO_DATABASE_PROTOCOL  *PlatformInfo;
+  UINT32                                 Size;
+  VOID                                   *FsbHob;
+  UINT64                                 FsbFreq;
+
+  FsbFreq = 0;
```
vit9696, line 245: 1. Introduce a guard checking no recursive entrance.
2. Add a Recalculate parameter.
3. Change DEBUG_VERBOSE back to DEBUG_INFO.

Something like that will do:

```
{
STATIC BOOLEAN  ObtainedFreqs = FALSE;
STATIC UINT64 FsbFreq = 0;
STATIC UINT64 TscFreq = 0;

if (Recalculate) {
  ObtainedFreqs = FALSE;
  FsbFreq = TscFreq = 0;
}

if (!ObtainedFreqs) {
  ObtainedFreqs = TRUE;
  /* Reading code */

  TscFreq = InternalConvertAppleFSBToTSCFrequency (FsbFreq);
}

if (FsbFrequency != NULL) {
  *FsbFrequency = FsbFreq;
}

return TscFreq;
}
```

PR #234, opened 2021-03-09
OcSmbiosLib: Fix index in OcSmbiosGetSmcVersion

kylon:
I don't know if it was intended behaviour, if so, sorry, ignore this pr.

it produces different result for revisions like {0x02, 0x12, 0x0f, 0x00, 0x01, 0x43}

master: 2.1271143

this pr: 2.1270143


vit9696:
Need to check real Macs. I do not remember where the algo went from. CC @khronokernel @dhinakg

PR #235, opened 2021-03-09
"macOS Installer" boot entry name fix

BbIKTOP:
"macOS Installer" boot entry name fix - as on the real mac!


vit9696:
This is not necessarily the installer in this code path, so we cannot apply this change.

PR #237, opened 2021-03-15
OcAcpiLib: Provide `Base` lookup for ACPI patches

vit9696:
Based on pilot project by @koralexa and @Ubsefor.


dbisping:
Unsubscribe

dbisping:
Unsubscribe 

O the joy of my spirit - it is uncaged - it darts like lightning!
It is not enough to have this globe or a certain time,
I will have thousands of globes and all time.

- Walt Whitman

PR #241, opened 2021-03-23
HotKeys: Improve allocation/deallocation logic and timing

MikeBeaton:
@mhaeuser in response to requested changes, could do with a re-review if possible - many thanks

### DIFF
--- a/Library/OcBootManagementLib/HotKeySupport.c
+++ b/Library/OcBootManagementLib/HotKeySupport.c
@@ -113,7 +113,7 @@ OcLoadPickerHotKeys (
 //
 // Initialise picker keyboard handling.
 //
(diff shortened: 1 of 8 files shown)


Library/OcBootManagementLib/HotKeySupport.c
```diff
@@ -153,12 +157,34 @@ OcInitHotKeys (
...
+
+//
+// Free picker keyboard handling resources.
+//
+VOID
+OcFreeHotKeys (
+  IN     OC_PICKER_CONTEXT  *Context
+  )
+{
+    if (Context == NULL) {
```
mhaeuser, line 180: Too large indentation

PR #245, opened 2021-03-29
Docs: Clarity on Kext Order

dakanji:


Docs/Configuration.tex
```diff
@@ -1882,11 +1882,10 @@ \subsection{Properties}\label{kernelprops}
...
   \emph{Note 2}: To track the dependency order, inspect the \texttt{OSBundleLibraries}
   key in the \texttt{Info.plist} file of the kext being added. Any kext included
-  in the \texttt{OSBundleLibraries} key is a dependency that must precede
-  the kext being added.
+  under the key is a dependency that must appear before the kext being added.
 
-  \emph{Note 3}: Kexts may have inner kexts (\texttt{Plugins}) included
-  in the bundle and each such plugin must be added separately.
+  \emph{Note 3}: Kexts may have inner kexts (\texttt{Plugins}) included in the bundle.
+  Each such plugin must be added separately and must appear after the host kext.
```
vit9696, line 1888: This is not correct. Some kexts put plugins as their dependencies. Do not ask me why.

dakanji: Fixed

Docs/Configuration.tex
```diff
@@ -1875,18 +1875,21 @@ \subsection{Properties}\label{kernelprops}
...
-  in the \texttt{OSBundleLibraries} key is a dependency that must precede
-  the kext being added.
+  under the key is a dependency that must appear before the kext being added.
 
-  \emph{Note 3}: Kexts may have inner kexts (\texttt{Plugins}) included
-  in the bundle and each such plugin must be added separately.
+  \emph{Note 3}: Kexts may have inner kexts (\texttt{Plugins}) included in the bundle.
+  Such plugins must be added separately.
+
+  \emph{Note 4}: Some inner kexts may in fact be dependencies required by the host kext and
```
vit9696, line 1890: Is such level of detail really needed?

dakanji: At least no body reading it will ever have any doubt or make any mistake. 
Can't be a bad thing don't you think?

vit9696: In my opinion, it takes the wrong approach really, what we have here is a total order over the OSBundleLibraries. Plugin kexts are simply not an exception.

dakanji: Lost now.

Better to close and leave as it was?

vit9696: Kexts may have inner kexts (\texttt{Plugins}) included in the bundle. Such plugins must be added separately. \texttt{Plugin} kext follow the same global ordering rules as all other kexts.

PR #248, opened 2021-04-02
OpenCanopy: Align cursor speed with UIScale

crwnet:


mhaeuser:
1. No description what this change is supposed to be good for.
2. AppleEvent already scales by UIScale on its own, this would scale it twice. See 1.
3. The pointer speed is likely going to be configurable soon.

PR #252, opened 2021-04-10
OcCryptoLib: Added Avx-acceleration of SHA512 and SHA384.

MikhailKrichanov:
Let's add Avx-acceleration of SHA512 and SHA384.


Library/OcCryptoLib/OcCryptoLib.inf
```diff
@@ -34,6 +34,7 @@
   Md5.c
   RsaDigitalSign.c
   Sha1.c
+  Sha2-avx.h
```
vit9696, line 37: I think EDK II prefers CamelCase names for the files as well, so better to have Sha2Avx.h.

Library/OcCryptoLib/OcCryptoLib.inf
```diff
@@ -47,6 +48,8 @@
 
 [Sources.X64]
   X64/BigNumWordMul64.c
+  X64/Sha2-avx.c
```
vit9696, line 51: Similar changes here. You will also need to provide a dummy implementation for Ia32, so that while it never supports AVX, it always compiles without issues.

I.e. have all your functions with `ASSERT (FALSE);` contents and return `FALSE` from the AVX detection function.

Library/OcCryptoLib/Sha2-avx.h
```diff
@@ -0,0 +1,92 @@
...
+  This program and the accompanying materials
+  are licensed and made available under the terms and conditions of the BSD License
+  which accompanies this distribution.  The full text of the license may be found at
+  http://opensource.org/licenses/bsd-license.php
+
+  THE PROGRAM IS DISTRIBUTED UNDER THE BSD LICENSE ON AN "AS IS" BASIS,
+  WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
+**/
+
+#ifndef SHA2_AVX_H
```
vit9696, line 15: I think that header is now common between AVX and non-AVX version, so better to rename to `Sha2Internal.h` or something like that.

Library/OcCryptoLib/Sha2-avx.h
```diff
@@ -0,0 +1,92 @@
...
+  UINT64 TotalLength;
+  UINT8  Block[2 * SHA512_BLOCK_SIZE];
+} SHA512_STATE;
+
+typedef SHA512_STATE SHA384_STATE;
+
+//
+// Sha 384 Init State
+//
+STATIC CONST UINT64 SHA384_H0[8] = {
```
vit9696, line 57: In my opinion, better make these declarations externs, i.e. `extern CONST UINT64 SHA384_H0[8];` and provide the definition in a separate C file. That will reduce the resulting binary size by avoiding the duplication of these arrays. While not by much, I believe it is worth it.

Library/OcCryptoLib/Sha2-avx.h
```diff
@@ -0,0 +1,92 @@
...
+VOID
+Sha384Avx (
+  IN OUT UINT8        *Hash,
+  IN     CONST UINT8  *Data,
+  IN     UINTN        Len
+  );
+
+BOOLEAN
+EFIAPI
+IsAvxSupported ();
```
vit9696, line 90: I think this function is more about trying to enable AVX on the platform and return TRUE on success. Maybe rename it accordingly? Should also have void in functions not taking arguments in C.

```
BOOLEAN
EFIAPI
TryEnableAvx (
  VOID
  );
```

Library/OcCryptoLib/Sha2-avx.h
```diff
@@ -0,0 +1,92 @@
...
+           | LShiftU64 (*((str) + 1), 48)        \
+           | LShiftU64 (*((str) + 0), 56);       \
+  } while (0)
+
+typedef struct {
+  UINT64 State[8];
+  UINT64 Length;
+  UINT64 TotalLength;
+  UINT8  Block[2 * SHA512_BLOCK_SIZE];
+} SHA512_STATE;
```
vit9696, line 50: Could you please make the state common between AVX and non-AVX functions? I guess it will be easier to use this state and update C versions, since they are mostly renames anyway.

The primary reason for this is that I would like Sha384Update to use AVX rather than have separate branching in Sha384. See below.

Library/OcCryptoLib/Sha2.c
```diff
@@ -494,6 +450,11 @@ Sha512 (
 {
   SHA512_CONTEXT  Ctx;
 
+  if (IsAvxSupported ()) {
```
vit9696, line 453: There are places where we use call Sha384Update directly, for example, where we update the hash from non-contiguous data. For this use the branching for AVX and non-AVX version needs to happen inside Sha384Update rather than here.

Library/OcCryptoLib/Sha2.c
```diff
@@ -599,6 +560,11 @@ Sha384 (
 {
   SHA384_CONTEXT Ctx;
 
+  if (IsAvxSupported ()) {
```
vit9696, line 563: I would suggest not making a detection check for AVX implicitly. While many firmwares will not have any issues, I would still try to avoid doing it without explicit user consent. For this reason I suggest the following:

- Make a `TryEnableAvx` function (ex `IsAvxSupported`) accessible from `OcCryptoLib.h`
- Introduce a `extern BOOLEAN mIsAvxEnabled` inside Sha2Internals.h, which gets assigned by this function.
- In all the Sha2 functions check the boolean as needed instead of trying to auto-enable AVX.

In OC codebase we will call the function ourselves later as it will probably need some configuration thought. For Amaranth just go with   AmStorageLib probably?

CC @mhaeuser

Library/OcCryptoLib/X64/Sha2-avx.c
```diff
@@ -0,0 +1,227 @@
...
+
+VOID
+Sha512InitAvx (
+  IN OUT SHA512_STATE  *Context
+  )
+{
+  UINTN  Index;
+
+  for (Index = 0; Index < 8; ++Index) {
+    Context->State[Index] = SHA512_H0[Index];
```
vit9696, line 38: Cannot we use CopyMem here?

Also, I get a feeling that after the unification you would only have custom versions of Sha512UpdateAvx / Sha384UpdateAvx with Init and Final being common between the implementations.

Library/OcCryptoLib/X64/sha512-avx.nasm
```diff
@@ -0,0 +1,532 @@
+; @file
+; Copyright (C) 2021, ISP RAS. All rights reserved.
```
vit9696, line 2: This one is based on Intel code even if modified, so better to have their copyrights as well.

Library/OcCryptoLib/X64/sha512-avx.nasm
```diff
@@ -0,0 +1,532 @@
...
+section .rodata
+align 16
+; Mask for byte-swapping a couple of qwords in an XMM register using (v)pshufb.
+XMM_QWORD_BSWAP:
+	dq 0x0001020304050607,0x08090a0b0c0d0e0f
+
+align 64
+; K[t] used in SHA512 hashing
+K512:
+	dq 0x428a2f98d728ae22,0x7137449123ef65cd
```
vit9696, line 32: These constants are already made available in SHA512_K of the C code. Since AVX version needs custom alignment, in my opinion, drop the C version and make this symbol globally accessible for C code to use. Doing this will also reduce the resulting code size.

Library/OcCryptoLib/X64/sha512-avx.nasm
```diff
@@ -0,0 +1,532 @@
...
+  and ecx, 014000000H
+  cmp ecx, 014000000H ; check both XSAVE and AVX feature flags
+  jne noAVX
+  ; processor supports AVX instructions
+  mov rax, cr4
+  bts rax, 18  ; OSXSAVE: enables XGETBV and XSETBV
+  mov cr4, rax
+
+  mov ecx, 0          ; read the contents of XCR0 register
+  XGETBV              ; result in EDX:EAX
```
vit9696, line 414: I guess that was written in upper case by accident, change to lower here and below.

Library/OcCryptoLib/X64/sha512-avx.nasm
```diff
@@ -0,0 +1,532 @@
...
+	mov	rax, rsp
+	sub rsp, frame_size
+	and	rsp, ~(0x20 - 1)
+	mov	[rsp + frame_RSPSAVE], rax
+
+	; Save GPRs
+  ; Registers RBX, RBP, RDI, RSI, R12, R13, R14, R15, and XMM6-XMM15 are nonvolatile,
+  ; but XMM6-XMM15 are not used.
+	mov     [rsp + frame_GPRSAVE], rbx
+  mov     [rsp + frame_GPRSAVE + 8*1], rbp
```
vit9696, line 449: I think there is a mix of tabs and spaces in the asm code used for indentation, could you update to two spaces everywhere?

Library/OcCryptoLib/X64/sha512-avx.nasm
```diff
@@ -0,0 +1,532 @@
...
+;  Purpose: Updates the SHA512 digest stored at "state" with the message
+;  stored in "data".
+;  The size of the message pointed to by "data" must be an integer multiple
+;  of SHA512 message blocks.
+;  "blocks" is the message length in SHA512 blocks
+; #######################################################################
+align 8
+global ASM_PFX(Sha512TransformAvx)
+ASM_PFX(Sha512TransformAvx):
+	test msglen, msglen
```
vit9696, line 436: I wonder whether we need to
- protect xmm registers by saving them on stack before doing these transformations in case some firmwares decide not to do this.
- disable interrupts before calling this function.

Want to hear @mhaeuser on this first.

vit9696 (commented):
Big thanks for doing this. Made the comments inline. Please also have a look on what the CI has found.

Library/OcCryptoLib/Sha2.c
```diff
@@ -583,11 +589,13 @@ Sha384Final (
...
-
-  for (Index = 0 ; Index < 6; ++Index) {
-    UNPACK64 (Context->State[Index], &HashDigest[Index << 3]);
+  if (mIsAvxEnabled) {
+    Sha512TransformAvx (Context->State, Context->Block, BlockNb);
+  } else {
+    Sha512Transform (Context, Context->Block, BlockNb);
   }
+
+  CopyMem (HashDigest, Context->State, SHA512_DIGEST_SIZE);
```
vit9696, line 598: I noticed there is a change in the logic for non-AVX version, is it still correct?

Library/OcCryptoLib/Ia32/Sha512AvxDummy.nasm
```diff
@@ -0,0 +1,14 @@
+BITS 32
```
vit9696, line 1: Might want a copyright header.

Library/OcCryptoLib/Sha2.c
```diff
@@ -91,7 +91,7 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
           + SHA512_SIG0(W[Index - 15]) + W[Index - 16];     \
   } while(0)
 
-
+BOOLEAN mIsAvxEnabled;
```
vit9696, line 94: This variable should be assigned in `TryEnableAvx`.

Library/OcCryptoLib/Sha2.c
MikhailKrichanov: According to APPENDIX C: SHA-512 EXAMPLES in FIPS Publication 180-2, non-AVX version became correct only after that change.

Library/OcCryptoLib/Sha2.c
```diff
@@ -447,8 +451,13 @@ Sha512Update (
...
 
-  Sha512Transform (Context, Context->Block, 1);
-  Sha512Transform (Context, ShiftedMsg, BlockNb);
+  if (mIsAvxEnabled) {
+    Sha512TransformAvx (Context->State, Context->Block, 1);
+    Sha512TransformAvx (Context->State, ShiftedMsg, BlockNb);
+  } else {
+    Sha512Transform (Context, Context->Block, 1);
+    Sha512Transform (Context, ShiftedMsg, BlockNb);
+  }
```
MikeBeaton, line 460: Was there a strong reason for having the signatures of `Sha512Transform` and `Sha512TransformAvx` different? (e.g. can perhaps update signature of `Sha512Transform` to match new version - unless there is some requirement I'm not aware of for keeping it as is).
Suggesting this tentatively - happy to be overruled by @vit9696

MikeBeaton (changes requested):

Library/OcCryptoLib/X64/Sha512Avx.nasm
```diff
@@ -0,0 +1,493 @@
...
+  ; Registers RBX, RBP, RDI, RSI, R12, R13, R14, R15, and XMM6-XMM15 are nonvolatile,
+  ; but XMM6-XMM15 are not used.
+  mov [rsp + frame_GPRSAVE], rbx
+  mov [rsp + frame_GPRSAVE + 8*1], rbp
+  mov [rsp + frame_GPRSAVE + 8*2], rdi
+  mov [rsp + frame_GPRSAVE + 8*3], rsi
+  mov [rsp + frame_GPRSAVE + 8*4], r12
+  mov [rsp + frame_GPRSAVE + 8*5], r13
+  mov [rsp + frame_GPRSAVE + 8*6], r14
+  mov [rsp + frame_GPRSAVE + 8*7], r15
```
vit9696, line 416: Unfortunately we cannot guarantee in UEFI world that xmm0~xmm5 actually are volatile. This is because while UEFI does follow Microsoft 64-bit ABI, it does not (officially) support vector registers as a part of the context. This means that:
- a context switch may happen and clutter the xmm registers
- another function might have forgotten about that and was itself using xmm registers when our code interrupted it.

To workaround this the following changes need to be made:
- We should save all the xmm registers we use here on the stack.
- We should restore all the saved xmm registers before exiting the function.
- We should save flags register here on the stack and disable interrupts (e.g. via cli).
- We should reenable the interrupts if they were previously enabled (can check the saved flags) in the end of the function.

MikeBeaton: Is the vector register issue here more about security (e.g. a maliciously crafted service could interrupt this code and cause faked hashes) or more about correctness (e.g. this code gets interrupted, the registers get trashed, so the hash is incorrect)?

For the latter, is there any scope for 'working round' it by only enabling this code on a settings flag, and documenting that it _may_ be incompatible with some other drivers?

For the former, I guess we can disable some but not all interrupts? Therefore cannot completely prevent the issue? And also that the same issue would in theory apply to the pre-existing, non-avx algorithm (i.e. not a 'to spec' interrupt, but a malicious one, interrupting and changing things).

Those are just some comments on general principles - again offered somewhat tentatively - but in case anything I'm saying here seems useful to @vit9696 to feed back into what he's requesting.

vit9696: The issue is not about security: if malicious executes, we are doomed in the first place. The issue is that we need to ensure compatibility, and this sounds like the only possible way.

User/Library/UserMisc.c
```diff
@@ -8,6 +8,8 @@
 #include <Library/UefiApplicationEntryPoint.h>
 #include <Library/DebugLib.h>
 
+BOOLEAN mIsAvxEnabled;
```
mhaeuser, line 11: There is a convention that any global variable declaration that is not STATIC gets GLOBAL_REMOVE_IF_UNREFERENCED.

MikeBeaton (approved):

Library/OcCryptoLib/Sha2.c
MikeBeaton: Thank you for the update

Library/OcCryptoLib/Ia32/Sha512AvxDummy.nasm
```diff
@@ -0,0 +1,29 @@
...
+; http://opensource.org/licenses/bsd-license.php
+;
+; THE PROGRAM IS DISTRIBUTED UNDER THE BSD LICENSE ON AN "AS IS" BASIS,
+; WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
+;
+BITS 32
+
+section .bss
+global ASM_PFX(mIsAvxEnabled)
+ASM_PFX(mIsAvxEnabled): db 0
```
mhaeuser, line 16: I'm not sure a .bss section will work well with CLANGPDB and MSVC. EDK II seems to merge .bss and .data sections into a .data segment for ELFs, which is good, but I think those that compile to PE directly would create a separate .bss PE section, which wastes space. Can the declaration maybe just be moved to C so the compiler can choose where to place it?

cc @vit9696

Library/OcCryptoLib/Ia32/Sha512AvxDummy.nasm
```diff
@@ -0,0 +1,29 @@
...
+
+section .text
+
+align 8
+global ASM_PFX(TryEnableAvx)
+ASM_PFX(TryEnableAvx):
+  mov eax, 0
+  ret
```
mhaeuser, line 24: Why not use your code from UserMisc.c instead of NASM?

Library/OcCryptoLib/X64/Sha512Avx.nasm
```diff
@@ -0,0 +1,511 @@
...
+
+extern ASM_PFX(SHA512_K)
+
+section .rodata
+align 16
+; Mask for byte-swapping a couple of qwords in an XMM register using (v)pshufb.
+XMM_QWORD_BSWAP:
+	dq 0x0001020304050607,0x08090a0b0c0d0e0f
+
+section .bss
```
mhaeuser, line 30: See https://github.com/acidanthera/OpenCorePkg/pull/252/files#r614912101

Library/OcCryptoLib/X64/Sha512Avx.nasm
```diff
@@ -0,0 +1,511 @@
...
+  mov ecx, 0          ; read the contents of XCR0 register
+  xgetbv              ; result in EDX:EAX
+  or  eax, 06H        ; enable both XMM and YMM state support
+  ; XSETBV must be executed at privilege level 0 or in real-address mode.
+  xsetbv
+  mov rax, 1
+  mov byte [rel ASM_PFX(mIsAvxEnabled)], 1
+  jmp done
+noAVX:
+  mov rax, 0
```
mhaeuser, line 382: xor maybe here and above? :)

User/Library/UserMisc.c
```diff
@@ -672,3 +674,23 @@ DevicePathFromHandle (
...
+
+BOOLEAN
+EFIAPI
+TryEnableAvx (
+  VOID
+  )
+{
+  mIsAvxEnabled = FALSE;
+  return FALSE;
+}
```
MikeBeaton, line 696: Is the asm code actually plumbed in, in this version? I would have expected to see a real and a dummy implementation of this method in the ASM files, and no implementation here, and similar for variable `mIsAvxEnabled`.
(cc @mhaeuser )

mhaeuser: @MikeBeaton Vice-versa, I think we want this C code in the lib over the dummy ASM: https://github.com/acidanthera/OpenCorePkg/pull/252#discussion_r614914319

MikeBeaton: I saw the comment, hence the cc to yourself - was not expecting that link would work with > 1 implementation - possibly I'm just not aware of how this is getting resolved.

Trying to compile and run this, have spotted that multiple utilities are also using Sha2.c (hence currently linking with the .c implementation, so definitely not using this acceleration). I guess as part of final commit, we'd want to make the tools link to this new implementation too?

MikeBeaton: Apologies, I see now that the versions here _are_ the versions designed to link into the utilities.

However, would still want to suggest (@vit9696 @mhaeuser ?) that it might be ideal to link the accelerated version into the utilities, too.

mhaeuser: @MikeBeaton cli will throw an exception, imo not worth the effort

Library/OcCryptoLib/Sha2.c
```diff
@@ -324,11 +324,19 @@ Sha256 (
...
+  );
+
 VOID
 Sha512Transform (
-  SHA512_CONTEXT  *Context,
-  CONST UINT8     *Data,
-  UINTN           BlockNb
+  UINT64      *State,
+  CONST UINT8 *Data,
+  UINTN       BlockNb
```
MikeBeaton, line 339: Would still suggest updating the original `Sha512Transform` signature completely, to match the now more carefully constructed signature of your new method, including use of `IN`, `OUT`, here and in the method definition.

Library/OcCryptoLib/Sha2.c
MikeBeaton: Can confirm that it all works, going from about 8 secs to about 3 secs to run the 5,000,000 x password hash OC password verification.

As expected (presumably?) pre-existing password hash, salt no longer work, and have to be regenerated using new ocpasswordgen built against this changed version.

Library/OcCryptoLib/Ia32/Sha512AvxDummy.nasm
MikeBeaton: Second the idea of declaring the var in C.

Library/OcCryptoLib/X64/Sha512Avx.nasm
```diff
@@ -0,0 +1,511 @@
...
+  vpaddq xmm0, xmm0, K_t(idx) ; Compute W[t]+K[t]
+  vmovdqa WK_2(idx), xmm0     ; Store W[t]+K[t] for next rounds
+%endmacro
+
+; #######################################################################
+; BOOLEAN TryEnableAvx ()
+; To run in QEMU use options: -enable-kvm -cpu Penryn,+avx,+xsave,+xsaveopt
+; #######################################################################
+align 8
+global ASM_PFX(TryEnableAvx)
```
MikeBeaton, line 359: What's the plan for calling this to enable new version other than manually, as now? On config.plist setting presumably.

cc @vit9696

User/Library/UserMisc.c
MikeBeaton: Okay, will take @mhaeuser's steer that using this outside OC itself is more trouble than it's worth, in which case nothing else to add on this one.

Library/OcCryptoLib/Sha2.c
vit9696: Huh? How so? @mhaeuser was our previous implementation incorrect or what?

@MikeBeaton does Mac Secure Boot work?

Library/OcCryptoLib/X64/Sha512Avx.nasm
vit9696: Yes, a config.plist setting: UEFI → Quirks → EnableVectorAcceleration. Separate from the change in my opinion.

Library/OcCryptoLib/Sha2.c
MikeBeaton: > @MikeBeaton does Mac Secure Boot work?

No, unfortunately not with this new implementation - and yes back on the previous one.

@MikhailKrichanov

mhaeuser: Is it confirmed it is *this* change that causes the issue?

mhaeuser: @MikhailKrichanov No, the original was correct.
> Throughout this specification, the “big-endian” convention is used when expressing both 32- and 64-bit words, so that within each word, the most significant bit is stored in the left-most bit position.

As calculations internally are done in little-endian order (e.g. you can see SHA*_H0 are stored word- and not byte-wise matching the spec pattern), the hash must be converted word-wise on finalisation. That is what the UNPACK code did.

MikeBeaton: Can confirm that restoring the `UNPACK64` `for` loop, in the two places where this was changed to a `CopyMem`, changes back to working fine with previous password hashes and to passing Apple Secure Boot again.

cc @mhaeuser @vit9696 @MikhailKrichanov

MikeBeaton: @MikhailKrichanov -

Some timings from the 'fixed' code (i.e. your PR, just slightly modified as above to restore `UNPACK64` in the two places where it was in fact required to match, e.g. Apple Secure Boot, previous OC passwords, etc.), compiled OC password verify times:

 - DEBUG: ASM: 5.3-5.4s C: 10.0-10.1s
 - RELEASE: ASM: 3.5-3.7s C: 4.8-4.9s
 
Just noting that while certainly not completely negligible, the speed up of ASM vs C in RELEASE is not _huge_ either... is this all in fact as expected? (Or am I - possibly - or the new code - possibly - doing something wrong?)

Cheers!

MikeBeaton: Okay, so @vit9696 has mentioned that the existing speedup sounds about exactly right to him - and I gather @vit9696 and @mhaeuser both think that any additional remaining speedup to be gained would be in the shared code path (specifically `UNPACK64`) so it all sounds about right, speed-wise - many thanks!

MikeBeaton: From a little more research, `UNPACK64` appears to be already optimised as a `bswap` in RELEASE build (only), so we may already (now, with this PR, that is!) be at peak SHA. 😊

Library/OcCryptoLib/Ia32/Sha512AvxDummy.nasm
MikhailKrichanov: Will such kind of dependencies look good  in a separate library module?

User/Library/UserMisc.c
```diff
@@ -672,3 +674,23 @@ DevicePathFromHandle (
   return NULL;
 }
 
+VOID
+EFIAPI
+Sha512TransformAvx (
```
vit9696, line 679: Move these to `OcCryptoLib/Sha512AvxDummy.c` and remove `Ia32/Sha512AvxDummy.nasm`. Then update `User/Makefile` to include `Sha512AvxDummy.o` in:

```
	#
	# OcCryptoLib targets.
	#
	OBJS    += RsaDigitalSign.o BigNumMontgomery.o BigNumPrimitives.o BigNumWordMul64.o Sha2.o SecureMem.o
```

Library/OcCryptoLib/OcCryptoLib.inf
```diff
@@ -44,9 +44,11 @@
 
 [Sources.Ia32]
   Ia32/BigNumWordMul64.c
+  Ia32/Sha512AvxDummy.nasm
```
vit9696, line 47: Change that to Sha512AvxDummy.c.

Library/OcCryptoLib/Ia32/Sha512AvxDummy.nasm
vit9696: In my opinion, just a separate .c file within OcCryptoLib. I replied below.

vit9696:
Merged, thanks a lot!

MikeBeaton:
PS I double checked the timings, it decreases from ~4.80s in C to ~3.45s in AVX asm for OC password verification hash time on RELEASE build. Vit said the expected timing increase from the Intel paper was 30-40%, and that Δt is 30% of the larger time and 40% of the smaller - so pretty much spot on! Is looking excellent, btw. :)

PR #253, opened 2021-04-15
Removed dead code

Gengik84:
DTGP is not used in _DSM, so this is died code


matatablack:
Hi! Don't really understand whats happening but this commit broke the example. Needed to add the code back to get it working! 
Thanks for all the work you are doing guys!

PR #259, opened 2021-05-14
OcDevicePathLib: Apply separator to multi-node paths

MikeBeaton:
Fixes https://github.com/acidanthera/bugtracker/issues/1637

### DIFF
--- a/Library/OcDevicePathLib/OcDevicePathLib.c
+++ b/Library/OcDevicePathLib/OcDevicePathLib.c
@@ -1067,7 +1067,10 @@ OcFileDevicePathFullNameLen (
     }
 
     FilePath    = (FILEPATH_DEVICE_PATH *)DevicePath;
-    PathLength += OcFileDevicePathNameLen (FilePath);
+    //
+    // Each node requires separator or CHAR_NULL
+    //
+    PathLength += OcFileDevicePathNameLen (FilePath) + 1;
 
     DevicePath = NextDevicePathNode (DevicePath);
   } while (!IsDevicePathEnd (DevicePath));
@@ -1090,7 +1093,7 @@ OcFileDevicePathFullNameSize (
   IN CONST EFI_DEVICE_PATH_PROTOCOL  *DevicePath
   )
 {
-  return (OcFileDevicePathFullNameLen (DevicePath) + 1) * sizeof (CHAR16);
+  return OcFileDevicePathFullNameLen (DevicePath) * sizeof (CHAR16);
 }
 
 /**
@@ -1118,10 +1121,6 @@ OcFileDevicePathFullName (
   ASSERT (IsDevicePathValid (&FilePath->Header, 0));
   ASSERT (PathNameSize == OcFileDevicePathFullNameSize (&FilePath->Header));
 
-  //
-  // FIXME: Insert separators between nodes if not present already.
-  //
-
   do {
     PathLen = OcFileDevicePathNameLen (FilePath);
     CopyMem (
@@ -1130,10 +1129,11 @@ OcFileDevicePathFullName (
       PathLen * sizeof (*FilePath->PathName)
       );
     PathName += PathLen;
+    *PathName++ = L'\\';
 
     FilePath = (CONST FILEPATH_DEVICE_PATH *)NextDevicePathNode (FilePath);
   } while (!IsDevicePathEnd (FilePath));
-  *PathName = CHAR_NULL;
+  *(PathName - 1) = CHAR_NULL;
 }
 
 CHAR16 *
@@ -1277,7 +1277,7 @@ OcDevicePathHasFilePathSuffix (
     return FALSE;
   }
 
-  PathNameLen = OcFileDevicePathFullNameLen (&FilePath->Header);
+  PathNameLen = OcFileDevicePathFullNameLen (&FilePath->Header) - 1;
   if (PathNameLen < SuffixLen) {
     return FALSE;
   }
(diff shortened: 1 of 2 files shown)


MikeBeaton:
Before:

WhatsApp Image 2021-05-14 at 02 59 42

After:

WhatsApp Image 2021-05-14 at 03 09 44

Library/OcDevicePathLib/OcDevicePathLib.c
```diff
@@ -1118,10 +1121,6 @@ OcFileDevicePathFullName (
   ASSERT (IsDevicePathValid (&FilePath->Header, 0));
   ASSERT (PathNameSize == OcFileDevicePathFullNameSize (&FilePath->Header));
 
-  //
-  // FIXME: Insert separators between nodes if not present already.
-  //
-
```
MikeBeaton, line 1124: Based on MdePkg DevicePathToText.c https://github.com/acidanthera/audk/blob/master/MdePkg/Library/UefiDevicePathLib/DevicePathToText.c#L2433 I believe it is reasonable to assume that all nodes always need a separator.

Library/OcDevicePathLib/OcDevicePathLib.c
```diff
@@ -1090,7 +1093,7 @@ OcFileDevicePathFullNameSize (
   IN CONST EFI_DEVICE_PATH_PROTOCOL  *DevicePath
   )
 {
-  return (OcFileDevicePathFullNameLen (DevicePath) + 1) * sizeof (CHAR16);
```
MikeBeaton, line 1093: It looks like the contract of this method `@retval 0  DevicePath does not start with a File Path node or contains non-terminating nodes that are not File Path nodes` was not being met before? (The return value is the same as before, with this change, when there is a path, but would have been 1 with no path.)

vit9696:
I am quite sure multinode paths must not contain slashes in them. Next node is enough for them to be assumed as part of the path.

CC @mhaeuser

Library/OcDevicePathLib/OcDevicePathLib.c
mhaeuser: I don't understand? We return 0 if we encounter a non-file node except for the DP terminator.

Library/OcDevicePathLib/OcDevicePathLib.c
mhaeuser: The spec says they are optional, on both ends actually. I don't understand why this garbage is supported anyway, but should maybe leave a note about it?

Library/OcDevicePathLib/OcDevicePathLib.c
MikeBeaton: _This_ method `OcFileDevicePathFullNameSize` (if you expand up, I could not add this comment on the right line, really, as GitHub isn't showing it as part of the diff lines).

So previously (and now!) `OcFileDevicePathFullNameLen` returns 0 for non-file, and therefore _this_ method `OcFileDevicePathFullNameSize` (as it was) returned 2 (`(0 + 1) * sizeof (CHAR16)`) at this point.

Now, with this change, both will return 0 (`0 * sizeof (CHAR16)` from here).

I believe it is correct now and was wrong before?

Library/OcDevicePathLib/OcDevicePathLib.c
MikeBeaton: I'll add back in a note saying that it's supposedly possible but we haven't seen it (and is not currently supported here), at the place of this removed comment.

I'll hold off a sec, until we've thrashed out the previous point!

Library/OcDevicePathLib/OcDevicePathLib.c
MikeBeaton: (Just to be clear, I'm basing this on the @retval comment if this method, which I already quoted - which *says* this method will return 0 if not a file, but it would actually have returned 2 before. I think?!)

Library/OcDevicePathLib/OcDevicePathLib.c
MikeBeaton: @mhaeuser Do you have a link to the documentation (quick search has not yet turned it up for me)? Thank you.

Library/OcDevicePathLib/OcDevicePathLib.c
mhaeuser: Ugh, yes, GitHub diff messed with me. Thank you :)

Library/OcDevicePathLib/OcDevicePathLib.c
mhaeuser: image

Library/OcDevicePathLib/OcDevicePathLib.c
```diff
@@ -1277,7 +1277,7 @@ OcDevicePathHasFilePathSuffix (
     return FALSE;
   }
 
-  PathNameLen = OcFileDevicePathFullNameLen (&FilePath->Header);
+  PathNameLen = OcFileDevicePathFullNameLen (&FilePath->Header) - 1;
```
mhaeuser, line 1280: This can now wrap around and bork the check below. Please purge the subtraction, change the operator below (<=) and remove the addition further down.

MikeBeaton: Thx - I probably shouldn't really have changed `OcFileDevicePathFullNameLen` from producing a normal strlen value (not including terminator); I'll fix it by changing that instead.

MikeBeaton: Believe all correct now - https://github.com/acidanthera/OpenCorePkg/commit/991cd3b2dce69823a25803f0773306e691917a21

Sorry about that - lazy! - tyvm for review

PR #263, opened 2021-05-24
BootManagement: Add `csrutil` EFI tool and Toogle SIP boot option

MikeBeaton:
WIP - going to add some comments


Include/Acidanthera/Library/OcBootManagementLib.h
```diff
@@ -785,9 +790,9 @@ struct OC_PICKER_CONTEXT_ {
   //
   OC_CUSTOM_DESCRIBE         CustomDescribe;
   //
-  // Context to pass to CustomRead and CustomDescribe, optional.
+  // Storage context.
   //
-  VOID                       *CustomEntryContext;
+  OC_STORAGE_CONTEXT         *StorageContext;
```
MikeBeaton, line 795: I already abused the fact that CustomEntryContext contained OC_STORAGE_CONTEXT in the flavour code, I thought it was better to make it explicit.

Include/Acidanthera/Library/OcBootManagementLib.h
```diff
@@ -52,12 +52,15 @@ typedef struct OC_HOTKEY_CONTEXT_ OC_HOTKEY_CONTEXT;
 #define OC_MENU_DISK_IMAGE           L" (dmg)"
 #define OC_MENU_SHUTDOWN             L"Shutting Down"
 #define OC_MENU_RESTART              L"Restarting"
+#define OC_MENU_SIP_IS_DISABLED      L"SIP: Disabled"
+#define OC_MENU_SIP_IS_ENABLED       L"SIP: Enabled"
```
MikeBeaton, line 56: It is very confusing when these are labelled "Enable SIP" and "Disable SIP" - or at least I found it so.

Library/OcBootManagementLib/BootAudio.c
```diff
@@ -194,6 +194,14 @@ OcPlayAudioEntry (
     OcPlayAudioFile (Context, OcVoiceOverAudioFileExternalOS, FALSE);
   } else if (Entry->Type == OC_BOOT_RESET_NVRAM) {
     OcPlayAudioFile (Context, OcVoiceOverAudioFileResetNVRAM, FALSE);
+  } else if (Entry->Type == OC_BOOT_TOGGLE_SIP) {
+    if (StrCmp (Entry->Name, OC_MENU_SIP_IS_DISABLED) == 0) {
+      OcPlayAudioFile (Context, OcVoiceOverAudioFileEnableSIP, FALSE);
+    } else if (StrCmp (Entry->Name, OC_MENU_SIP_IS_ENABLED) == 0) {
+      OcPlayAudioFile (Context, OcVoiceOverAudioFileDisableSIP, FALSE);
```
MikeBeaton, line 201: These sound files are needed.

Library/OcMainLib/OpenCoreMisc.c
```diff
@@ -248,7 +247,7 @@ STATIC
 EFI_STATUS
 EFIAPI
 OcToolDescribeEntry (
```
MikeBeaton, line 249: Is this method still needed? Is this needed? I have replaced the code to do with system tool icons with [this line](https://github.com/acidanthera/OpenCorePkg/pull/263/files#diff-7ea53dcd0ee7bb9a7c053fd83a44b3b6fc4833cc3a22dbd22cd4b75fe59b02f8L1378-R1378) (!). `.icns` was removed in the flavours commit. So all that remains in here is support for `.lbl`/`.l2x`. Does anyone use this?!

Platform/OpenCanopy/GuiApp.c
```diff
@@ -44,7 +44,9 @@ mLabelNames[LABEL_NUM_TOTAL] = {
   [LABEL_OTHER]                = "Other",
   [LABEL_TOOL]                 = "Tool",
   [LABEL_RESET_NVRAM]          = "ResetNVRAM",
-  [LABEL_SHELL]                = "Shell"
+  [LABEL_SHELL]                = "Shell",
+  [LABEL_SIP_IS_ENABLED]       = "SIPEnabled",
+  [LABEL_SIP_IS_DISABLED]      = "SIPDisabled"
```
MikeBeaton, line 49: I have put these labels in a `toggle-sip` branch of `OcBinaryData`.

Library/OcBootManagementLib/BootAudio.c
```diff
@@ -194,6 +194,14 @@ OcPlayAudioEntry (
     OcPlayAudioFile (Context, OcVoiceOverAudioFileExternalOS, FALSE);
   } else if (Entry->Type == OC_BOOT_RESET_NVRAM) {
     OcPlayAudioFile (Context, OcVoiceOverAudioFileResetNVRAM, FALSE);
+  } else if (Entry->Type == OC_BOOT_TOGGLE_SIP) {
+    if (StrCmp (Entry->Name, OC_MENU_SIP_IS_DISABLED) == 0) {
+      OcPlayAudioFile (Context, OcVoiceOverAudioFileSIPIsDisabled, FALSE);
+    } else if (StrCmp (Entry->Name, OC_MENU_SIP_IS_ENABLED) == 0) {
+      OcPlayAudioFile (Context, OcVoiceOverAudioFileSIPIsEnabled, FALSE);
```
MikeBeaton, line 201: These sound files are needed.

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -1396,6 +1396,16 @@ BootPickerEntriesSet (
       case OC_BOOT_RESET_NVRAM:
         Status = CopyLabel (&VolumeEntry->Label, &GuiContext->Labels[LABEL_RESET_NVRAM]);
         break;
+      case OC_BOOT_TOGGLE_SIP:
+        if (StrCmp (Entry->Name, OC_MENU_SIP_IS_DISABLED)) {
+          Status = CopyLabel (&VolumeEntry->Label, &GuiContext->Labels[LABEL_SIP_IS_ENABLED]);
+        } else if (StrCmp (Entry->Name, OC_MENU_SIP_IS_ENABLED)) {
+          Status = CopyLabel (&VolumeEntry->Label, &GuiContext->Labels[LABEL_SIP_IS_DISABLED]);
+        } else {
+          DEBUG ((DEBUG_ERROR, "OCUI: Unsupported toggle sip name\n"));
```
vit9696, line 1405: I think it should be an ASSERT, as this value should be coming from a trusted source.

Platform/OpenCanopy/GuiApp.c
vit9696: OK.

Library/OcMainLib/OpenCoreMisc.c
```diff
@@ -259,37 +258,26 @@ OcToolDescribeEntry (
...
-        sizeof (DescPath),
-        OPEN_CORE_TOOL_PATH "%s.%a",
-        ChosenEntry->PathName,
-        LabelScale == 2 ? "l2x" : "lbl"
-        );
-    }
+    Status = OcUnicodeSafeSPrint (
+      DescPath,
+      sizeof (DescPath),
+      OPEN_CORE_TOOL_PATH "%s.%a",
```
vit9696, line 274: What will this have in case of Reset NVRAM or CSR code? I get a feeling it may contain human readable text, which sounds wrong.

Library/OcMainLib/OpenCoreMisc.c
vit9696: Oh, right, it should then be unused, yes. Please remove that.

vit9696: Hrm, it is not that simple. For flavours we can group the icons, but for the names no decent grouping is really possible: we cannot guess the name of each tool. lbl support needs to stay I believe.

Library/OcBootManagementLib/VariableManagement.c
```diff
@@ -312,6 +314,109 @@ InternalGetBootstrapBootData (
...
+      }
+
+    if (OcIsSipEnabled (Status, CsrConfig)) {
+      CsrConfig = CsrActiveConfig;
+      Status = OcSetSip(&CsrConfig, Attributes);
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "OCB: Error disabling SIP - r\n", Status));
+      }
+    } else {
+      CsrConfig = CSR_ALLOW_APPLE_INTERNAL;
```
vit9696, line 396: Enabled SIP is just 0. This makes no sense as we are not targeting Apple Internal builds here.

Library/OcBootManagementLib/VariableManagement.c
```diff
@@ -312,6 +314,109 @@ InternalGetBootstrapBootData (
...
+  return Status;
+}
+
+BOOLEAN
+OcIsSipEnabled (
+  EFI_STATUS  FetchStatus,
+  UINT32      CsrActiveConfig
+  )
+{
+  return FetchStatus == EFI_NOT_FOUND || (CsrActiveConfig & ~CSR_ALLOW_APPLE_INTERNAL) == 0;
```
vit9696, line 363: I do not understand why Apple Internal is treated differently. It is not like many users can run HomeDiagnostic in the first place, and not many people in Apple would use OpenCore.

Library/OcBootManagementLib/BootEntryManagement.c
```diff
@@ -1555,14 +1563,40 @@ AddFileSystemEntryForCustom (
...
+    Status = OcGetSip (&CsrActiveConfig, NULL);
+    if (Status == EFI_NOT_FOUND || !EFI_ERROR(Status)) {
+      LoadOptions = AllocateCopyPool (sizeof (BootContext->PickerContext->ToggleSipMask), &BootContext->PickerContext->ToggleSipMask);
+      if (LoadOptions == NULL) {
+        Status = EFI_OUT_OF_RESOURCES;
+      } else {
+        Status = AddBootEntryFromSystemEntry (
+          BootContext,
+          FileSystem,
+          OcIsSipEnabled (Status, CsrActiveConfig) ? OC_MENU_SIP_IS_ENABLED : OC_MENU_SIP_IS_DISABLED,
```
vit9696, line 1576: Is this unconditionally initialised regardless of the Status? Sounds like an unusual contract for the function, which I do not like.

Library/OcBootManagementLib/BootEntryInfo.c
```diff
@@ -478,11 +482,15 @@ OcGetBootEntryIcon (
   *ImageData = NULL;
   *DataLength = 0;
 
-  if (BootEntry->Type == OC_BOOT_EXTERNAL_TOOL || BootEntry->Type == OC_BOOT_RESET_NVRAM) {
+  if ((BootEntry->Type & OC_BOOT_SYSTEM) != 0) {
```
vit9696, line 485: Is this legit? Sounds like it must be an ASSERT again.

Library/OcBootManagementLib/BootEntryInfo.c
```diff
@@ -406,11 +406,15 @@ OcGetBootEntryLabelImage (
...
```
vit9696, line 409: Same here?

Library/OcBootManagementLib/BootAudio.c
vit9696: cc @Andrey1970AppleLife

Include/Acidanthera/Library/OcBootManagementLib.h
vit9696: Fine to me.

Application/CsrUtil/CsrUtil.c
```diff
@@ -0,0 +1,187 @@
...
+  UINTN                                 Data;
+  UINTN                                 Len;
+  CHAR16                                *EndPtr;
+  UINTN                                 Argc;
+  CHAR16                                **Argv;
+  UINT32                                CsrConfig;
+  UINT32                                Attributes;
+  BOOLEAN                               Usage;
+
+  Status = GetArguments (&Argc, &Argv);
```
vit9696, line 55: I think outside of Shell only 1 argument is supported.

Application/CsrUtil/CsrUtil.c
```diff
@@ -0,0 +1,187 @@
...
+  //
+  if (Argc > 2) {
+    Data = 0;
+    Len = StrLen (Argv[2]);
+    if (Len > sizeof (Data) * 2) {
+      Status = EFI_UNSUPPORTED;
+    } else {
+      Status = StrHexToUintnS (Argv[2], &EndPtr, &Data);
+      if (!EFI_ERROR (Status)) {
+        if (EndPtr != &Argv[2][Len] || (Data & MAX_UINT32) != Data) {
```
vit9696, line 78: Any particular reason this is a separate if?

Application/CsrUtil/CsrUtil.c
```diff
@@ -0,0 +1,187 @@
...
+          Status = EFI_UNSUPPORTED;
+        }
+      }
+    }
+  }
+
+  Usage = TRUE;
```
vit9696, line 85:
```C
if (EFI_ERROR (Status)) {
  PrintUsage ();
  return Status;
}
```

No?

Platform/OpenCanopy/GuiApp.c
vit9696: CC @Andrey1970AppleLife

Library/OcBootManagementLib/VariableManagement.c
```diff
@@ -312,6 +314,109 @@ InternalGetBootstrapBootData (
...
+
+  *CsrActiveConfig = 0;
+  DataSize = sizeof (*CsrActiveConfig);
+
+  Status = gRT->GetVariable (
+    L"csr-active-config",
+    &gAppleBootVariableGuid,
+    Attributes,
+    &DataSize,
+    CsrActiveConfig);
```
vit9696, line 334: Trailing brace?

Library/OcBootManagementLib/VariableManagement.c
```diff
@@ -312,6 +314,109 @@ InternalGetBootstrapBootData (
   return OptionData;
 }
 
+EFI_STATUS
+OcGetSip (
+  UINT32 *CsrActiveConfig,
```
vit9696, line 319: IN / OUT / OPTIONAL missing here and below?

Application/CsrUtil/CsrUtil.c
```diff
@@ -0,0 +1,187 @@
...
+#include <Library/OcBootManagementLib.h>
+#include <Library/UefiApplicationEntryPoint.h>
+#include <Library/UefiLib.h>
+#include <IndustryStandard/AppleCsrConfig.h>
+#include <Guid/AppleVariable.h>
+
+STATIC
+VOID
+PrintUsage (
+  )
```
vit9696, line 20: Missing VOID?

Application/CsrUtil/CsrUtil.c
```diff
@@ -0,0 +1,187 @@
...
+{
+  Print (L"usage: csrutil <command>\n");
```
vit9696, line 22: How to use HEX?

Application/CsrUtil/CsrUtil.c
```diff
@@ -0,0 +1,187 @@
...
+  if (!EFI_ERROR (Status)) {
+    Usage = FALSE;
+
+    //
+    // Use existing attributes where present (e.g. keep changes made while WriteFlash = false as volatile only)
+    //
+    Status = OcGetSip (&CsrConfig, &Attributes);
+
+    if (Status != EFI_NOT_FOUND && EFI_ERROR (Status)) {
+      Print (L"Error getting SIP status - %r\n", Status);
```
vit9696, line 95: I think that should not print usage and should just early leave the function.

Docs/Configuration.tex
```diff
@@ -3920,6 +3932,21 @@ \subsection{Security Properties}\label{miscsecurityprops}
...
 
+\item
+  \texttt{ToggleSipMask}\\
+  \textbf{Type}: \texttt{plist\ integer}, 32 bit\\
+  \textbf{Failsafe}: Zero (selects current macOS default)\\
+  \textbf{Description}: Value used for disabling System Integrity Protection
+  in combination with \texttt{AllowToggleSip}.
+
+  \emph{Note1}: If failsafe or zero is specified here, the value used will be
+  correct for the current version of macOS.
```
vit9696, line 3943: I think we can specify the actual value here.

Docs/Flavours.md
```diff
@@ -150,6 +150,15 @@ If providing just one file, name it `NVRAMTool.icns` if it is themed so that it
    - As another example of how flavours work: **ResetNVRAM:NVRAMTool** will look for `ResetNVRAM.icns`, then `NVRAMTool.icns` (and then, by OC default behaviour, `Tool.icns`, then `HardDrive.icns`)
    - _**NB**: Including **ResetNVRAM** anywhere in the flavour triggers picker audio-assist for "Reset NVRAM"_
 
+### SIP Tools
```
vit9696, line 153: In my opinion, this is an NVRAMTool. I.e. SIPTool:NVRAMTool. Thus the category.

Include/Acidanthera/Protocol/OcAudio.h
```diff
@@ -105,8 +105,10 @@ typedef enum {
   OcVoiceOverAudioFileShutDown              = 0x104A,
   OcVoiceOverAudioFileRestart               = 0x104B,
   OcVoiceOverAudioFileDiskImage             = 0x104C,
+  OcVoiceOverAudioFileSIPIsDisabled         = 0x104D,
+  OcVoiceOverAudioFileSIPIsEnabled            = 0x104E,
```
vit9696, line 109: Extra space?

vit9696: Also, the list was sorted originally.

Library/OcBootManagementLib/BootAudio.c
```diff
@@ -194,6 +194,14 @@ OcPlayAudioEntry (
...
+    } else {
+      OcPlayAudioFile (Context, OcVoiceOverAudioFileExternalTool, FALSE);
```
vit9696, line 203: Is this possible? In my opinion, ASSERT.

Library/OcMainLib/OpenCoreMisc.c
MikeBeaton: What I mean is, there are many paths (including basically always, in all real cases, as far as I can see, for tools in the Tools directory) where there is no pre-generated label and so one is generated on the fly instead, in this code: https://github.com/MikeBeaton/OpenCorePkg/blob/toggle-sip/Platform/OpenCanopy/Views/BootPicker.c#L1427-L1459 

Am not sure the (unused?) support for pre-generated labels is needed, in this case?

Platform/OpenCanopy/Views/BootPicker.c
MikeBeaton: Yes - this felt wrong - thanks

Library/OcBootManagementLib/VariableManagement.c
MikeBeaton: It's not just 0.
This is matching the behaviour of my Mac.
Oh.
It's because I'm on a beta version. Possibly.
But if it is we still need to support that.
There is definitely not nothing going on, here!

Library/OcBootManagementLib/VariableManagement.c
MikeBeaton: I do now know what it does - I started to look, but did not get to the bottom of it. I was just matching the behaviour of current macOS on my Mac (i.e. of csrutil). We may need to do some more tests.

Library/OcBootManagementLib/BootEntryManagement.c
MikeBeaton: ? I may be confused. The intention is that AddBootEntryFromSystem has not initialise the entry (and has probably printed some warning) if it returns a failure, and if it does the associated LoadOptions memory is then freed just below. I may not be understanding what you are seeing is wrong, yet.

Application/CsrUtil/CsrUtil.c
MikeBeaton: Oh - though I have not intended this for use outside shell. I was trying to make a thing for inside the shell that looked nicely quite like the Apl tool.

Application/CsrUtil/CsrUtil.c
MikeBeaton: ?

Application/CsrUtil/CsrUtil.c
MikeBeaton: Yes

Docs/Flavours.md
MikeBeaton: Good

Library/OcBootManagementLib/BootAudio.c
MikeBeaton: Yes

Library/OcMainLib/OpenCoreMisc.c
MikeBeaton: This is no longer called for internal tools, at least as I have changed it so far. That is what I mean about that other line, getting their label from the pre-loaded internal ones even if generic labels are not set, replacing this code.

Library/OcMainLib/OpenCoreMisc.c
```diff
@@ -259,37 +258,26 @@ OcToolDescribeEntry (
...
     *LabelDataSize = 0;
 
-    if (ChosenEntry->Type == OC_BOOT_RESET_NVRAM) {
-      Status = OcUnicodeSafeSPrint (
-        DescPath,
-        sizeof (DescPath),
-        OPEN_CORE_LABEL_PATH "ResetNVRAM.%a",
-        LabelScale == 2 ? "l2x" : "lbl"
-        );
-    } else {
```
MikeBeaton, line 281: To be clear, I'm really not sure this removed code made sense in the first place - it is loading the ResetNVRAM icon from the OpenCore LABELS path. But that is already in the pre-loaded system labels.

Application/CsrUtil/CsrUtil.c
vit9696: Easier to read.

Library/OcBootManagementLib/BootEntryManagement.c
vit9696: You branch from CsrActiveConfig, but it may have returned NOT_FOUND above.

Library/OcBootManagementLib/VariableManagement.c
vit9696: csrutil should not set Apple Internal. Something is wrong with your computer. You may have AppleInternal somewhere due to a bug in macOS updates, thus the bit. Setting Apple Internal bit will also lead to broken updates without Apple Connect, so please do not do this.

Library/OcBootManagementLib/VariableManagement.c
vit9696: We do not need to support this. Consumer builds should not have AppleInternal, this is a bug of your beta at most.

Library/OcMainLib/OpenCoreMisc.c
vit9696: Well, hard to say. Now that the font support is much better, maybe indeed drop that. Ok.

Include/Acidanthera/Library/OcBootManagementLib.h
MikeBeaton: Is okay to resolve this one and go with these labels? Esp. since you've already pinged Andrey and I guess he's currently producing audio corresponding to it this way - but just to double-check it's okay to go with this? Thx.

vit9696: This looks fine to me, but maybe Toggle SIP ON/ Toggle SIP OFF is more intuitive to be honest. Because a clickable entry is an action in the first place.

Library/OcBootManagementLib/VariableManagement.c
MikeBeaton: This is weird, It's running like this on different versions of the Big Sur beta, on my MBP and on my Dell hack.
I could leave in support for showing that when just this bit is set in csr-active-config, SIP is still considered enabled (in fact, I probably should, since that is definitely what csrutil does - you can dbl check this, on your setup, if you have a moment?), but then still not set it as our default for enabled and rather use 0 as it should be.

Library/OcBootManagementLib/BootEntryManagement.c
MikeBeaton: Carrying on on NOT_FOUND (and treating that as SIP enabled) is correct, I am almost certain. It is definitely 100% what the Mac does. If there is no such var it (treats it as zero, presumably) and just reports SIP a enabled. So that is what I am trying to do too, here and in lots of places.

Don't forget that with Apple's own CMD+OPT+P+R implementation, this is the state the NVRAM ends up in, so it is definitely not an unsupported corner case.

Library/OcBootManagementLib/VariableManagement.c
vit9696: OK. You can show, but you really must not set this bit at all.

Library/OcBootManagementLib/BootEntryManagement.c
vit9696: You did not understand my claim. I do agree that the logic is correct, but I do not like the fact the OUT parameter is initialised (overwritten) when the function fails. I think it is EDK-idiomatic not to change the param in such a case. Not a big deal, but a bit strange.

Include/Acidanthera/Library/OcBootManagementLib.h
MikeBeaton: I know it's an action, but the thing is, when the label says ON, or Enable, that is the first thing you see (and that would mean it is off!!). Basically a label which describes what effect the action will have necessarily describes the opposite of the current state - which is confusing when you actually use it. I do appreciate that it negates the logic of describing what the action will be, but I have tried both, and this way round is much easier in use.

Library/OcBootManagementLib/VariableManagement.c
MikeBeaton: Sure, will change.

Include/Acidanthera/Library/OcBootManagementLib.h
vit9696: Well, maybe @mhaeuser can say a word or two, I think I am ok with the suggested code.

Library/OcBootManagementLib/BootEntryManagement.c
MikeBeaton: Hang on, I'm getting closer to understanding - I think I stayed up too late last night - sorry! Specifically which param is getting initialised even if which function falls? I agree that shouldn't happen, I'm still not seeing where it's happening. Sorry!

Library/OcBootManagementLib/BootEntryInfo.c
MikeBeaton: I don't think so? This says 'if the boot entry is a system tool' (OC_BOOT_SYSTEM is an OR of all OC_BOOT_ bit values which are system tools) - should be okay? (Also, is not a direct replacement for what it replaces, that is split into two, above and just below.)

```
#define OC_BOOT_RESET_NVRAM         BIT8
#define OC_BOOT_TOGGLE_SIP          BIT9
#define OC_BOOT_SYSTEM              (OC_BOOT_RESET_NVRAM | OC_BOOT_TOGGLE_SIP)
```

MikeBeaton: (The assert is when the internal tool _name_ is invalid for the tool type.)

vit9696: I mean, can this function be called for such entries?

Library/OcBootManagementLib/BootEntryManagement.c
vit9696: Status == EFI_NOT_FOUND → OcIsSipEnabled (Status, CsrActiveConfig) is called. I do not think CsrActiveConfig is initialised here.

Docs/Sample.plist
```diff
@@ -1057,6 +1059,8 @@
 			<integer>17760515</integer>
 			<key>SecureBootModel</key>
 			<string>Default</string>
+			<key>ToggleSipMask</key>
```
vit9696, line 1062: To be honest, I do not believe there is any sense for this value to be in the config. It will get dropped in the future when this migrates to a separate driver anyway.

Library/OcBootManagementLib/BootEntryManagement.c
MikeBeaton: Okay, so, still not sure it is an error. Maybe I'm not using the idioms correctly.

Yes, OcGetSip does NOT initialise CsrActiveConfig on error (including EFI_NOT_FOUND), but the contract of OcIsSipEnabled is in fact to not try to read CsrActiveConfig if EFI_NOT_FOUND is passed in in Status; i.e. it can perfectly well accept an uninitialised param here in that case - in fact expects it to be uninitialised in that case!

So maybe it is alright? I'm not sure.

mhaeuser: EDK II is not consistent regarding this at all, should not matter.

MikeBeaton: Actually I can see that OcGetSip does initialise CsrActiveConfig, okay - then I need to move that out of there, and do it outside if I need it, right? Okay. Thanks.

Include/Acidanthera/Library/OcBootManagementLib.h
mhaeuser: "Toggle SIP ON/ Toggle SIP OFF" does not make it perfectly clear whether ON/OFF is the state or action in my opinion, I don't think "toggle on/off" is common language (rather "turn on/off"). I don't see the issue with "Enable/Disable SIP".

MikeBeaton: As I said, I actually used "Enable/Disable SIP" and I found it confusing to use. When it says "Enable SIP" it is telling you (what you probably want to know) that SIP is currently disabled, and vice versa. (Like I said, if you name it the action it will take, then you necessarily name it the opposite of the current state.) I am not a fan of changing to "ON/OFF" in the wording either.

I would be happy, lets say, to choose between "Enable SIP"/"Disable SIP", and "SIP: Disabled"/"SIP: Enabled". But in actual day-to-day use, I found that the one which shows the _state_ ("SIP: Disabled"/"SIP: Enabled") is much preferable to the one which shows what the next action will be ("Enable SIP"/"Disable SIP") (since that necessarily has wording which reflects the _opposite_ of what the current state is).

MikeBeaton: It obviously usable either way - it's just nicer (I believe, having used both) with the way which shows the state.

MikeBeaton: (FWIW - It gets even more confusing when you add in checking how it's working, as SIP is Enabled when it is zero and disabled when it is non-zero, so there is a profusion of double-negatives around here!)

mhaeuser: @MikeBeaton Well, this is a debugging utility, you should turn SIP off only shortly to complete your testing. OC is not really a dashboard where you monitor your system's state, this is an action like any other to me.

Library/OcBootManagementLib/BootEntryInfo.c
MikeBeaton: Yes, it always was. And it in fact returned values for them - the red stuff that I removed in `OcToolDescribeEntry`. This _is_ `CustomDescribe`. It's arguable whether it needs to be called for them now, but certainly on the current path this is legal. And made more sense when there was more in  `OcToolDescribeEntry` - but all recent changes are making more and more code drop out of that - possibly to the extent of it disappearing completely.

MikeBeaton: At which point you get a line like the above, saying 'nothing to do here'. But not illegal call.

Include/Acidanthera/Library/OcBootManagementLib.h
MikeBeaton: (You need SIP for some other things actually, like even setting flavours on macOS boot directories - though (fortunately) that's not really needed much with the auto-detection of macOS version.)

So I mean, as ever, if I'm implementing something at least partly because I'd like to use it, then I'd *rather* implement the version I'd actually like to use. As I said, the naming I've switched to (at least for now, after using the initially suggested ones initially) do make it more usable, to me.

But ofc I'm not going to not do it, if there's a strong feeling from you and Vit that I should not do the version I'd really prefer...! I gather there's quite a strong feeling from you, and so far a somewhat neutral feeling either way from @vit9696 . Hmmmmm....

MikeBeaton: An alternative is to maybe make some text feedback about the state, after selecting the action. But - again - I think it is quicker and simpler to use without that, i.e. it is basically pretty much quick, simple and intuitive to use as it is now...

vit9696: Maybe Toggle SIP (Enabled) / Toggle SIP (Disabled)?

MikeBeaton: Last comment - but - obviously with most boot entries, there's a specific label which says what it will do - which is always what it will do. (For this, this label would be "Toggle SIP".) But this one specifically needs to (or rather, does - as a convenience) let the user know what the current state is, in order to be sure what it is going to do. So it either implicitly tells the user the current state ("Enabled SIP" tells the user SIP is disabled, "Disable SIP" tells the user SIP is enabled), or it does it more explicitly. So there is some motivation for treating this one differently - in that it's already different in this way, more or less, whichever of the two options under discussion we use... hope that verbose-ness also makes some kind of sense!

MikeBeaton: Okay - thank you wise leader! Will do that.

Library/OcBootManagementLib/BootAudio.c
MikeBeaton: @Andrey1970AppleLife FYI The equivalent text labels finally read "Toggle SIP (Disabled)" and "Toggle SIP (Enabled)".

Docs/Sample.plist
MikeBeaton: Not convinced by this - in that if someone was using Catalina and using ToggleSip in OC, they would definitely prefer to set 0x77 here instead of 0x7F.

Application/CsrUtil/CsrUtil.c
MikeBeaton: @vit9696 - Do I need to make any changes here?

MikeBeaton: Actually, in particular, do I need to confirm this can be called usefully as an entry in `Tools` from OC as well? Have not done yet, but guess it makes sense?

Actually - also relevant - I commented out the link with OC console entry mode lib in the .inf file, because using it makes the screen flash and redraw every time the tool is called from UEFI Shell, which I guess is not so great if its main use case is as a shell tool. Just to double check with you that that decision and reason sounds right?

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -1396,6 +1396,15 @@ BootPickerEntriesSet (
       case OC_BOOT_RESET_NVRAM:
         Status = CopyLabel (&VolumeEntry->Label, &GuiContext->Labels[LABEL_RESET_NVRAM]);
         break;
+      case OC_BOOT_TOGGLE_SIP:
+        if (StrCmp (Entry->Name, OC_MENU_SIP_IS_DISABLED)) {
+          Status = CopyLabel (&VolumeEntry->Label, &GuiContext->Labels[LABEL_SIP_IS_ENABLED]);
+        } else if (StrCmp (Entry->Name, OC_MENU_SIP_IS_ENABLED)) {
+          Status = CopyLabel (&VolumeEntry->Label, &GuiContext->Labels[LABEL_SIP_IS_DISABLED]);
+        } else {
+          ASSERT (FALSE);
```
MikeBeaton, line 1405: @vit9696 - Still not entirely sure of the logic of these, should I insert e.g. `CpuDeadLoop` after this, or is okay as is?

Application/CsrUtil/CsrUtil.c
MikeBeaton: Okay - doing this now!

MikeBeaton: @vit9696 Okay made it work as tool or from command line.

Docs/Sample.plist
vit9696: It is not about conviction, this thing will not be supporred once SIP control moves to a separate driver, and thus there is no adequate need to add such an option for soon removal.

Application/CsrUtil/CsrUtil.c
```diff
@@ -0,0 +1,317 @@
...
+    PrintUsage ();
+    return Status;
+  }
+
+  if (OldArgv != Argv) {
+    //
+    // Cls if detect called as tool
+    //
+    gST->ConOut->ClearScreen (gST->ConOut);
+  }
```
MikeBeaton, line 164: I've hacked this a bit to work as a command line tool and when called from tools - I wasn't keen on just linking it to the OC console mode lib, since then - as a command line tool - it flashes the whole screen every time you use it.

But there might be a better way than this? Or this might be okay - it's all working fine, at least on my setup.

(As it is now, when used in Tools the entry needs to be set up in TextMode to see any output but ofc works fine (with no o/p shown, and one key press required) even if not in text mode.)

vit9696: In my opinion, no tool, be that run in shell or not should clear screen. It is what the user should do. Please remove.

Platform/OpenCanopy/Views/BootPicker.c
vit9696: There is an UNREACHABLE macro. You can use it.

Docs/Sample.plist
MikeBeaton: Okay - will remove - as I've re-done the tool (as should have been anyway) - Catalina users and the like could always set up a Tools entry.

Application/CsrUtil/CsrUtil.c
MikeBeaton: Hang on - just to be clear - when this is used from Tools without that, then you get a whole stream of whatever was in the console before, then with the latest line from this output at the end. It looks pretty bad. Not at Mac to take a screenshot. Is okay? (I mean, there is no way for the user to make the screen clear, when calling this from Tools, afaik? Using TextMode does not do that.)

Platform/OpenCanopy/Views/BootPicker.c
MikeBeaton: Okay - not so much worried about compiler - unless it warns - as about what would actually happen - where the instruction point would run to next. Not clear if UNREACHABLE stops the CPU from progressing - would imagine not - and not clear if ASSERT stops the CPU from progressing (e.g. in RELEASE build)? Guess I need to compile and run some tests.

MikeBeaton: Okay - thanks for pms clarifying! Will change as discussed.

Library/OcBootManagementLib/OcBootManagementLib.c
```diff
@@ -329,14 +329,11 @@ OcRunBootPicker (
         }
 
         //
-        // Do screen clearing for builtin menu here, so that it is possible to see the action.
-        // TODO: Probably remove that completely.
+        // Clear screen from previous console contents (builtin picker,
+        // or previous console tool even in Canopy) before loading the entry.
         //
+        gST->ConOut->ClearScreen (gST->ConOut);
```
MikeBeaton, line 335: I believe doing this always is correct here - it makes OpenCanopy behave as builtin picker always behaved.

Without this, then whenever you start a console tool, or anything in console, e.g. macOS verbose boot, then the first thing you see is the previous contents of the console scrolling onto the screen, before the chosen entry starts drawing its stuff. With this, it's fine

Library/OcBootManagementLib/OcBootManagementLib.c
```diff
@@ -329,14 +329,11 @@ OcRunBootPicker (
...
         if (Context->ShowMenu == OcShowSimpleBootMenu) {
-          //
-          // Clear screen from picker contents before loading the entry.
-          //
-          gST->ConOut->ClearScreen (gST->ConOut);
           gST->ConOut->TestString (gST->ConOut, OC_CONSOLE_MARK_UNCONTROLLED);
```
MikeBeaton, line 337: I am less certain that having this (here, and only when using builtin picker) is correct, but I think so?

MikeBeaton: Decision to leave for now, since that doesn't change the order of operations for builtin (which were optimised to avoid flashes etc.) and does appear to work fine on Canopy - at least on my system.

Library/OcBootManagementLib/BootAudio.c
Andrey1970AppleLife: Done

MikeBeaton: tyvm

1alessandro1:
@roddy20 I already did that, as mentioned here https://github.com/acidanthera/bugtracker/issues/1653

roddy20:
> @roddy20 I already did that, as mentioned here [acidanthera/bugtracker#1653](https://github.com/acidanthera/bugtracker/issues/1653)

I have both Toggle SIP and ResetNVRAM in OpenCanopy

1alessandro1:
Can you share your EFI without your serials?

1alessandro1:
@roddy20 @MikeBeaton I found the issue:

I had `AllowToggleSip` set to `Disabled` and `AllowNvramReset` set to `Enabled` and in this configuration the ResetNVRAM entry is not loaded. If I enable `AllowToggleSip` OpenCanopy will load the default `Tool.icns` icon (since I have no interest of adding a specific image for the SIP tool) and now both of them show up, `ResetNVRAM` and the AllowToggleSIP. 

To reproduce the bug one should enable only AllowNvramReset and disable `AllowToggleSip`, the Reset NVRAM does not appear anymore

PR #271, opened 2021-06-16
OcMainLib: Allow specifying custom kernels on ESP partition

PMheart:
closes https://github.com/acidanthera/bugtracker/issues/1602


Library/OcMainLib/OpenCoreKernel.c
```diff
@@ -1136,6 +1146,31 @@ OcKernelFileOpen (
...
+    if (mCustomKernelDirectory != NULL) {
+      DEBUG ((DEBUG_INFO, "OC: Redirecting %s to the custom one on ESP\n", FileName));
+      NewFileName = OcStrrChr (FileName, L'\\');
+      if (NewFileName == NULL) {
+        NewFileName = FileName;
+      }
+
+      DEBUG ((DEBUG_INFO, "OC: Filename after redirection: %s\n", NewFileName));
+
+      mCustomKernelDirectoryInProgress = TRUE;
```
vit9696, line 1161: This change is not correct, as currently there is no opened file, and we need to do fuzzy matching in the custom directory via OcKernelFuzzyMatch. As a result no opening should happen here, only directory name swapping.

vit9696:
```C
This = mCustomKernelDirectory;
FileName = NewFileName;
```

This is what should be enough.

Library/OcMainLib/OpenCoreKernel.c
```diff
@@ -1171,6 +1206,31 @@ OcKernelFileOpen (
...
+      Status = SafeFileOpen (mCustomKernelDirectory, &EspNewHandle, NewFileName, OpenMode, Attributes);
+      mCustomKernelDirectoryInProgress = FALSE;
+      if (!EFI_ERROR (Status)) {
+        if (*NewHandle != NULL) {
```
vit9696, line 1225: This `if` is unnecessary, as there was an `EFI_ERROR (Status)` check above with a `return`. `*NewHandle` is guaranteed to be set. Notably checking it for `NULL` here is not correct in the first place, as the handle may not be initialised on failure.

PR #272, opened 2021-06-18
add chinese

sumingyd:
add chinese


Andrey1970AppleLife:
CC @PMheart

stevezhengshiqi:
That link is not related to OpenCore discussion.

PMheart:
Sorry, but this link is not a specific discussion.

PR #275, opened 2021-06-29
Add SSDT-NVMe.dsl 

wy414012:
Add SSDT-NVMe.dsl to the working directory to solve the problem of external display of nvme disks.

### DIFF
--- a/Docs/AcpiSamples/Source/SSDT-NVMe.dsl
+++ b/Docs/AcpiSamples/Source/SSDT-NVMe.dsl
@@ -1,0 +1,59 @@
+/*
+ * Intel ACPI Component Architecture
+ * AML/ASL+ Disassembler version 20180427 (64-bit version)(RM)
+ * Copyright (c) 2000 - 2018 Intel Corporation
+ * 
+ * Disassembling to symbolic ASL+ operators
+ *
+ * Disassembly of SSDT-NVMe.aml, Fri Apr 16 23:34:27 2021
+ *
+ * Original Table Header:
+ *     Signature        "SSDT"
+ *     Length           0x000000C6 (198)
+ *     Revision         0x02
+ *     Checksum         0x29
+ *     OEM ID           "hack"
+ *     OEM Table ID     "nvme"
+ *     OEM Revision     0x00000000 (0)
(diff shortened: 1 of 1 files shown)


1alessandro1:
@wy414012 it's enough to add the DeviceProperty `built-in` DATA `01` to the pciroot path of the NVMe in question without the use of any SSDT

Docs/AcpiSamples/Source/SSDT-NVMe.dsl
```diff
@@ -0,0 +1,59 @@
...
+
+    Scope (\_SB.PCI0.NPE1)
+    {
+        Device (NVME)
+        {
+            Name (_ADR, Zero)  // _ADR: Address
+            Name (_SUN, One)  // _SUN: Slot User Number
+            Method (_INI, 0, NotSerialized)  // _INI: Initialize
+            {
+                \_SB.PCI0.NPE1.H000._ADR = 0x0F
```
vit9696, line 34: This is something custom. It is not needed.

Docs/AcpiSamples/Source/SSDT-NVMe.dsl
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    Return (Buffer (One)
+                    {
+                         0x03                                           
+                    })
+                }
+
+                Return (Package (0x02)
+                {
+                    "built-in",
```
vit9696, line 49: The value should be 00, but you are best to inject it with DeviceProperties. Only the ACPI node itself should be present in the table.

Andrey1970AppleLife: > Only the ACPI node itself should be present in the table.

The device is present always in ACPI, as NVME slot or PCIE slot.

Docs/AcpiSamples/Source/SSDT-NVMe.dsl
wy414012: The SSDT is submitted after verification of most models. Compared with config.plist, the middle parameter resolution should be a more convenient solution.

PR #278, opened 2021-07-16
Remove the repeated bless

Halo-Michael:


Halo-Michael:
Obviously bless was repeated in the command about personalizing macOS. Since it must be in macOS recovery mode at this time, also there is no need for sudo.

PR #282, opened 2021-08-03
Linux boot

mikebeaton:
Ready for any further review - still without additional support for system actions and associated sound protocol changes.

---

Boots Linux automatically, via OC, without chaining to Grub (in case anyone reading this is wondering "so what's new"?!).

**Tested booting fine:**

| Distro | OLB detection method |
| --- | --- |
| Fedora | blspec |
| Ubuntu | autodetect |
| Arch Linux | blspec |
| Linux Mint | autodetect |
| openSUSE | autodetect |
| Void Linux (musl version) | blspec |
| deepin | autodetect |
 
**Parsing boot files, but boot from default install would require EFI driver for Linux LVM:**

| Distro | OLB detection method |
| --- | --- |
| CentOS | blspec |
| Rocky Linux | blspec |

DONE:

 - [x] Parse and boot from `/loader/entries` (for BLSpec and systemd blscfg)
 - [x] Auto-detect vmlinuz-* and init* if on plausible root fs without `/loader/entries` dir
 - [x] Auto-detect Linux variant name and kernel version from kernel image file (cf. https://stackoverflow.com/a/11179559/795690 ), for use in auto-detect and when not present in blspec `.conf` file
   - [x] Above is working, but the info there isn't very authoritative (e.g. Linux Mint looks exactly like Ubuntu); to replace with better approach similar to/inspired by grub `os-prober`
 - [x] Make selectable as default boot entry
   - [x] As above for no-menu quick code path 
 - [x] Allow parameterise OC drivers for optional argument passing
 - [x] Various minor cleanups
 - [x] Documentation

TO ADD:

 - [ ] Add support for `.contentDetails` and `.contentFlavour` files, to set entry base name and entry flavour, where auto-detection sets them wrong or finds no value. Currently auto-detection is working well enough that this should not be required except in edge cases.

TO FIX:

 - [ ] Prevent over-long Linux boot entry names messing up layout of builtin menu (use ellipses when required)

Also requested/planned:

 - [ ] In order to allow entry protocol drivers to provide own audio-assist, update OC sound protocol to work from string rather than int sound ids
 - [ ] Upgrade boot entry protocol to support system actions
 - [ ] Redo existing system actions (Reset NVRAM, Toggle SIP) as entry protocol drivers
 - [ ] Use new FlexArrayLib in place of similar code in OcXmlLib
 - [ ] Use new Arg parsing code in place of existing less flashy arg parsing code


mikebeaton:
If anybody watching has some time to download the [build artefacts](https://github.com/acidanthera/OpenCorePkg/suites/3544600345/artifacts/84903724) for this version and try it out, that would be great. Instructions are in the attached `Configuration.pdf`, and forward search for OpenLinuxBoot.

Note that you need to [update your config file `Drivers` section](https://github.com/acidanthera/OpenCorePkg/pull/282/files#diff-bb8ee4e0e33402436f51481de220d939a7303bd2c3fd3b5e72d6443865911584) as this now supports passing parameters to drivers (used here, and intended to be used in future in a couple more). You could keep your current `Drivers` section unchanged, but renamed to `<key>#Drivers</key>` (keys starting with # are ignored by OC), and add the new layout under a new `<key>Drivers</key>`, so that it is easy to switch between them by just swapping the #. `ocvalidate` in the download package is up-to-date, to allow double-checking this change.

You also need to download and add to your OC drivers a UEFI EXT4 driver (such as the `ext4_x64.efi` driver found within the [current rEFInd download](https://sourceforge.net/projects/refind/postdownload)), as OC does not - and is not planning to - include this.

If you already have Linux entries in OC, no need to remove them, but you might want to use something like:

```
			<dict>
				<key>Driver</key>
				<string>OpenLinuxBoot.efi</string>
				<key>Enabled</key>
				<false/>
				<key>Params</key>
				<string>flags=0x8387</string>
			</dict>
```

where the flags mean scan everything (which is the default anyway), and then add some debugging text to boot entry names created by this driver, which is not added by default and will help identifying entries added by this new system.

It is fully working for me on latest Fedora and Ubuntu, and tested in principle (i.e. with the relevant config files, but not end to end yet) on a couple more, including CentOS. So any feedback as to how it works (and with which distro) would be gratefully received.

nadiaholmquist:
Just tested it with Arch Linux, from a normally `systemd-boot`-based setup. OpenCore successfully boots my install and it seems to work as it should, nice job :)

```
nhp@narshe /boot % ls
EFI                                      intel-ucode.img
initramfs-linux-ck-skylake-fallback.img  loader
initramfs-linux-ck-skylake.img           vmlinuz-linux-ck-skylake
nhp@narshe /boot % cat loader/entries/arch.conf 
title   Arch Linux
linux   /vmlinuz-linux-ck-skylake
initrd  /initramfs-linux-ck-skylake.img
initrd  /intel-ucode.img
options root=UUID=a4607b9a-f0d8-49e9-a077-a750c41852bd sysrq_always_enabled=1 quiet splash loglevel=3 rd.systemd.show_status=auto rd.udev.log_priority=3 acpi_enforce_resources=lax vga=current intel_iommu=on iommu=pt
```

`bootctl` reports that none of the features that would be supported by `systemd-boot` are supported, is this to be expected?
```
Current Boot Loader:
      Product: n/a
     Features: ✗ Boot counting
               ✗ Menu timeout control
               ✗ One-shot menu timeout control
               ✗ Default entry control
               ✗ One-shot entry control
               ✗ Support for XBOOTLDR partition
               ✗ Support for passing random seed to OS
               ✗ Boot loader sets ESP information
          ESP: n/a
         File: └─n/a

Random Seed:
 Passed to OS: no
```

mikebeaton:
@nadiaholmquist - thank you.

It would be nice to see the result of `cat /proc/cmdline` from inside your install, just to be sure that everything looks as it should! :)

We are not currently attempting to pass bootloader capabilities to anything - I could look into attempting to provide that info; though of course not much except Arch Linux uses systemd-boot directly, as I understand it - so it might be some work for little gain to most people - unless you (or anyone else wanting to comment) believes otherwise? I'm marginally surprised that your install has understood that it was not booted by systemd-boot as normal, in terms of what capabilities it reports - though I'm sure when I look into how this works, it'll make sense. (I'm assuming you didn't change that, and can still boot via systemd-boot if you wish. What does it report if you do that?)

Thanks again.

nadiaholmquist:
`/proc/cmdline` are identical between OpenCore and systemd-boot: `initrd=/initramfs-linux-ck-skylake.img initrd=/intel-ucode.img root=UUID=a4607b9a-f0d8-49e9-a077-a750c41852bd sysrq_always_enabled=1 quiet splash loglevel=3 rd.systemd.show_status=auto rd.udev.log_priority=3 acpi_enforce_resources=lax vga=current intel_iommu=on iommu=pt`

And yeah you are probably right about a minority of distros using systemd-boot, I know Fedora uses some pretty cursed setup with GRUB while somehow still implementing the boot loader spec, idk about Debian/Ubuntu/etc. Being able to support the `--boot-loader-entry` and `--boot-loader-menu` options when restarting the system with `systemctl` could be nice, though if those also require systemd-boot I figure it wouldn't be worth the effort.

mikebeaton:
@nadiaholmquist I had more of a look at the [systemd Boot Loader Interface](https://systemd.io/BOOT_LOADER_INTERFACE/) and yes, it is quite `systemd-boot` specific. Adding these features wouldn't just be adding more code to this new OC plugin I'm working on, OC itself would need code added for this (possibly as a further extension of the new Boot Entry Protocol, but it would still be specifically for this). So again it would come down to usage. Might even make sense as a separate new plugin. But might need a volunteer to write it! And also depends importantly on an overall decision as to whether it really is worth supporting.

For anyone else reading, this is nothing to do with whether you can boot Linux with this new system. You can. it's to do with whether you can issue additional commands from within Linux, to tell the bootloader what to do on next boot.

mhaeuser:
Is there any distro but Arch that is pushing systemd-boot adoption? I don't think I saw another distro, especially the big ones, that intend to support this garbage loader. I don't think Arch usage can be endorsed looking at the trust situation (AUR and friends), so I'd vote to keep anything to support (realistically) Arch-specific things, most especially systemd-boot, to third-party code. Default boot option should be handled via BootNext (so macOS and Windows can also be supported), for the boot menu option we could re-use the firmware setup functionality of UEFI. Linux should have tools to support those general functionalities over systemd-boot-specific mechanisms.

nadiaholmquist:
> @nadiaholmquist I had more of a look at the systemd Boot Loader Interface and yes, it is quite systemd-boot specific. Adding these features wouldn't just be adding more code to this new OC plugin I'm working on, OC itself would need code added for this (possibly as a further extension of the new Boot Entry Protocol, but it would still be specifically for this). So again it would come down to usage. Might even make sense as a separate new plugin. But might need a volunteer to write it! And also depends importantly on an overall decision as to whether it really is worth supporting.

Ah, I'd have thought those features were part of the boot loader spec but I guess leave it to systemd to invent its own thing for no clear reason... At this pint doesn't seem line it makes sense to support to me either.

mikebeaton:
> Is there any distro but Arch that is pushing systemd-boot adoption? I don't think I saw another distro, especially the big ones, that intend to support this garbage loader. I don't think Arch usage can be endorsed looking at the trust situation (AUR and friends), so I'd vote to keep anything to support (realistically) Arch-specific things, most especially systemd-boot, to third-party code. Default boot option should be handled via BootNext (so macOS and Windows can also be supported), for the boot menu option we could re-use the firmware setup functionality of UEFI. Linux should have tools to support those general functionalities over systemd-boot-specific mechanisms.

Yes, ofc we already support actual (used by many people) standards such as the BootNext var - this is already fully integrated into OC and works in OpenLinuxBoot automatically. (And yes, can already be set using `efibootmgr`.)

Boot Loader Interface uses its own, different, NVRAM namespace, with different vars.

mikebeaton:
i.e. While we indeed already have the majority of those features, just setting the NVRAM var which would put check marks next to them would be false, as we do not support them via the not-really-standard Boot Loader Interface variables

mikebeaton:
Added known booting list in OP. It seems like the basic approach is sound.

I'd still be very happy to receive reports from anyone else trying it out. Even if it worked! ;-) Thanks to @nadiaholmquist so far.

Include/Acidanthera/Library/OcMiscLib.h
```diff
@@ -263,4 +263,332 @@ OcConsoleFlush (
...
+typedef struct OC_STRING_BUFFER_ OC_STRING_BUFFER;
+
+/**
+  Initialize string buffer.
+
+  @retval Non-null                  Buffer was created.
+  @retval NULL                      Out of memory.
+**/
+OC_STRING_BUFFER *
+OcAsciiStringBufferInit (
```
vit9696, line 498: Why is not it using dynamic buffer beneath itself?

Include/Acidanthera/Protocol/OcBootEntry.h
```diff
@@ -0,0 +1,140 @@
...
+#define OC_BOOT_ENTRY_PROTOCOL_GUID                                                 \
+  {                                                                                 \
+    0x8604716E, 0xADD4, 0x45B4, { 0x84, 0x95, 0x08, 0xE3, 0x6D, 0x49, 0x7F, 0x4F }  \
+  }
+
+/*
+  OC_BOOT_ENTRY_PROTOCOL revision.
+  Needs to be incremented when OC_PICKER_ENTRY structure is updated.
+*/
+#define OC_BOOT_ENTRY_PROTOCOL_REVISION 0
```
vit9696, line 28: Better start with 1.

Include/Acidanthera/Protocol/OcBootEntry.h
```diff
@@ -0,0 +1,140 @@
...
+  Free list of OpenCore boot entries from previous call to OC_GET_BOOT_ENTRIES.
+
+  @param[in]  BootEntries       List of boot entries, as returned by previous call.
+  @param[in]  NumEntries        The number of entries, as returned by previous call.
+**/
+typedef
+VOID
+(EFIAPI *OC_FREE_BOOT_ENTRIES) (
+  IN   OC_PICKER_ENTRY          **Entries,
+  IN   UINTN                    *NumEntries
```
vit9696, line 126: Why is this a pointer?

Library/OcBootManagementLib/BootArguments.c
```diff
@@ -21,10 +12,33 @@
...
+*/
+typedef enum LOAD_OPTIONS_PARSE_STATE_ {
+  LOAD_OPTIONS_WHITE_SPACE,
+  LOAD_OPTIONS_NAME,
+  LOAD_OPTIONS_VALUE_FIRST,
+  LOAD_OPTIONS_VALUE,
+  LOAD_OPTIONS_QUOTED_VALUE
+} LOAD_OPTIONS_PARSE_STATE;
+
+#define IS_SPACE(c) ((c) == L' ' || (c) == L'\t' || (c) == L'\n')
```
vit9696, line 31: Better to put to OcStringLib and perhaps make a function.

Library/OcBootManagementLib/BootArguments.c
```diff
@@ -360,3 +374,226 @@ OcCheckArgumentFromEnv (
...
+EFI_STATUS
+OcParseLoadOptions (
+  IN     CONST EFI_LOADED_IMAGE_PROTOCOL  *LoadedImage,
+     OUT       OC_DYNAMIC_ARRAY           **ParsedLoadOptions
+  )
+{
+  CHAR16                      *Pos;
+  LOAD_OPTIONS_PARSE_STATE    State;
+  BOOLEAN                     Retake;
+  CHAR16                      c;
```
vit9696, line 387: I believe the EDK II convention is Ch.

Library/OcBootManagementLib/BootArguments.c
```diff
@@ -360,3 +374,226 @@ OcCheckArgumentFromEnv (
...
+  CHAR16                      *Name;
+  CHAR16                      *Value;
+  OC_LOAD_OPTION              *Option;
+
+  ASSERT (LoadedImage != NULL);
+  ASSERT (ParsedLoadOptions != NULL);
+  *ParsedLoadOptions = NULL;
+
+  if (LoadedImage->LoadOptionsSize > MAX_LOAD_OPTIONS_SIZE) {
```
vit9696, line 396: This lacks evenness checking. If LoadOptionsSize is not divisible by sizeof(CHAR16), it is not valid.

Library/OcBootManagementLib/BootArguments.c
```diff
@@ -360,3 +374,226 @@ OcCheckArgumentFromEnv (
 
   return HasArgument;
 }
+
+EFI_STATUS
+OcParseLoadOptions (
```
vit9696, line 379: Shall we merge this with XNU load options parsing? See OcGetArgumentFromCmd and related.

Library/OcBootManagementLib/BootArguments.c
```diff
@@ -360,3 +374,226 @@ OcCheckArgumentFromEnv (
...
+  ASSERT (Name != NULL);
+  ASSERT (Value != NULL);
+
+  if (ParsedLoadOptions == NULL) {
+    return FALSE;
+  }
+
+  ASSERT (ParsedLoadOptions->Items != NULL);
+
+  for (Index = 0; Index < ParsedLoadOptions->Count; Index++) {
```
vit9696, line 526: I believe the convention is prefix increment in EDK II.

Library/OcBootManagementLib/BootArguments.c
```diff
@@ -360,3 +374,226 @@ OcCheckArgumentFromEnv (
...
+
+BOOLEAN
+OcHasLoadOption (
+  IN     CONST OC_DYNAMIC_ARRAY   *ParsedLoadOptions,
+  IN     CONST CHAR16             *Name
+  )
+{
+  CHAR16      *Value;
+
+  return OcLoadOptionGetStr (ParsedLoadOptions, Name, &Value);
```
vit9696, line 548: Use explicit != NULL checking.

Library/OcBootManagementLib/BootEntryManagement.c
```diff
@@ -1905,12 +2020,35 @@ OcScanForBootEntries (
...
+      );
+
     //
     // Record predefined recoveries.
     //
     AddBootEntryFromSelfRecovery (BootContext, FileSystem);
   }
 
+  if (DefaultEntryProtocolId != NULL) {
+    FreePool ((VOID *) DefaultEntryProtocolId); // <- discards const
```
vit9696, line 2048: Why is this const in the first place?

vit9696 (commented):
Reviewed partially till BootEntryManagement.c

Docs/Configuration.tex
```diff
@@ -6154,6 +6158,124 @@ \subsection{OpenRuntime}\label{uefiruntime}
...
+	Available flags are: \medskip
+      
+  \begin{itemize}
+  \tightlist
+    \item \texttt{0x00000001} (bit \texttt{0}) --- \texttt{LINUX\_BOOT\_SCAN\_ESP},
+    Allows scanning for entries on EFI System Partition.
+    \item \texttt{0x00000002} (bit \texttt{1}) --- \texttt{LINUX\_BOOT\_SCAN\_XBOOTLDR},
+    Allows scanning for entries on Extended Bootloader partition.
+	  \item \texttt{0x00000004} (bit \texttt{2}) --- \texttt{LINUX\_BOOT\_SCAN\_EXT4},
+	  Allows scanning for entries on EXT4, and on other file systems (e.g. BTRFS)
```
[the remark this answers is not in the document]
mikebeaton: ScanPolicy is done by file system type GUID, so I believe cannot distinguish EXT4 and BTRFS as it currently works, since both are marked as Linux filesystem `mLinuxFileSystemPartitionTypeGuid`.

Include/Acidanthera/Library/OcBootManagementLib.h
```diff
@@ -360,6 +386,11 @@ typedef struct OC_BOOT_CONTEXT_ {
 **/
 #define OC_SCAN_ALLOW_FS_EXT             BIT12
 
+/**
+  Allow scanning XBOOTLDR filesystems.
+**/
+#define OC_SCAN_ALLOW_FS_XBOOTLDR        BIT13
```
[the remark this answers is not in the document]
mikebeaton: Yes, it is BootLoaderSpec specific, i.e. it might barely be used anywhere by anyone. It means that a partition other then the ESP is marked (by its type GUID) as the partition on which to find (all) boot info. I believe it is intended to be a larger-sized FAT partition, in order to fit all the copies of all the kernel files there.

(It is not the same as the systemd approach used by Fedora, where the kernel files more sensibly stay on Linux's own boot partition.)

Include/Acidanthera/Library/OcMiscLib.h
```diff
@@ -263,4 +263,332 @@ OcConsoleFlush (
...
+/**
+  Utility method to free memory pointed to by dynamic array
+  item when the item is a single pointer (or when the only
+  allocated memory is pointed to by a pointer which is the
+  first element).
+
+  @param[in]      Item          Dynamic array item to free.
+**/
+VOID
+OcDynamicArrayFreePointerItem (
```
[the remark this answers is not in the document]
mikebeaton: Not entirely sure if it is worth using in OcXmlLib, since it adds 8 bytes overhead per XML node, as well as (I suppose) more work and the risk of introducing bugs unnecessarily for possibly not enough gain?

Perhaps could just add comment there that these might be combined?

Include/Acidanthera/Protocol/OcBootEntry.h
mikebeaton: Yes, was kind of leaving it at 0 until ready to merge. But will probably forget, will up it now. ;-)

Library/OcBootManagementLib/BootArguments.c
mikebeaton: I guess I didn't want to tightly couple different programs which might have slightly different requirements for what counts as white space - but can do...

Library/OcBootManagementLib/BootArguments.c
mikebeaton: Can do... again this would add some alloc and code overhead to things which currently use OcGetArgumentFromCmd...

Library/OcBootManagementLib/BootArguments.c
mikebeaton: The intention was the NULL checking was already done inside the inner call - should be outside as well?

Library/OcBootManagementLib/BootEntryManagement.c
mikebeaton: Changed.

Include/Acidanthera/Library/OcMiscLib.h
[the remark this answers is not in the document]
mikebeaton: Naming: OcVectorLib sounds nice, but I believe misleadingly implies that it has something to do with vector operations (dot product, cross product, SIMD, interaction with matrices, or some such).

Other possible names:
 - DynamicArray
 - DynamicBuffer
 - ?

Docs/Configuration.tex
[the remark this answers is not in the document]
vit9696: Maybe make it clear then that it accepts both?

Include/Acidanthera/Library/OcBootManagementLib.h
```diff
@@ -473,14 +504,23 @@ EFI_STATUS
...
+  Note that OpenLinuxBoot OC_BOOT_ENTRY_PROTOCOL_REVISION needs incrementing
+  when this structure is updated.
 **/
 typedef struct {
+  //
+  // Used by OC_BOOT_ENTRY_PROTOCOL to reidentify entry.
+  // Multiple entries may share an id - allows e.g. newest version
+  // of Linux install to automatically become selected default.
+  //
+  CONST CHAR8  *Id;
```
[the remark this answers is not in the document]
vit9696: OK

Include/Acidanthera/Library/OcMiscLib.h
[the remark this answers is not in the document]
vit9696: I meant that OcXmlLib could use this one, not vice versa. OcDynamicArraryLib feels ok, can be also OcPoolArrayLib, OcDynArrayLib, or OcFlexArrayLib. I think I like the last one.

Include/Acidanthera/Library/OcMiscLib.h
vit9696: Right, ok.

Library/OcBootManagementLib/BootArguments.c
vit9696: Hmm, better to follow ctype here really. It matches a few more symbols, but I believe we can live with it.

Library/OcBootManagementLib/BootArguments.c
vit9696: Well, maybe we can refactor the existing functions, but I'd not try to overoptimise without a really big need.

Library/OcBootManagementLib/BootArguments.c
vit9696: No, I mean return value checking. Do not rely on NULL being implicitly FALSE.

mikebeaton: Ah, OcLoadOptionGetStr returns BOOLEAN.

mikebeaton: Because it is possible for an option to be present, with a NULL string value.

vit9696: Ah, never mind. Thanks.

Include/Acidanthera/Protocol/OcBootEntry.h
mikebeaton: Why is `*NumEntries` a pointer, yes? It zeros the value on call. I guess cannot enforce that all implementations of Boot Entry Protocol do that nicely... Equally `**Entries` is pointer to pointer in order to zero that too. Overkill/bad idea?

mhaeuser: I have not seen that before, not even in other codebases. For `NumEntries` I think it's unintuitive, it's a bad example, but what if you want to log how many entries you just destructed? It may make sense to preserve it. I makes no sense to preserve the pointer, but also it's unconventional I guess?

Library/OcBootManagementLib/BootArguments.c
mikebeaton: Ah. This might be a good reason why not use the new functions, though: https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcBootManagementLib/BootArguments.c#L321-L323 - new stuff is 100% destructive/assumes ownership of options area.

Maybe rename existing method to OcSafeGetArgumentsFromCmd and add comment that it is non-desructive?

Include/Acidanthera/Library/OcMiscLib.h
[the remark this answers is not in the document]
mikebeaton: 💪 lol!

I still wasn't sure whether it was worth digging into and changing insides of OcXmlLib, for reasons stated, but can do...

Library/OcBootManagementLib/BootArguments.c
vit9696: Well, my primary point is to use common parsing code. APIs can vary.

Include/Acidanthera/Protocol/OcBootEntry.h
mikebeaton: There's definitely a few places around OC where **ptr is passed in for freeing pointer memory - so leave **ptr but just change to NumEntries? Or change both? Definitely can do - perhaps makes more sense not to use ** at something which is more of an interface than some method internal to OC.

Library/OcBootManagementLib/BootArguments.c
mikebeaton: My parsing code is destructive (shifts memory to handle `'\'` escaped characters and inserts `'\0'` for ends of names and values). I believe cannot be used as is on options which are to be passed on to someone else. However, could be used if we make of copy of the options (which it seems we do anyway, at around the point I linked to).

Docs/Configuration.tex
```diff
@@ -6154,6 +6158,124 @@ \subsection{OpenRuntime}\label{uefiruntime}
...
+In both cases, the method of starting the kernel relies on it being compiled with
+EFISTUB, however this applies to almost all modern distros, particularly those which use systemd.
+
+The latest kernel version of a given install is always shown. Additional versions, recovery versions,
+etc. are added as auxiliary boot entries, so depending on OpenCore's \texttt{HideAuxiliary} setting
+may not be shown until the space key is pressed.
+
+By default most Linux distributions keep their boot files on an EXT4 file system (even
+if the main system partition is something else such as BTRFS), therefore a suitable UEFI EXT4 file system
+driver is required. Amongst other sources, a suitable driver \texttt{ext4\_x64.efi} can be
```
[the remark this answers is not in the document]
mikebeaton: Looking now...

Include/Acidanthera/Protocol/OcBootEntry.h
mikebeaton: Updated in latest push to PR

Library/OcBootManagementLib/BootArguments.c
mikebeaton: Have added to to-do list in main PR post - could be moved into an issue and done after merge?

Include/Acidanthera/Library/OcMiscLib.h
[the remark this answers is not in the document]
mikebeaton: Have added to to-do list in main PR post - could be moved into an issue and done after merge?

Docs/Configuration.tex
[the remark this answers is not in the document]
mikebeaton: Reported back by PM - but it's not currently looking too promising... builds, loads and runs, but immediately halts with (seemingly?) incorrect errors.

TaiPhamD:
I tried it on my Ubuntu install and OpenCore was able to auto detect both kernel images from Ubuntu.

When I select the Linux boot entry I get a black screen and nothing boots after waiting for about 5 minutes. Pressing the power button will instantly turn off the PC. 

The Ubuntu grub boot loader exists on a separate ESP than OpenCore. Booting to grub boot loader directly works.

I get this log from OC debug build:

```
13:252 00:001 OCB: Should boot from 3. Ubuntu 20.04.3 LTS (5.11.0-34-generic) (T:64|F:0|G:0|E:0|DEF:0)
13:265 00:013 OCB: Perform boot Ubuntu 20.04.3 LTS (5.11.0-34-generic) to dp PciRoot(0x1)/Pci(0x0,0x0)/Pci(0x0,0x0)/NVMe(0x1,3C-07-48-44-8B-44-1B-00)/HD(2,GPT,45D659E4-BF85-4E50-993B-5A91F4875E18,0x100800,0x3A285000)/\boot\vmlinuz-5.11.0-34-generic (0/0)
13:278 00:013 OCSB: No IMG4 found - Not Found
13:280 00:001 OCB: Arch filtering 41B01018(10132256)->41B01018(10132256) caps 4 - Success
13:367 00:086 OCB: Matching <> args on type 64 43C1E018
13:369 00:001 OCB: Custom (64) DeviceHandle 475E1C18 FilePath 43C1E198
```

Not sure what other logs would be useful here to diagnose.

mikebeaton:
@TaiPhamD The full log would be useful to diagnose (as contains other LNX lines about what it encountered as it detected your distro). Also the output of `mount` from your successfully booted distro is useful. Thx.

TaiPhamD:
@mikebeaton Here it is. Let me know if you want me to get any other logs. Thank you for creating this PR!

[opencore-2021-09-08-080114.txt](https://github.com/acidanthera/OpenCorePkg/files/7129561/opencore-2021-09-08-080114.txt)
[ubuntu_mount.txt](https://github.com/acidanthera/OpenCorePkg/files/7129563/ubuntu_mount.txt)
[ubuntu_fdisk.txt](https://github.com/acidanthera/OpenCorePkg/files/7129562/ubuntu_fdisk.txt)

mikebeaton:
Hi @TaiPhamD - Thanks for the info so far. Bear with me, I will produce a build with some additional Linux-related debug output in it, hopefully tomorrow.

mikebeaton:
Hi - looking again now and have realised you've provided the log with a debug version of OpenCore but still with the release version of OpenLinuxBoot.efi - please run again using debug versions of both.

TaiPhamD:
My bad I copied debug files but missed the most important one. 

[opencore-2021-09-10-162731.txt](https://github.com/acidanthera/OpenCorePkg/files/7145278/opencore-2021-09-10-162731.txt)


```
✘  /Volumes/EFI  rg LNX opencore-2021-09-10-162731.txt
547:07:640 00:008 LNX: TypeGUID: C12A7328-F81F-11D2-BA4B-00A0C93EC93B (ESP) PARTUUID: 9C7CCD10-86F3-47B3-A2D5-3E32BA65C0E0
548:07:648 00:008 LNX: Nothing found
555:07:667 00:001 LNX: TypeGUID: EBD0A0A2-B9E5-4433-87C0-68B6B72699C7 (NTFS/FAT) PARTUUID: 5300C38E-331D-4DD2-A763-3DCE82A9D927
556:07:669 00:001 LNX: Nothing found
563:07:693 00:003 LNX: TypeGUID: C12A7328-F81F-11D2-BA4B-00A0C93EC93B (ESP) PARTUUID: E878C4E6-004E-4E42-BE19-2E1027E45ECD
564:07:697 00:003 LNX: Nothing found
575:07:719 00:001 LNX: TypeGUID: C12A7328-F81F-11D2-BA4B-00A0C93EC93B (ESP) PARTUUID: A812C758-EA83-431F-A484-A030AEFD6492
576:07:720 00:001 LNX: Nothing found
577:07:723 00:002 LNX: TypeGUID: C12A7328-F81F-11D2-BA4B-00A0C93EC93B (ESP) PARTUUID: 3DF99478-B7C8-4C1B-B16C-D994132DDD6A
578:07:725 00:002 LNX: Does not appear to be root filesystem - Not Found
579:07:728 00:002 LNX: Nothing found
586:07:740 00:001 LNX: TypeGUID: EBD0A0A2-B9E5-4433-87C0-68B6B72699C7 (NTFS/FAT) PARTUUID: 1F5AB430-2474-48FD-9387-357B13234735
587:07:742 00:001 LNX: Does not appear to be root filesystem - Not Found
588:07:744 00:002 LNX: Nothing found
602:07:770 00:002 LNX: TypeGUID: C12A7328-F81F-11D2-BA4B-00A0C93EC93B (ESP) PARTUUID: 33002EF2-526D-4664-A7CD-42E9AC95A835
603:07:773 00:002 LNX: Nothing found
610:07:786 00:001 LNX: TypeGUID: 0FC63DAF-8483-4772-8E79-3D69D8477DE4 (DATA) PARTUUID: 45D659E4-BF85-4E50-993B-5A91F4875E18
611:07:797 00:011 LNX: Reading \etc\os-release
612:07:799 00:001 LNX: Found distro Ubuntu 20.04.3 LTS
613:07:803 00:004 LNX: Reading \etc\default\grub
618:07:812 00:002 LNX: APFS - not scanning
622:07:819 00:001 LNX: APFS - not scanning
626:07:826 00:001 LNX: APFS - not scanning
630:07:833 00:001 LNX: APFS - not scanning
```

mikebeaton:
Thanks for the required additional info. I've still no idea why this isn't booting, it all looks right. Can you send the output of `cat /proc/cmdline` from inside your distro when it's booted normally, please?

mikebeaton:
Actually, and also attach `/etc/default/grub`, if you wouldn't mind. Thanks.

TaiPhamD:
Here it is: 
[default_grub.txt](https://github.com/acidanthera/OpenCorePkg/files/7146220/default_grub.txt)
[proc_cmdline.txt](https://github.com/acidanthera/OpenCorePkg/files/7146221/proc_cmdline.txt)

mikebeaton:
Please could you try again with the latest build https://github.com/acidanthera/OpenCorePkg/suites/3739572732/artifacts/91687026 . Please add `flags=0xffff` to the Arguments of your OpenLinuxBoot.efi driver before running the test. This is only to collect more info, I am not expecting it to boot yet, unfortunately.

I am now wondering if it is actually booting Linux, but stuck in a graphics mode where nothing displays - which may be a hardware-specific issue which we need to look at. (Interaction of OC with your hardware, I know it can boot via GRUB.)

I need a little bit of info from the above log then I can give you a test of that theory (the idea will be to manually create an Entries entry for your Ubuntu and see how that works).

TaiPhamD:
Yes I am using an AMD 6900xt so could be something different about this gpu. I do have another AMD rx5500 that I could test to see if it makes a difference.

Here it is with the artifact 0.7.4 debug build:

[opencore-2021-09-11-005048.txt](https://github.com/acidanthera/OpenCorePkg/files/7146981/opencore-2021-09-11-005048.txt)

mikebeaton:
Thanks, as said can you do same test with the new debug build, but please add `flags=0xffff` as plain text, in your config.plist, within the Arguments section of your OpenLinuxBoot driver. Thanks.

mikebeaton:
You need to make an custom entry in the config.plist Entries section, with the Path set equal to the whole path after "dp" in the log line "OCB: Should boot from... dp ...." and the Arguments set equal to the value inside &lt;&gt; (but not including &lt;&gt;) in the log line "OCB: Args".

My guess, if you can get this to work and try to boot, is that it may well show the same behaviour as now with TextMode false, and may well boot into Ubuntu with TextMode true.

Can you try this?

TaiPhamD:
> You need to make a custom entry in the config.plist Entries section, with the Path set equal to the whole path after "dp" in the log line "OCB: Should boot from... dp ...." and the Arguments set equal to the value inside <> (but not including <>) in the log line "OCB: Args".
> 
> My guess, if you can get this to work and try to boot, is that it may well show the same behaviour as now with TextMode false, and may well boot into Ubuntu with TextMode true.
> 
> Can you try this?

Yes I can try it. FYI the previous 0.7.4 flags I had already updated the flags to 0xffff

mikebeaton:
Hmm it doesn't appear to be showing the additional info? Anyway above quoted test is what we need next.

TaiPhamD:
[opencore-2021-09-11-080620.txt](https://github.com/acidanthera/OpenCorePkg/files/7147439/opencore-2021-09-11-080620.txt)
Screen Shot 2021-09-11 at 1 04 47 AM

same behavior even with the custom manual entry.

mikebeaton:
Did you try setting `true` under `TextMode` in the natural entry?

TaiPhamD:
- I definitely have tried setting TextMode to true which also didn’t work . (I  made sure to choose my custom entry -I mistakenly choose the auto scanned entry the first time reboot)

- I have tried changing Picker mode to Builtin to get rid of OpenCanopy which didn’t work either .

- I tried changing to using UUID instead of PARUUID didn’t work either

- Tried changing gpu that didn’t work either 

I think I want to try to add manual bless override for the custom entry

mikebeaton:
Well the only thing we haven't successfully tried is `flags=0xffff`. When that is working then each entry generated by OpenLinuxBoot.efi has a little bit of extra debug info added to its title, and we definitely didn't see that.

If you can get that working, then maybe just run the log again, so I can see what OLB reports, but start the custom entry set to `root=UUID=...`  and `TextMode=true`.

Thanks for testing these many and various combinations.

mikebeaton:
Actually, new question, does Ubuntu start if you chainload GRUB - if you use OC to start GRUB, then GRUB to start Ubuntu? Or only when you start GRUB directly?

TaiPhamD:
Chainloading using method A from: https://github.com/dortania/OpenCore-Multiboot/blob/master/oc/linux.md used to work for me but now has the same behavior. 

I also have to add \EFI\ubuntu\grubx64.efi in BlessOverride entry or else I get the bless error.

mikebeaton:
I would guess that chainloading will work again if you do not load ext4_x64.efi driver (since grub is also going to try to load it's own ext4 driver). Can you confirm.

TaiPhamD:
When booting directly to grub . I see this message before it goes into the asus logo + Ubuntu logo then starts up normally
978E80F4-3F20-44B4-8FE9-F4BD63CE1903

chainloading without additional ext4 driver in the advanced boot options show that it’s stuck at loading initial RAM disk.
880C9DB3-679C-425F-A355-4A734D845692

I did a google search and people suggested adding additional flag such as dis_ucode_ldr
didn’t seem to help.

mikebeaton:
So, without loading OpenLinuxBoot.efi and without loading ext4_x64.efi, chainloading to GRUB then Ubuntu no longer works in the current version of OC, but used to work before?

I'm that case could you unload both those drivers (and go back to the old drivers layout, since you are going to be checking previous versions of OC), and then do a binary search for the commit in OC which breaks your GRUB chainloading? Because it is sounding in this case as if something else new in OC (not directly this new Linux driver) is preventing your Ubuntu from starting, and the next step is to find out what. (I'm sure you already know the process for binary search to a problem commit, but just in case @82ghost82 wrote a [very nice writeup](https://github.com/acidanthera/bugtracker/issues/1737#issuecomment-891836354) just recently.)

TaiPhamD:
I got it to work finally it was something stupid. SyncRuntimePermissions was false, after I enable it to true it works perfectly!

TaiPhamD:
By the way is there a good way to hide grub2 entry so I can only see the direct kernel image boot ?

vit9696:
Should not show if you do not use `BlessOverride` as grub does not install to standard paths in most distros. You can also avoid installing grub in your Linux distro, which is commonly an option.

TaiPhamD:
@mikebeaton 
I would love to have a way to allow windows to know about the scanned linux kernel images by OC. If we could somehow save this information to NVRAM for example to OC Vendor GUID then I can scan it in my windows [EFI boot selector app](https://github.com/TaiPhamD/EFIBootSelector) to allow users to restart to Linux. We would also need OpenLinuxBoot allows us to change default boot entry in OC for this to work.

mikebeaton:
> I got it to work finally it was something stupid. SyncRuntimePermissions was false, after I enable it to true it works perfectly!

That old chestnut. Thank you for spotting, it would have taken me a while. We need to add something to the docs...

mikebeaton:
> @mikebeaton
> I would love to have a way to allow windows to know about the scanned linux kernel images by OC. If we could somehow save this information to NVRAM for example to OC Vendor GUID then I can scan it in my windows [EFI boot selector app](https://github.com/TaiPhamD/EFIBootSelector) to allow users to restart to Linux. We would also need OpenLinuxBoot to allow us to change default boot entry in OC for this to work.

OpenLinuxBoot working with OpenCore both sets and responds to NVRAM default boot settings. If you boot into Windows via OC when Linux is the OC default, then you will be able to pick up the Boot#### NVRAM setting which OC has created - and perhaps store that as an option to select later? Not perfect, for you, I know, but might help?

TaiPhamD:
> > @mikebeaton
> > I would love to have a way to allow windows to know about the scanned linux kernel images by OC. If we could somehow save this information to NVRAM for example to OC Vendor GUID then I can scan it in my windows [EFI boot selector app](https://github.com/TaiPhamD/EFIBootSelector) to allow users to restart to Linux. We would also need OpenLinuxBoot to allow us to change default boot entry in OC for this to work.
> 
> OpenLinuxBoot working with OpenCore both sets and responds to NVRAM default boot settings. If you boot into Windows via OC when Linux is the OC default, then you will be able to pick up the Boot#### NVRAM setting which OC has created - and perhaps store that as an option to select later? Not perfect, for you, I know, but might help?

Thank you this almost works. However, It seems like whenever you change default boot , it will create Boot#### in the global boot GUID and OCbt#### in OC GUID. Unfortunately, it seems to create it at the same index such as Boot0080 so if you had triple boot Windows , MacOS, or Linux then you can only see Boot0000 as windows for example, then Boot0080 as either Ubuntu or MacOS (depending which was the default). So I think the code that stores the Ubuntu entry into the Boot#### or OCBt#### Variable need to use a different integer index than MacOS so that both linux & macos don't store to the same 0x80 location and wiping themselves out each time a different default is selected.

mikebeaton:
> > > @mikebeaton
> > > I would love to have a way to allow windows to know about the scanned linux kernel images by OC. If we could somehow save this information to NVRAM for example to OC Vendor GUID then I can scan it in my windows [EFI boot selector app](https://github.com/TaiPhamD/EFIBootSelector) to allow users to restart to Linux. We would also need OpenLinuxBoot to allow us to change default boot entry in OC for this to work.
> > 
> > 
> > OpenLinuxBoot working with OpenCore both sets and responds to NVRAM default boot settings. If you boot into Windows via OC when Linux is the OC default, then you will be able to pick up the Boot#### NVRAM setting which OC has created - and perhaps store that as an option to select later? Not perfect, for you, I know, but might help?
> 
> Thank you this almost works. However, It seems like whenever you change default boot , it will create Boot#### in the global boot GUID and OCbt#### in OC GUID. Unfortunately, it seems to create it at the same index such as Boot0080 so if you had triple boot Windows , MacOS, or Linux then you can only see Boot0000 as windows for example, then Boot0080 as either Ubuntu or MacOS (depending which was the default). So I think the code that stores the Ubuntu entry into the Boot#### or OCBt#### Variable need to use a different integer index than MacOS so that both linux & macos don't store to the same 0x80 location and wiping themselves out each time a different default is selected.

Hi, yes that is how it works. That's what I meant about storing it as an option to boot later. I mean you'd have to start Windows once with each OS that you want to be able to return to as the default, then store that value as something to set again later. As I say, not ideal, but might be better than nothing?

HJebbour:
I don't know if I should post my issue here, but when I boot from OC to Ubuntu or Fedora using OpenLinuxBoot my Audio setup is not working (It shows Dummy Output and no Input), if I boot from GRUB audio works properly both input and output

mikebeaton:
@HJebbour It should ideally work. It would make more sense to create a separate issue. Can you post config.plist, debug OC boot log booting into Ubuntu, also your machine and sound specs.

HJebbour:
@mikebeaton Thank you for your reply, please find attached the requested files and find below my machine's specs:

- Lenovo ThinkPad X1 Carbon Gen 8 (i7-10510U)
- Audio device: Intel Corporation Comet Lake PCH-LP cAVS (ALC285)

[log_config.zip](https://github.com/acidanthera/OpenCorePkg/files/8935839/log_config.zip)

askwakhid:
> @mikebeaton Thank you for your reply, please find attached the requested files and find below my machine's specs:
> 
> * Lenovo ThinkPad X1 Carbon Gen 8 (i7-10510U)
> * Audio device: Intel Corporation Comet Lake PCH-LP cAVS (ALC285)
> 
> [log_config.zip](https://github.com/acidanthera/OpenCorePkg/files/8935839/log_config.zip)

Hi, have you solved this issue ?
Thank you.

PR #283, opened 2021-08-08
It should be OpenHfsPlus.efi

fengchen-github:
It seems that HfsPlus.efi can no longer be used in version 0.7.2.


vit9696:
This is garbage. HFSPlus remains being the preferred version and works fine. Make sure you do not need the legacy variant.

PR #286, opened 2021-08-28
Set MinDate and MinVersion default to -1

alyxferrari:
I'm pretty active in the AMD OS X discord server as a helper, and we've been a little bit flooded with people trying to use Catalina or lower confused on why their drive won't show up after stage 1 installation when they use APFS. Since this fixes the issue, you can still boot Big Sur and up with these set to -1, and there are (to my knowledge) no downsides, I figured setting the default to -1 may be a good thing to do.

### DIFF
--- a/Docs/Sample.plist
+++ b/Docs/Sample.plist
@@ -1259,9 +1259,9 @@
 			<key>JumpstartHotPlug</key>
 			<false/>
 			<key>MinDate</key>
-			<integer>0</integer>
+			<integer>-1</integer>
 			<key>MinVersion</key>
-			<integer>0</integer>
+			<integer>-1</integer>
 		</dict>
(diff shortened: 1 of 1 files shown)


MysticalOS:
It was documented well on https://dortania.github.io/hackintosh/updates/2021/08/02/acidanthera-august.html
and is in the 0.7.2 change log.

vit9696:
There is a security downside from doing it: any driver that looks like an APFS driver signed by Apple, including known to be vulnerable drivers, will be loaded. We cannot put our users at the risk, and thus the recommended parameters are chosen this way.

Changing APFS driver version to -1 will still not let you boot Catalina, as Secure Boot will also need to be disabled, so the change is not valid anyway.

PR #291, opened 2021-09-10
Maintain Alphabetical Order in Sample Config Files

dakanji:


.github/workflows/main.yml
```diff
@@ -185,10 +185,31 @@ jobs:
...
           if [ "$(git status --porcelain)" != "" ]; then echo "DataBase desync"; exit 1 ; fi
 
-      - name: Lint Samples
+      - name: Lint and Order Samples
         run: |
-          plutil -lint Docs/Sample.plist || exit 1
-          plutil -lint Docs/SampleCustom.plist || exit 1
+          OurPlistFile='Docs/Sample.plist'
+          plutil -lint "$OurPlistFile" || exit 1
+          if [ "$(git diff $OurPlistFile)" != '' ]; then
```
vit9696, line 192: This does not make sense here. Should be after the perl commands with an exit code in case of success. We do not commit the changes here, so we can only detect incorrect commits and fail the CI.

.github/workflows/main.yml
```diff
@@ -185,10 +185,33 @@ jobs:
...
 
-      - name: Lint Samples
+      - name: Lint Samples and Sort Keys in Alphabetical Order
         run: |
-          plutil -lint Docs/Sample.plist || exit 1
-          plutil -lint Docs/SampleCustom.plist || exit 1
+          OurPlistFile='Docs/Sample.plist'
+          plutil -lint "$OurPlistFile" || exit 1
+          /usr/libexec/PlistBuddy -c 'Save' "$OurPlistFile"
+          /usr/bin/perl -i -p -e 's#<data>\n#<data>#g' "$OurPlistFile"
```
vit9696, line 193: I believe all this code can be done much more efficiently with awk. Check [this answer](https://stackoverflow.com/a/33120488) on SO.

dakanji: Possibly ... used what I know.
Perhaps some awk guru can convert it

dakanji: I suppose the key questions to answer would be:
1. Does it work as is?
2. Is it not being at max "inefficiency" an issue?
3. Does implementation in current form prevent optimisation later?

vit9696: Basically the issue with the existing code is not the algorithmic efficiency, but rather:

- understandability
- amount of lines put here
- correctness

I am not positive that even altogether all these commands will handle large multiline `</data>` strings. So yes, potentially all the three.

dakanji: Yes, not meant for large multiline strings but rather the current short single line items present in the files.
if there is a good possibility that such may come up as it seems, then another solution is needed indeed.

vit9696: Could not you use something like that? This seems to work fine for me to fix `plist` after `Save` in `PlistBuddy`.

```sh
awk ' /<data>$/     {inside=1; printf $0; next}
      /<\/data>$/   {inside=0; print "</data>"; next}
      inside        {gsub ("\t", "", $0); printf $0; next}
                    {print}' "${1}" > "${1}.tmp" && mv "${1}.tmp" "${1}"
```

dakanji: Looks like it goes in a standalone script and is passed something as ${1} right?

dakanji: Not up to speed with requirements. Might be best to get someone with better knowledge level to implement now that the basic idea is set.

dakanji: Wait ... just tested with ${1} changed to ${OurPlistFile} and it does seem to work.

Can you reopen then PR so I can modify?

dakanji: > Can you reopen then PR so I can modify?

Didn't realise I could reopen it myself

.github/workflows/main.yml
```diff
@@ -185,10 +185,25 @@ jobs:
           python2 ./update_generated.py  || exit 1
           if [ "$(git status --porcelain)" != "" ]; then echo "DataBase desync"; exit 1 ; fi
 
-      - name: Lint Samples
+      - name: Lint Samples and Sort Keys in Alphabetical Order
```
vit9696, line 188: I think it is not really sorting them, as it only reports an error if the files are unsorted, which is technically linting as well. Perhaps leave as is?

vit9696:
Looks pretty good besides the minor comment. Please consider it & merge after CI passes.

stevezhengshiqi:
@dakanji Hi, would you mind introducing this sorting algorithm to AppleALC/Resources/PinConfig.kext/info.plist in the order of ALCXXX - layout-id? It's a mess there, and this sorting seems pretty good.

dakanji:
Please raise an issue on the BugTracker so that the Dear Leader can give the request a thumbs up or thumbs down.
Please give me a mention there so I can see what he decides.

PR #294, opened 2021-10-11
Verify recoveryOS dmg using chunklist

jspraul:
Feature request: [...] verify downloaded Internet recoveryOS images
acidanthera/bugtracker#1817

Incorporate @zhangyoufu's [implementation](https://gist.github.com/MCJack123/943eaca762730ca4b7ae460b731b68e7#gistcomment-3061078) of chunklist verification to ensure unencrypted http download is valid.

I am not knowledgable enough to guarantee the signature verification works, but it does not import any non-stdlib dependencies (no pycryptodome) and I have verified that it successfully processes a couple of the available images.

### DIFF
--- a/Utilities/macrecovery/macrecovery.py
+++ b/Utilities/macrecovery/macrecovery.py
@@ -9,6 +9,7 @@ Copyright (c) 2019, vit9696
 from __future__ import print_function
 
 import argparse
+import binascii
 import datetime
 import hashlib
 import json
@@ -82,6 +83,11 @@ def mlb_from_eeee(eeee):
 
   return '00000000000' + eeee + '00'
 
+def int_from_unsigned_bytes(bytes, byteorder):
+  if byteorder == 'little': bytes = bytes[::-1]
+  encoded = binascii.hexlify(bytes)
+  return int(encoded, 16)
+
 # zhangyoufu https://gist.github.com/MCJack123/943eaca762730ca4b7ae460b731b68e7#gistcomment-3061078 2021-10-08
 Apple_EFI_ROM_public_key_1 = 0xC3E748CAD9CD384329E10E25A91E43E1A762FF529ADE578C935BDDF9B13F2179D4855E6FC89E9E29CA12517D17DFA1EDCE0BEBF0EA7B461FFE61D94E2BDF72C196F89ACD3536B644064014DAE25A15DB6BB0852ECBD120916318D1CCDEA3C84C92ED743FC176D0BACA920D3FCF3158AFF731F88CE0623182A8ED67E650515F75745909F07D415F55FC15A35654D118C55A462D37A3ACDA08612F3F3F6571761EFCCBCC299AEE99B3A4FD6212CCFFF5EF37A2C334E871191F7E1C31960E010A54E86FA3F62E6D6905E1CD57732410A3EB0C6B4DEFDABE9F59BF1618758C751CD56CEF851D1C0EAA1C558E37AC108DA9089863D20E2E7E4BF475EC66FE6B3EFDCF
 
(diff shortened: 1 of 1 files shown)


jspraul:
The [recommendation](https://github.com/jspraul/OpenCorePkg/pull/1#issuecomment-939742786) from the primary author, zhangyoufu:

> I would recommend remove the `elif signature_method == 2:` section for production usage.

I've inserted a `RuntimeError` since I wanted to keep the implementation in case anyone else needed it even though it is an error in the Internet recoveryOS case. The rest of the `verify_chunklist` method is directly from [the gist comment linked in the code](https://gist.github.com/MCJack123/943eaca762730ca4b7ae460b731b68e7#gistcomment-3061078) (and `verify_image` is adapted from it).

Utilities/macrecovery/macrecovery.py
```diff
@@ -208,9 +272,10 @@ def action_download(args):
     print(info)
   print('Downloading ' + info[INFO_PRODUCT] + '...')
   dmgname = '' if args.basename == '' else args.basename + '.dmg'
-  save_image(info[INFO_IMAGE_LINK], info[INFO_IMAGE_SESS], dmgname, args.outdir)
+  dmgname = save_image(info[INFO_IMAGE_LINK], info[INFO_IMAGE_SESS], dmgname, args.outdir)
   cnkname = '' if args.basename == '' else args.basename + '.chunklist'
-  save_image(info[INFO_SIGN_LINK], info[INFO_SIGN_SESS], cnkname, args.outdir)
+  cnkname = save_image(info[INFO_SIGN_LINK], info[INFO_SIGN_SESS], cnkname, args.outdir)
+  verify_image(dmgname, cnkname)
```
vit9696, line 278: Could you please add exception handling here? Would not be great to have signature verification failure look like a crash.

Utilities/macrecovery/macrecovery.py
```diff
@@ -208,9 +272,10 @@ def action_download(args):
     print(info)
   print('Downloading ' + info[INFO_PRODUCT] + '...')
   dmgname = '' if args.basename == '' else args.basename + '.dmg'
-  save_image(info[INFO_IMAGE_LINK], info[INFO_IMAGE_SESS], dmgname, args.outdir)
+  dmgname = save_image(info[INFO_IMAGE_LINK], info[INFO_IMAGE_SESS], dmgname, args.outdir)
```
vit9696, line 275: Feels like dmgpath to me? Name is generally just the filename.

vit9696 (commented):
Looks very nice, left two small nitpicks.

Utilities/macrecovery/macrecovery.py
jspraul: Since none of the asserts have any exception text, I thought it might be worth displaying the assertion that failed.

<details>
<summary>temp code to corrupt chunklist (inserted in action_download at line ~279-ish)</summary>

```
  path = cnkpath
  count = int(random.random() * 10)
  offset = 10 # int((os.stat(path).st_size - count) * random.random())
  with open (path, 'r+b') as f:
    f.seek(offset)
    f.write(random.randbytes(count))
  print(f'Wrote {count} bytes at offset {offset} to {path}')
```
</details>

Sample output:
```
Downloading 694-35959...
Saving http://oscdn.apple.com/content/downloads/41/28/071-71279/5djpevqmc2w0ti2c4n1m8mkz7fm791t1ic/RecoveryImage/BaseSystem.dmg to BaseSystem.dmg...
Download complete!MBs downloaded...
Saving http://oscdn.apple.com/content/downloads/41/28/071-71279/5djpevqmc2w0ti2c4n1m8mkz7fm791t1ic/RecoveryImage/BaseSystem.chunklist to BaseSystem.chunklist...
Download complete!25 MBs downloaded...
Wrote 4 bytes at offset 10 to /Users/user/Downloads/dev/OpenCorePkg/Utilities/macrecovery/BaseSystem.chunklist
Verifying image with chunklist...
Image verification failed. (assert signature_method in [1, 2])
```

Utilities/macrecovery/macrecovery.py
jspraul: Renamed *name -> *path variables for `verify_image`.

jspraul:
Please hold off until I have time to look into supporting Python <3.2, which doesn't have `int.from_bytes` used in digital signature verification.

Utilities/macrecovery/macrecovery.py
vit9696: Looks nice, thank you!

jspraul:
I tested with recovery_urls.txt again for both Python 2.7.16 and Python 3.9.7.

Reimplementation of int.get_bytes leaning on StackOverflow ([1](https://stackoverflow.com/questions/444591/how-to-convert-a-string-of-bytes-into-an-int/33585619#33585619), [2](https://stackoverflow.com/questions/6624453/whats-the-correct-way-to-convert-bytes-to-a-hex-string-in-python-3/6624521#6624521)) after fighting Python 2 byte<->str attempting to re-implement from the [source](https://github.com/python/cpython/blob/8e5de40f90476249e9a2e5ef135143b5c6a0b512/Objects/longobject.c#L787).

PR #295, opened 2021-10-21
Build: Made stack-protector guard use global addressing instead of tls.

MikhailKrichanov:
Needs at least clang-12 for successful compilation.


vit9696:
@dhinakg, @Goldfish64, @mhaeuser we need to update clang to 12, but it does not seem to work right away. Could you suggest anything?

PR #296, opened 2021-10-29
SSDT-PNLF.dsl

gervasiodias:


khronokernel:
What is the point of this PR? In the header it's even written to adjust on a per-user/system basis:
```c
// Rename GFX0 to anything else if your IGPU name is different.
```

PR #297, opened 2021-11-03
Work around partial overwriting of text by adding spaces at the end of the string

Mahas1:
This issue is observed when downloading macOS using macrecovery.py. Text is partially overwritten, and to work around this, spaces can be added to the end of the strings. 

The issue
> Download complete!Bs downloaded...

Another approach would be to replace `\r` with `\n`, which should print the text in a new line. However, I am not completely sure how it would affect things in the variety of OSes this tool can be used in.

### DIFF
--- a/Utilities/macrecovery/macrecovery.py
+++ b/Utilities/macrecovery/macrecovery.py
@@ -295,7 +295,7 @@ def action_download(args):
         err = linecache.getline(tb.tb_frame.f_code.co_filename, tb.tb_lineno, tb.tb_frame.f_globals).strip()
       except:
         err = "Invalid chunklist"
-    print('\nImage verification failed. ({})'.format(err))
+    print('\rImage verification failed. ({})'.format(err))
     return 1
 
 def action_selfcheck(args):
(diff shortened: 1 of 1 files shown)


dreamwhite:
The code itself can be optimized by using `' '*20` but in any case... 

I wanted to ask to the core team (CC @vit9696)

Is there any specific reason macrecovery.py is based on Python2 code? Can't it be adapted with Python3 syntax and thus edit the shebang at the first line? Python3 is used pretty everywhere and I don't see any reason behind the usage of a deprecated Python version as Python2.

Just to mention one of, the use of `f-strings` can improve the readability of the code itself.
If it's appreciated I think I can make a PR in the next days.

Best regards
dreamwhite

vit9696:
The reason we provide python2-compatible tools is that this version of python is still bundled with macOS.

dreamwhite:
> The reason we provide python2-compatible tools is that this version of python is still bundled with macOS.

Alright so will a Python 3 converted PR be appreciated?

vit9696:
Was I unclear? python2 is still bundled with macOS, there is no reason to require python3, which is unsupported without third-party or first-party extra actions (like Xcode tools installation).

iabtw:
Odd question, but wouldn't it be easier to preview what's going on, code-wise, by simply replacing the manual insertion of spaces (each individual character) with simply multiplying a singular character `N` times to yield the same output? 

Not a breaking change, just something for readability.

Apologies if this is a redundant or dumb question.

PR #299, opened 2021-11-07
ocvalidate: Add checks for UIScale

PMheart:


Utilities/ocvalidate/ValidateUefi.c
```diff
@@ -521,6 +523,13 @@ CheckUEFIOutput (
   if (UIScale < -1 || UIScale > 2) {
     DEBUG ((DEBUG_WARN, "UEFI->Output->UIScale is borked (Can only be between -1 and 2)!\n"));
     ++ErrorCount;
+  } else if (UIScale != -1) {
+    mHasUefiOutputUIScale = TRUE;
```
vit9696, line 527: You do not need to make this one a global variable.

PR #307, opened 2021-12-11
AudioDxe: Updates

mikebeaton:
General:
  - Added DisconnectHda quirk
    - Useful on at least Apple hardware and VMware Fusion guest
  - Added AudioOutRange support [**EDIT**: Changed to AudioOutMask instead.]
    - Allows playing sound on treble and bass speakers where these are different (adjacent) outputs - noticeable improvement in sound quality
    - This requires new EFI_AUDIO_IO2 protocol, since EFI_AUDIO_IO protocol was not versioned, and changes are not compatible (new argument in some protocol methods); have implemented versioned protocol to avoid needing this again [**EDIT**: Now just changed GUID of existing protocol, as suggested.]
  - Sound working on Apple hardware (Cirrus Logic)
    - Requires GPIO bit write to enable speakers (and another bit to enable headphones), discovered from reading Linux drivers for this card
  - QEMU (working around bugs in QEMU intel-hda)
    - Improved fix for one bug in intel-hda (which had a partial fix present, but commented out)
    - New fix for another bug in intel-hda
  - VMWARE
    - DisconnectHda quirk removes need for separate VMware detection at OpenProtocol
    - Introducing to-spec poll of CORB reset write (see 'bugfixes' below) hangs VMware Fusion, so added controller quirk for this based on VendorId

Sound quality on real hardware remains excellent, of course. Sound quality on the VMs (QEMU and VMware Fusion) are each similar to each other - the initial chime is fine, but when playing multiple samples in the boot menu, the sound is very 'droppy' (i.e. whole samples seem to get missed or partially played). It almost sounds as if this is to do with the DMA in the host playing from the wrong part of the guest's DMA buffer, but I have no idea if this is correct.

Driver connection is still noticeably slow, but no slower than before.

Bugfixes:
  - Incorrect sense of bit check in ResetController (would have just left running controllers running)
  - Add required (according to spec) bit poll in ResetController (required after previous fix, or we end up stopping running controllers)
  - Add required (according to spec) bit poll in CORB reset write
  - On some firmware (i.e. DxeCore-based, e.g. in QEMU) SetTimer raises TPL to TPL_HIGH_LEVEL - 1; we had one point where we raised TPL to TPL_HIGH_LEVEL and then called SetTimer inside it, which causes an assert in the UEFI code on such systems; changed that code so that we raise our TPL to TPL_HIGH_LEVEL - 1
  - If failure during initialising streams, reset streams could be called for all streams, result in freeing NULL ptrs + segfault (happened in QEMU before work-arounds for QEMU intel-hda). Added NULL check.

Code changes:
  - Changed all logging which is produced (i.e. not commented out or VERBOSE) to "HDA: ..."
  - Changed code style to EDK-II where I was touching the code or closely related code anyway, have not attempted to change comprehensively everywhere
  - Changed several methods called from ConnectController to return EFI_STATUS instead of BOOLEAN
    - I think there were a couple of cases where the return BOOLEAN was (incorrectly, since err. and success are reversed) being used as an EFI_STATUS later in the code path, after an error
    - More generally, makes it easier to consistently log where and what the error was if controller setup fails, which I have done
  - Moved playback delay code inside playback setup. Avoids duplication, also allows:
  - Implemented simple caching in playback setup (meaning, in general, playback delay is only needed once, before first sample, since setup is normally only run non-cached once, if all samples are the same rate and bitness)

Other:
  - Removed unwanted warning about blank device path from ocvalidate, since this is actually a very useful setting
  - Updated documentation for new features and other minor changes about audio

https://github.com/acidanthera/bugtracker/issues/740


mikebeaton:
I've rebased this and added a changelog - I'm pretty happy for it to go live and into test sooner rather than later.

Two issues to maybe check first:

1. @vit9696 Might want to check that I haven't said anything obviously wrong in the new text in `Configuration.tex`.

2. The biggest change in how AudioDxe works, for anyone who doesn't use the new features, is that it now always runs RIRB/CORB in the HDA controller in a kind of hybrid polling/interrupt mode (previously it was pure polling):
 - Sets up interrupt mode, but no interrupt handler
 - Actually polls, but when data is ready, clears the 'data present' interrupt bit, causing the next data to transmit

The reason for this is that QEMU intel-hda only works in this mode (and unless we use another way to detect QEMU, QEMU can only be detected _after_ this is set up, since its codecs report as QEMU, but its controller reports as Intel; and you can't scan the codecs from the controller until RIRB/CORB is set up).

I actually believe this approach should be safe, and possibly even beneficial on other firmware, since polling is the less preferred mode of operation, and only intended for use in boot services (therefore I guess more likely to be broken elsewhere too). I can already say that it works fine in QEMU (ofc), VMware Fusion guest, MBP10,2 and a Dell-based hack.

Docs/Configuration.tex
```diff
@@ -6908,14 +6918,40 @@ \subsection{Audio Properties}\label{uefiaudioprops}
   The quickest way to find the right port is to bruteforce the values from \texttt{0} to \texttt{N - 1}.
 
 \item
+  \texttt{AudioOutRange}\\
```
vit9696, line 6921: Perhaps this should be a bit mask?

Include/Acidanthera/Library/OcDriverConnectionLib.h
```diff
@@ -46,6 +46,14 @@ OcDisconnectGraphicsDrivers (
   VOID
   );
 
+/**
+  TODO:
```
vit9696, line 50: TODO :)

Include/Acidanthera/Library/OcHdaDevicesLib.h
```diff
@@ -58,6 +58,17 @@
...
 
+//
+// Structure used for HDA PCI class code parsing.
+//
+#pragma pack(1)
+typedef struct {
+  UINT8 ProgInterface;
+  UINT8 SubClass;
+  UINT8 Class;
+} HDA_PCI_CLASSREG;
```
vit9696, line 69: I think this structure already exists in EDK II, and generally 4 bytes are read at once (including revision) through UINT32 read mode. Better to change.

Include/Acidanthera/Library/OcMiscLib.h
```diff
@@ -25,9 +25,15 @@
...
+  Conversions to microseconds for use in e.g. gBS->Stall.
 **/
-#define SECONDS_TO_MICROSECONDS(x) ((x)*1000000)
+#define SECONDS_TO_MICROSECONDS(x) ((x) * 1000000)
+#define MS_TO_MICROSECONDS(x)      ((x) * 1000)
+
+/**
+  Conversions to nanoseconds for use in e.g. PciIo->PollMem.
+**/
+#define MS_TO_NANOSECONDS(x)       ((x) * 1000000)
```
vit9696, line 36: Maybe also add SECONDS_TO_NANOSECONDS for completeness.

Include/Acidanthera/Protocol/AudioIo.h
```diff
@@ -28,13 +28,15 @@
 #include <Uefi.h>
 
 /**
-  Audio I/O protocol GUID.
+  Audio I/O 2 protocol GUID - versioned.
```
vit9696, line 31: The file does not change to AudioIo2, so it is already a breaking change. Just swap the GUID in my opinion and leave the name untouched.

Library/OcAudioLib/OcAudioLib.c
```diff
@@ -89,19 +100,31 @@ OcAudioInstallProtocols (
...
+    FoundOtherProtocols = FALSE;
     for (Index = 0; Index < ARRAY_SIZE (mAudioProtocols); ++Index) {
       Status = gBS->LocateProtocol (
         mAudioProtocols[Index],
         NULL,
         &Protocol
         );
       if (!EFI_ERROR (Status)) {
+        //
+        // Don't exit early, so we can log all found protocols.
```
vit9696, line 113: Guard with DEBUG_CODE_BEGIN? No need to log it in RELEASE builds.

Library/OcMainLib/OpenCoreUefi.c
```diff
@@ -345,7 +345,10 @@ OcReinstallProtocols (
   BOOLEAN       InstallAppleEvent;
   BOOLEAN       OverrideAppleEvent;
 
-  if (OcAudioInstallProtocols (Config->Uefi.ProtocolOverrides.AppleAudio) == NULL) {
+  if (OcAudioInstallProtocols (
```
vit9696, line 348: Better to make a variable than try to pass into an if in my opinion.

Staging/AudioDxe/HdaCodec/HdaCodecAudioIo.c
```diff
@@ -254,13 +248,34 @@ HdaCodecAudioIoSetupPlayback(
...
+
+  DEBUG((DEBUG_VERBOSE, "HdaCodecAudioIoSetupPlayback(): start\n"));
+
+  // Basic settings caching.
+  if (mOutputIndex == OutputIndex &&
+      mVolume == Volume &&
+      mFreq == Freq &&
+      mBits == Bits &&
+      mChannels == Channels
+    ) {
```
vit9696, line 269: I think it generally is attached to `if`.

Staging/AudioDxe/HdaController/HdaController.c
```diff
@@ -811,12 +853,14 @@ HdaControllerDriverBindingSupported (
...
     && (HdaClassReg.Class != PCI_CLASS_MEDIA
     || HdaClassReg.SubClass != PCI_CLASS_MEDIA_MIXED_MODE)) {
-    Status = EFI_UNSUPPORTED;
+    return EFI_UNSUPPORTED;
   }
 
   return Status;
 }
 
+BOOLEAN mUseGetProtocolVMwareOnly = FALSE;
```
vit9696, line 862: Is this one dead?

vit9696:
Looks generally good. Left a few notes.

Library/OcMainLib/OpenCoreUefi.c
mikebeaton: This is the same layout as e.g. OcAppleEventInstallProtocol just below in this file - or am I misunderstanding?

Library/OcAudioLib/OcAudioLib.c
mikebeaton: I think I really want to `break;` at the end of `!EFI_ERROR (Status) { ... }` block, but only on RELEASE build. Is there something like RELEASE_CODE_BEGIN?

Docs/Configuration.tex
mikebeaton: (I was trying not to overcomplicate, but it should be.)

Include/Acidanthera/Library/OcHdaDevicesLib.h
mikebeaton: Is it the same thing as `USB_CLASSC` in EDK-II? If so, then it is indeed read as 3 bytes, but I can change to use that one instead.

Library/OcAudioLib/OcAudioLib.c
mikebeaton: Or, as I think there is not, is it okay to add it in OC, modelled on DEBUG_CODE_BEGIN but with `!DebugCodeEnabled ()`??

Docs/Configuration.tex
vit9696: That will work as long as we have not too many AudioOuts, but I think I have never seen over 63…

Include/Acidanthera/Library/OcHdaDevicesLib.h
vit9696: While it is, the one for USB semantically does not fit the use. I rather meant:

https://github.com/acidanthera/OpenCorePkg/blob/f61a1a35cbd12ec3758b0d510c0aec99d665abb2/Library/OcDeviceMiscLib/ActivateHpetSupport.c#L65
https://github.com/acidanthera/OpenCorePkg/blob/f61a1a35cbd12ec3758b0d510c0aec99d665abb2/Library/OcDeviceMiscLib/ResetAudioTrafficClass.c#L64
https://github.com/acidanthera/OpenCorePkg/blob/0.7.6/Library/OcDeviceMiscLib/SetResizableBar.c#L353

To be fair, maybe it is actually a good idea to change all four as the existing approach is not too great anyway. Could you define a PCI_CLASSCODE structure same as USB_CLASSC and switch to it everywhere?

Library/OcAudioLib/OcAudioLib.c
vit9696: I think what you might want here is to log protocols in the DEBUG build and return the right protocol in both builds. Perhaps split the loops, leaving the logging one outside in DEBUG_CODE brackets? This will make the code cleaner.

Library/OcMainLib/OpenCoreUefi.c
vit9696: OcAppleEventInstallProtocol is also ugly :)

Include/Acidanthera/Library/OcHdaDevicesLib.h
mikebeaton: Hmm - just looking at the naming, I don't think `USB_CLASSC` can be the correct struct? (I think it must be the same thing, but named for a specific use-case?).

In which case, by hunting for references to the `PCI_CLASSCODE_OFFSET` in the EDK-2 code, we find that there is, indeed, no (other) struct for this in EDK-2, and also that it is always read as 3 bytes (not 4 bytes as one uint32).

So I think the net result is no change? Or possibly just change the naming `SubClass`->`SubClassCode` and `Class`->`ClassCode` which are used everywhere (both in `USB_CLASSC`, and also in various `UINT8` local variable names) in EDK-2.

mikebeaton: Sorry, we commented at the same time. Okay, will do.

Library/OcAudioLib/OcAudioLib.c
mikebeaton: Is this okay: https://github.com/acidanthera/OpenCorePkg/blob/1d60f1d07a6a2116d7708b0e715dc34b7328e263/Library/OcAudioLib/OcAudioLib.c#L116-L121 ?

Library/OcMainLib/OpenCoreUefi.c
mikebeaton: Hang on, basically every single line below this works the same way ... !

Library/OcAudioLib/OcAudioLib.c
vit9696: Well, I personally would prefer my solution. It would be shorter and cleaner for RELEASE builds. Note, the code requires reordering OC protocol as the last one:

```C
    DEBUG_CODE_BEGIN ();
    for (Index = 0; Index < ARRAY_SIZE (mAudioProtocols); ++Index) {
      Status = gBS->LocateProtocol (
        mAudioProtocols[Index],
        NULL,
        &Protocol
        );
      DEBUG ((DEBUG_INFO, "OCAU: %g protocol - %r\n", mAudioProtocols[Index], Status));
    }
    DEBUG_CODE_END ();
    for (Index = 0; Index < ARRAY_SIZE (mAudioProtocols) - 1; ++Index) {
      Status = gBS->LocateProtocol (
        mAudioProtocols[Index],
        NULL,
        &Protocol
        );
      if (!EFI_ERROR (Status) {
        return NULL;
      }
    }
    Status = gBS->LocateProtocol (
      mAudioProtocols[ARRAY_SIZE (mAudioProtocols) - 1],
      NULL,
      &Protocol
      );
    if (!EFI_ERROR (Status)) {
      return Protocol;
    }
```

Library/OcMainLib/OpenCoreUefi.c
vit9696: Well, we already often use `Result` variable for OcOverflow functions for exactly that purpose. But can leave here as we will use Uncrustify soon anyway.

Include/Acidanthera/Library/OcHdaDevicesLib.h
mikebeaton: https://github.com/acidanthera/OpenCorePkg/pull/307/commits/4364ac86b926ce2ba5770969639bb15fcd2e3b71

Library/OcAudioLib/OcAudioLib.c
mikebeaton: The meaning changed in the above (should return OC protocol if found, else null if anything else found, else go ahead and install OC protcol). I've gone for this and not reordering the list, I think it is fine:

```C
    DEBUG_CODE_BEGIN ();
    for (Index = 0; Index < ARRAY_SIZE (mAudioProtocols); ++Index) {
      Status = gBS->LocateProtocol (
        mAudioProtocols[Index],
        NULL,
        &Protocol
        );
      DEBUG ((DEBUG_INFO, "OCAU: %g protocol - %r\n", mAudioProtocols[Index], Status));
    }
    DEBUG_CODE_END ();
    for (Index = 0; Index < ARRAY_SIZE (mAudioProtocols); ++Index) {
      Status = gBS->LocateProtocol (
        mAudioProtocols[Index],
        NULL,
        &Protocol
        );
      if (!EFI_ERROR (Status)) {
        if (Index == 0) {
          return (OC_AUDIO_PROTOCOL *) Protocol;
        }
        return NULL;
      }
    }
```

PR #308, opened 2021-12-17
Patches to several Svace warnings.

MikhailKrichanov:


Legacy/BootPlatform/GraphicsConsoleDxe/GraphicsConsole.c
```diff
@@ -1153,6 +1153,9 @@ GraphicsConsoleConOutSetMode (
                           ModeData->GopHeight,
                           0
                           );
+      if (EFI_ERROR (Status)) {
```
vit9696, line 1156: Here and below no Status checks are intentional as Blt may fail before mode-set. Should probably write a comment to avoid a future miscorrection.

Legacy/BootPlatform/GraphicsConsoleDxe/GraphicsConsole.c
```diff
@@ -1198,6 +1201,9 @@ GraphicsConsoleConOutSetMode (
...
+        goto Done;
```
vit9696, line 1205: Intentional as mentioned above.

Library/OcCryptoLib/BigNumPrimitives.c
```diff
@@ -125,7 +125,7 @@ BigNumLeftShiftWordsAndBits (
   IN     UINT8             NumBits
   )
 {
-  UINTN Index;
+  UINT16 Index;
```
vit9696, line 128: This change looks weird, could you explain the reasoning behind it? I suspect that UINT16 is actually OC_BN_NUM_WORDS, but even in this case I fail to properly understand the idea.

Library/OcFileLib/FileProtocol.c
```diff
@@ -31,9 +31,9 @@ EfiGetEpochDays (
   IN  EFI_TIME  *Time
   )
 {
-  UINTN a;
+  UINT8 a;
   UINTN y;
-  UINTN m;
+  UINT8 m;
```
vit9696, line 36: Are these changes trying to reduce stack usage or fixing any potential security problem?
* If it is the former, UINTN alignment will not let it become less, will have to reorder the variable declaration.
* If it is the latter, I do not get the reasoning behind either.

Platform/OpenUsbKbDxe/EfiKey.c
```diff
@@ -468,24 +468,23 @@ USBKeyboardDriverBindingStart (
...
-      gBS->CloseEvent (UsbKeyboardDevice->SimpleInputEx.WaitForKeyEx);
-    }
-    if (UsbKeyboardDevice->KeyNotifyProcessEvent != NULL) {
-      gBS->CloseEvent (UsbKeyboardDevice->KeyNotifyProcessEvent);
-    }
-    ReleaseKeyboardLayoutResources (UsbKeyboardDevice);
-    UsbKbFreeAppleKeyMapDb (UsbKeyboardDevice);
-    FreePool (UsbKeyboardDevice);
-    UsbKeyboardDevice = NULL;
+  if (UsbKeyboardDevice->TimerEvent != NULL) {
```
vit9696, line 471: This change looks good to me, nice find, but please also change `ASSERT (UsbKeyboardDevice != NULL);` to be a proper NULL-check with goto Error1 on failure. We do not want surprises in RELEASE builds in case memory ends.

Platform/OpenUsbKbDxe/KeyBoard.c
```diff
@@ -520,9 +520,9 @@ InitKeyboardLayout (
   EFI_KEY_DESCRIPTOR        *TableEntry;
   EFI_KEY_DESCRIPTOR        *NsKey;
   USB_NS_KEY                *UsbNsKey;
-  UINTN                     Index;
-  UINTN                     Index2;
-  UINTN                     KeyCount;
+  UINT8                     Index;
+  UINT8                     Index2;
+  UINT8                     KeyCount;
```
vit9696, line 525: Is this change reducing type sizes because USB_KEYBOARD_LAYOUT_PACK_BIN DescriptorCount is UINT8? Can be a good idea, but I believe at least indices would want some suffix to denote that it is not a normal Index of UINTN type. It has happened for me when I reused an Index variable of a different type in a loop assuming it was UINTN by mistake.

vit9696 (commented):
Would like @mhaeuser to give a look.

Library/OcCryptoLib/BigNumPrimitives.c
mhaeuser: I like the change because, despite being very implausible here obviously, it removes architecture-specific behaviour that could be a source for bugs (i.e. this index would have a higher range for x64 over x86). I try to omit UINTN like the plague nowadays.

Library/OcFileLib/FileProtocol.c
mhaeuser: Imo walk the whole way and also change the other variables away from UINTN (and potentially rearrange), or don't bother at all

Library/OcCryptoLib/BigNumPrimitives.c
vit9696: Well, at least OC_BN_NUM_WORDS Index maybe?

mhaeuser: Yes, I took that for granted :)

Platform/OpenUsbKbDxe/KeyBoard.c
vit9696: EDK II codestyle requires capitalised local variables, so better use something like `DescIndex` and `DescIndex2` probably.

PR #309, opened 2021-12-19
Docs: Add Github action that validates sample configs

glekner:


glekner:
@vit9696 is this beneficial? 
Is there a more capable `dtd` that we can use?
All the sanity checkers iv'e seen are not supporting the latest versions, understandable as its hard to keep up and maintain those

glekner:
looks like `plutil` is already taking care of that.
I wonder if besides linting and validating we can have a more strict validation and sanity checking @vit9696

PR #311, opened 2021-12-27
Audio test

mikebeaton:
@vit9696 , @Goldfish64 - if you had a moment to review some bits of this, I made some more proposed changes to Audio.

It is split into separate small commits to make it easier to review, I hope.

The only two changes I'd really appreciate some feedback on are:

 - @vit9696, @Goldfish64 : be54db75394ad39c6d3863c1d6bb35d8dc00f1d1 , I have not just moved the repeated 'reset pci attributes' code into the shared call which it previously came before in two places, but also moved that code to the end of the shared method, even though it was called before it, i.e. I am now resetting to the original PCI state at a _different_ time to when it was done before. According to this commit, any previous PCI settings are now restored after the rest of the controller clear down, since they are set before the rest of the controller set up; I wanted to double-check that I haven't missed any reason why this is the wrong ordering
 - @vit9696 : b375de3dbf38f0c54023107c317e42e3075c333c , I've introduced 3 PCDs for AudioDxe
   -  Firstly, perhaps you could double-check that I've introduced the PCDs correctly (though I think I have), but mainly:
   - `PcdAudioCodecErrorOnNoOutputs`: pretty sure this one makes sense, and that the code this introduces is neater than the `#define` I had used before
   - `PcdAudioControllerTryProtocolGetMode`: I think this one is good, as it preserves your code to try EFI_OPEN_PROTOCOL_GET_PROTOCOL as a fallback after EFI_OPEN_PROTOCOL_BY_DRIVER, even though it not turned on (as I don't _think_ we need it, with 'DisconnectHda' quirk)
   - `PcdAudioControllerResetOnExitBootServices`: least sure about this one, there was an (unused) var in the controller struct to hold an Exit Boot Services event already (@Goldfish64, perhaps if you have time, you might want to check that I've used it for more or less what you had planned for it?), but even though I have now implemented this - to try resetting the controller on exit, to try to resolve the no-sound-after-AudioDxe-in-Windows-on-a-few-machines issue (and @Goldfish64 suggested the same idea on OC private discussion thread, if I understood him correctly, I think I did) - this did not actually make any difference to that problem, unfortunately; so not sure whether to leave this more or less untested/not-proven-worthwhile code in there on a turned-off PCD or not?


mikebeaton:
@vit9696 : there is actually a not-quite-related commit in there ef34d463a5c19e359794821303fb477ea2276c46 to add some more logging (either VERBOSE or error-condition only) in OcExitBootServices, particularly to log a warning if it won't exit.

As you're no doubt well aware, I found if you do something illegal in OcExitBootServices (e.g. I was trying to disconnect the sound controller in AudioExitBootServices....) and thereore OcExitBootServices, and so ExitBootServices, fails, then all OSes (macOS, Linux, Windows) retry gBS->ExitBootServices (and therefore re-call OcExitBootServices) (a bit like ForceExitBootServices, but it is not ForceExitBootServices) at least once, sometimes more when I added some test code to fail on purpose; which means it's possible for gBS->ExitBootServices to fail silently once then succeed, so I thought adding a warning to the log when it fails was probably useful.

Library/OcAfterBootCompatLib/ServiceOverrides.c
```diff
@@ -964,10 +966,14 @@ OcExitBootServices (
   // For non-macOS operating systems return directly.
   //
   if (BootCompat->ServiceState.AppleBootNestedCount == 0) {
-    return BootCompat->ServicePtrs.ExitBootServices (
+    Status = BootCompat->ServicePtrs.ExitBootServices (
       ImageHandle,
       MapKey
       );
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_WARN, "OCABC: OcExitBootServices (%amacOS) - %r\n", "non-", Status));
```
vit9696, line 974: You are not allowed to print within/after ExitBootServices. Even on failure.

Library/OcAfterBootCompatLib/ServiceOverrides.c
```diff
@@ -995,6 +1001,7 @@ OcExitBootServices (
   // Abort on error.
   //
   if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_WARN, "OCABC: OcExitBootServices (%amacOS) - %r\n", "", Status));
```
vit9696, line 1004: Same as above.

Library/OcAfterBootCompatLib/ServiceOverrides.c
mikebeaton: https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcMainLib/OpenCoreUefiInOut.c#L63-L67

mikebeaton: Actually, I guess the above is not (strictly) within or after, so is different?

mikebeaton: Actually, again, no - the above is still after the memory map has been taken - so not entirely clear of the difference... ... I suppose in this case here, the memory map has been taken, and the debug lib might have been torn down? (Probably the other comment is sightly misleading still, though? Or could do with a new comment - maybe?? - where I tried to add some code?)

Library/OcAfterBootCompatLib/ServiceOverrides.c
mikebeaton: Yes, agreed, as above.

Library/OcAfterBootCompatLib/ServiceOverrides.c
mikebeaton: I won't push this - ofc! - and I'll offer sightly updated comments in as separate commit and see what you think.

Sher1ocks:
hi mike. i report log message.

my desktop board is asus b150m.

Library/OcAfterBootCompatLib/ServiceOverrides.c
vit9696: The example is different, as the print is almost impossible to appear, and is a bug. Here the print will frequently trigger on many machines and will bork them.

Library/OcDriverConnectionLib/DriverDisconnection.c
```diff
@@ -243,8 +246,17 @@ OcDisconnectHdaControllers (
         continue;
       }
 
+      DevicePath = ConvertDevicePathToText (
```
vit9696, line 249: Use DebugPrintPath in my opinion.

mikebeaton:
> hi @mikebeaton . i report log message.  my desktop mainboard is asus b150m.

@Sher1ocks Thank you for the report. Just to be clear, which versions does this occur on?
 - [ ] this `audio-test` branch?
 - [ ] current `master` branch with work-in-progress 0.7.7 after recent changes to AudioDxe?
 - [ ] 0.7.6 release before any changes to AudioDxe?

Sher1ocks:
> > hi @mikebeaton . i report log message.  my desktop mainboard is asus b150m.
> 
> @Sher1ocks Thank you for the report. Just to be clear, which versions does this occur on?
> 
> * [ ]  this `audio-test` branch?
> * [ ]  current `master` branch with work-in-progress 0.7.7 after recent changes to AudioDxe?
> * [ ]  0.7.6 release before any changes to AudioDxe?

1. no audio-test branch result. just master branch

Sher1ocks:
sorry my mistake close button in mobile. please back reopen.

mikebeaton:
> > > hi @mikebeaton . i report log message.  my desktop mainboard is asus b150m.
> > 
> > 
> > @Sher1ocks Thank you for the report. Just to be clear, which versions does this occur on?
> > 
> > * [ ]   this `audio-test` branch?
> > * [ ]   current `master` branch with work-in-progress 0.7.7 after recent changes to AudioDxe?
> > * [ ]   0.7.6 release before any changes to AudioDxe?
> 
> 1. no audio-test branch result. just lastest master branch
> 2. yes.
> 3. there is no log like attached pic in 0.7.6

Okay, there is no log like attached pic in 0.7.6, but again just to be clear: did the driver actually work for you before in 0.7.6, i.e. did it produce sound for you before (in 0.7.6) but now (0.7.7) it does not? Or is it just a difference in what is logged?

Sher1ocks:
> > 1. no audio-test branch result. just lastest master branch
> > 2. yes.
> > 3. there is no log like attached pic in 0.7.6
> 
> Okay, there is no log like attached pic in 0.7.6, but again just to be clear: did the driver actually work for you before in 0.7.6, i.e. did it produce sound for you before (in 0.7.6) but now (0.7.7) it does not? Or is it just a difference in what is logged?

@mikebeaton 
1. 0.7.7(also 0.7.6) sound is working before this 706cb4e AudioDxe: Updates
2. 0.7.7(also 0.7.6) there is log, but since AudioDxe: Updates, i always get time out log.

mikebeaton:
@Sher1ocks Thanks. Can you send me a debug boot log from before and after the commit which breaks sound for you?

Sher1ocks:
> @Sher1ocks Thanks. Can you send me a debug boot log from before and after the commit which breaks sound for you?
can you use element? i'm in online now

Library/OcAfterBootCompatLib/ServiceOverrides.c
```diff
@@ -927,6 +927,17 @@ OcStartImage (
...
+  Notes:
+    - Most OSes attempt to call ExitBootServices more than once if it fails initially
+      (similar to OpenCore ForceExitBootServices)
+      - Therefore, OcExitBootServices may get called more than once
+      - However this should never be relied upon for correct operation
+    - Any logging within this call but before original ExitBootServices is attempted
+      (e.g. within a scheduled handler) may cause ExitBootServices to fail (e.g. it may
+      change the memory map by allocating), and should only be done, if at all, in the
+      case of unexpected errors
+    - Never log after original ExitBootServices has been attempted, not even on error
```
mikebeaton, line 940: @vit9696 Are these comments correct, and do you think worth having if so?

Library/OcDebugLogLib/DebugPrint.c
```diff
@@ -36,6 +36,23 @@ DebugPrintDevicePath (
...
+  DEBUG_CODE_BEGIN ();
+  
+  EFI_DEVICE_PATH_PROTOCOL  *DevicePath;
+
+  DevicePath = DevicePathFromHandle (Handle);
+  DebugPrintDevicePath (ErrorLevel, Message, DevicePath);
+
+  DEBUG_CODE_END ();
+}
+
```
mikebeaton, line 55: Also just to double-check this is okay to add?

Library/OcAfterBootCompatLib/ServiceOverrides.c
vit9696: Yes.

Library/OcDebugLogLib/DebugPrint.c
vit9696: DebugPrintDevicePath does not take OPTIONAL DevicePath, so better to check the result of DevicePathFromHandle for NULL. This change will need to be added to all Debug libraries in OpenCore.

Staging/AudioDxe/HdaController/HdaController.c
```diff
@@ -324,10 +330,31 @@ HdaControllerGetName (
   DEBUG ((DEBUG_INFO, "HDA: Controller is %s\n", HdaControllerDev->Name));
 }
 
+STATIC
+VOID
+EFIAPI
+HdaControllerExitBootServicesHandler (
```
vit9696, line 336: Need to clarify this is for Windows I believe?

Staging/AudioDxe/HdaController/HdaController.c
```diff
@@ -364,6 +391,24 @@ HdaControllerReset (
     }
   }
 
+  if (!Restart) {
+    return EFI_SUCCESS;
+  }
+
+  if (PcdGetBool (PcdAudioControllerResetPciOnExitBootServices) &&
```
vit9696, line 398: We usually put && to the next line.

Library/OcDebugLogLib/DebugPrint.c
```diff
@@ -20,14 +20,17 @@ VOID
 DebugPrintDevicePath (
   IN UINTN                     ErrorLevel,
   IN CONST CHAR8               *Message,
-  IN EFI_DEVICE_PATH_PROTOCOL  *DevicePath
+  IN EFI_DEVICE_PATH_PROTOCOL  *DevicePath  OPTIONAL
```
mikebeaton, line 23: @vit9696 I am proposing making this argument OPTIONAL in this one and the new one, that seemed like the cleanest change. Is that okay?

Library/OcDebugLogLib/DebugPrint.c
mikebeaton: I've updated and added new question on revised commit.

Library/OcDebugLogLib/DebugPrint.c
vit9696: Yes.

mikebeaton:
@vit9696 - Thank you for all the feedback. Give me a short while (might not be till tomorrow) to double-check that everything's behaving as I expect on the test systems I have here, and then I will build the docs and push to master.

PR #312, opened 2022-01-03
OcAppleEventLib: Allow tuning pointer poll list and period

vit9696:


vit9696:
This pull request resolves issues mentioned in #1899.

Include/Acidanthera/Library/OcAppleEventLib.h
```diff
@@ -33,6 +36,9 @@
   @param[in] GraphicsInputMirroring If true, disable Apple default behaviour which can
                                     prevent keyboard input reaching non-Apple GUI UEFI apps.
                                     OC builtin AppleEvent only.
+  @param[in] PointerPollMin         Pointer polling minimal frequency.
+  @param[in] PointerPollMax         Pointer polling maximum frequency.
```
mikebeaton, line 40: Specify the units here?

Library/OcAppleEventLib/PointerHandler.c
```diff
@@ -90,7 +91,11 @@ STATIC UINTN mNumberOfPointerProtocols = 0;
 STATIC EFI_EVENT mSimplePointerPollEvent = NULL;
 
 // mSimplePointerPollTime
-STATIC UINT64 mSimplePointerPollTime;
+STATIC UINT64 mSimplePointerPollTime = POINTER_POLL_ALL_MASK;
```
mikebeaton, line 94: !

Docs/Configuration.tex
```diff
@@ -6828,6 +6828,51 @@ \subsection{AppleInput Properties}\label{uefiappleinputprops}
   \item \texttt{false} --- Prevent key input mirroring to non-Apple protocols when in graphics mode.
   \end{itemize}
 
+  \item
+  \texttt{PointerPollMin}\\
+  \textbf{Type}: \texttt{plist\ integer}\\
+  \textbf{Failsafe}: \texttt{0}\\
+  \textbf{Description}: Configure minimal pointer polling period in ms.
```
mhaeuser, line 6835: Do we have any convention on "minimal" vs "minimum" (see "maximum" below)?

Include/Acidanthera/Library/OcAppleEventLib.h
```diff
@@ -18,6 +18,9 @@
 
 #include <Protocol/AppleEvent.h>
 
+#define POINTER_POLL_DEFAULT_FREQUENCY       0
```
mikebeaton, line 21: Not as sure about this comment, but possibly the code is more readable if this is called `POINTER_POLL_DEFAULT`? With the current name then where used in the code, I initially read it as containing the default frequency value to use, rather than as being a placeholder which will be replaced with the default frequency.

Docs/Configuration.tex
```diff
@@ -6828,6 +6828,51 @@ \subsection{AppleInput Properties}\label{uefiappleinputprops}
...
+
+  This is the minimal period AppleEvent polls pointer devices (e.g. mice, trackpads)
+  for motion events. Defaults to 10 ms in the OpenCore AppleEvent implementation
+  and to 2 ms in the OEM Apple implementation.
```
mhaeuser, line 6839: Should probably explicitly state "when 0" because we have no defaults. :) But hmm, for TiimerResolution "0" means "unchanged", and "0" here just picks a hardcoded default, why not just set the default in the Sample as with more properties?

Library/OcAppleEventLib/PointerHandler.c
```diff
@@ -43,8 +44,8 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 //         most machines. Poll with 10 ms, which matches the keyboard behaviour,
 //         and also is the minimum for QEMU.
 //
-#define POINTER_POLL_FREQUENCY  EFI_TIMER_PERIOD_MILLISECONDS (10)
-#define MAX_POINTER_POLL_FREQUENCY  EFI_TIMER_PERIOD_MILLISECONDS (80)
+#define MIN_POINTER_POLL_FREQUENCY  /*EFI_TIMER_PERIOD_MILLISECONDS*/ (10 * 10000)
+#define MAX_POINTER_POLL_FREQUENCY  /*EFI_TIMER_PERIOD_MILLISECONDS*/ (80 * 10000)
```
mhaeuser, line 48: Why?

Library/OcAppleEventLib/PointerHandler.c
```diff
@@ -725,6 +752,11 @@ InternalSimplePointerPollNotifyFunction (
     CommonStatus = EFI_NOT_READY;
 
     for (Index = 0; Index < mNumberOfPointerProtocols; ++Index) {
+      if (mSimplePointerPollMask != POINTER_POLL_ALL_MASK
+        && (LShiftU64 (1, Index) & mSimplePointerPollMask) == 0) {
```
mhaeuser, line 756: Why U64 when we are limited to U32 by the PLIST format (and mSimplePointerPollMask in fact is UINT32 already)? Also, should we maybe warn if there are pointer protocols that cannot be described by the mask? Will this not be UB for Index >= 64?

Docs/Configuration.tex
vit9696: I believe not. We use Minimum in the options, but otherwise choose whatever fits best in the context. I think I will leave it unchanged here.

Docs/Configuration.tex
vit9696: The issue is the OEM and Builtin implementations, which have different defaults. I do not want this to be confusing.

Library/OcAppleEventLib/PointerHandler.c
vit9696: The original has MulDiv function, so is not a constant expression.

PR #313, opened 2022-01-06
Update SSDT-PNLF.dsl

trxxgkxxn7:


khronokernel:
Couple lines up:
```asl
// Rename GFX0 to anything else if your IGPU name is different.
```

Samples are to be used as examples, not as-is. Configure on per system basis.

Additionally why is this such a common PR for this file specifically?
* https://github.com/acidanthera/OpenCorePkg/pull/285/files
* https://github.com/acidanthera/OpenCorePkg/pull/296/files

Can't tell if these are just bot accounts here to troll.

PR #315, opened 2022-01-08
Update SSDT-EC.dsl

huuduc141:


mswgen:
Looks like just another #285 #288 #296 #303 thing.

PR #316, opened 2022-01-11
AudioDxe: Allow gain to track OS volume on old macOS without SystemAudioVolumeDB

mikebeaton:
Quite a lot of changes to get this to work, but I think it's worth it - not least because it is also the only access to system volume which we have when in-OS volume is muted (as `SystemAudioVolumeDB` contains junk in that case; but `SystemAudioVolume` is still reasonable) (and we need a volume when muted for audio assist).

Confirmed working on old, un-upgraded MacPro4,1 running OS Lion(!).


Docs/Sample.plist
```diff
@@ -1204,8 +1204,6 @@
 				<integer>0</integer>
 				<key>SystemAudioVolume</key>
 				<data>Rg==</data>
-				<key>SystemAudioVolumeDB</key>
-				<data>4g==</data>
```
mikebeaton, line 1208: I've removed these again, because on machine running old macOS this NVRAM value would be treated as authoritative by OpenCore, but would never be updated by macOS.

mikebeaton:
Also worth noting that on the above test machine, as well a not having `SystemAudioVolumeDB` there is also no volume limit on `SystemAudioVolume` - it follows the main OS slider all the way up to 0 dB/Max.

I'm wondering if `SystemAudioVolumeDB` was introduced in macOS at the same time as the volume limit on `SystemAudioVolume` - but just a guess, it would need find the OS version where both are present and the one before that to double-check.

Docs/Sample.plist
vit9696: Yes, makes sense.

vit9696 (commented):
Looks good to me, no obvious issues spotted.

mikebeaton:
Many thanks for the review. I will rebase and merge when I get a chance to, within the next couple of days.

PR #317, opened 2022-01-13
Update and rename Docs/AcpiSamples/Source/SSDT-RTC0.dsl to Dokumenty/…

HiKiKoFuToKo:
…AcpiSamples/Źródło/SSDT-RTC0.dsl


HiKiKoFuToKo:
asus ux325

khronokernel:
* Why are you changing the file pathing for a single file and translating it to your own language when the rest of OpenCorePkg is in English.
* Why are adding comments as objects in the source itself instead of as comments?
* This file will not compile as-is due incorrect syntax

PR #320, opened 2022-02-05
OcAppleKernelLib: Kext block strategy for prelinked

PMheart:


Library/OcAppleKernelLib/KextPatcher.c
```diff
@@ -364,6 +364,103 @@ PatcherApplyGenericPatch (
...
+        KextIdentifier = XmlNodeContent (KextPlistValue);
+        if (PlistNodeCast (KextPlistValue, PLIST_NODE_TYPE_STRING) == NULL || KextIdentifier == NULL) {
+          return EFI_NOT_FOUND;
+        }
+        if (AsciiStrCmp (KextIdentifier, Identifier) == 0) {
+          DEBUG ((DEBUG_INFO, "OCAK: Erasing %a from prelinked at index %u\n", Identifier, Index));
+          //
+          // Erase kext.
+          //
+          XmlNodeRemove (PrelinkedContext->KextList, KextPlist);
```
mhaeuser, line 454: We have the index in "Index", no? Can we not call XmlNodeRemoveByIndex() directly?

Library/OcAppleKernelLib/KextPatcher.c
```diff
@@ -364,6 +364,103 @@ PatcherApplyGenericPatch (
...
+  //
+  Segment    = NULL;
+  AddressMax = 0;
+  MaxSize    = 0;
+  while ((Segment = MachoGetNextSegment (&PatcherContext->MachContext, Segment)) != NULL) {
+    VirtualAddress = PatcherContext->Is32Bit ? Segment->Segment32.VirtualAddress : Segment->Segment64.VirtualAddress;
+    Size           = PatcherContext->Is32Bit ? Segment->Segment32.Size : Segment->Segment64.Size;
+
+    AddressMax = MAX (VirtualAddress + Size, AddressMax);
+    MaxSize    = AddressMax - PatcherContext->VirtualBase;
```
mhaeuser, line 404: Can be set once outside the loop as it is now.

Library/OcAppleKernelLib/KextPatcher.c
PMheart: Makes sense, fixed.

Library/OcAppleKernelLib/KextPatcher.c
```diff
@@ -364,6 +364,103 @@ PatcherApplyGenericPatch (
...
+    return EFI_UNSUPPORTED;
+  }
+  DEBUG ((
+    DEBUG_INFO,
+    "OCAK: Excluding %a - VirtualBase %Lx, MaxSize %Lx\n",
+    Identifier,
+    PatcherContext->VirtualBase,
+    MaxSize
+    ));
+  ZeroMem (KextData, (UINTN) (AddressMax + MaxSize));
```
mhaeuser, line 422: It holds that
``MaxSize  == AddressMax - PatcherContext->VirtualBase``
So we have
``AddressMax + MaxSize == AddressMax + (AddressMax - PatcherContext->VirtualBase) == 2 * AddressMax - PatcherContext->VirtualBase``
Should this not be just ``MaxSize``?

PMheart: My bad, indeed!

Library/OcAppleKernelLib/KextPatcher.c
```diff
@@ -364,6 +364,102 @@ PatcherApplyGenericPatch (
...
+        if (PlistNodeCast (KextPlistValue, PLIST_NODE_TYPE_STRING) == NULL || KextIdentifier == NULL) {
+          return EFI_NOT_FOUND;
+        }
+        if (AsciiStrCmp (KextIdentifier, Identifier) == 0) {
+          DEBUG ((DEBUG_INFO, "OCAK: Erasing %a from prelinked at index %u\n", Identifier, Index));
+          //
+          // Erase kext.
+          //
+          XmlNodeRemoveByIndex (PrelinkedContext->KextList, Index);
+          return EFI_SUCCESS;
```
vit9696, line 454: In addition to removing the kext from `KextList` you also need to remove it from `Prelinked->PrelinkedKexts`. Otherwise you may happen to run into a situation where `InternalCachedPrelinkedKext` returns a corrupt node from the cache after the deletion.

I guess what you need to do is to add a new function:
```C
EFI_STATUS
InternalDropCachedPrelinkedKext (
  IN OUT PRELINKED_CONTEXT  *Prelinked,
  IN     CONST CHAR8        *Identifier
  );
```

This function should look up the kext in the cache like `InternalCachedPrelinkedKext` and return `EFI_NOT_FOUND` otherwise, and then do something like:

```C
    Kext = GET_PRELINKED_KEXT_FROM_LINK (Link);
    RemoveEntryList (Link);
    InternalFreePrelinkedKext (Kext);
```

See `PrelinkedContextFree` for more details.

Library/OcAppleKernelLib/PrelinkedInternal.h
```diff
@@ -198,6 +198,15 @@ InternalCachedPrelinkedKext (
   IN     CONST CHAR8        *Identifier
   );
 
+/**
+  Drops cached PRELINKED_KEXT from PRELINKED_CONTEXT.
```
vit9696, line 202: We should honestly update the docs in this file, but perhaps not in this PR.

vit9696:
This code looks good to me. @khronokernel, @dhinakg we need this to be tested today.

PR #323, opened 2022-02-24
Update firmware data

wy414012:


PMheart:
@Andrey1970AppleLife Hi Andrey! Could you please review this? Thanks!

Andrey1970AppleLife:
> @Andrey1970AppleLife Hi Andrey! Could you please review this? Thanks!

I can fix and merge it, but usually I don't update on a firmware from beta macOS.

Andrey1970AppleLife:
Oh, OEMStrings is edited manually, but OEMStrings is necessary to copy entirely. And I doubt that the utility for unpacking was used.
Closed

wy414012:
@Andrey1970AppleLife  Thanks for pointing out the inadequacies, I can use `strings ./*.fd | grep -A 10 "** ROM Version"` to match the printout directly to the yam file, as I found some firmwares have UUIDs and some don't. I don't know if direct output replacement will cause an exception.

PR #327, opened 2022-03-08
OcDebugLibProtocol: Early log preservation support

PMheart:


Library/OcDebugLibProtocol/OcDebugLibProtocol.c
```diff
@@ -81,9 +208,19 @@ DebugPrint (
 
   if (OcLog != NULL) {
     OcLog->AddEntry (OcLog, ErrorLevel, Format, Marker);
-  } else if ((ErrorLevel & GetDebugPrintErrorLevel ()) != 0) {
+  } else {
     UnicodeVSPrintAsciiFormat (Buffer, sizeof (Buffer), Format, Marker);
-    gST->ConOut->OutputString (gST->ConOut, Buffer);
+
```
mikebeaton, line 213: Might as well only do ^^^ if it is used on either branch.

PMheart: `Buffer` is allocated on the stack anyway I think?

mikebeaton: True - so won't unnecessarily fragment memory. Minor point really, then - but still seems worth checking, maybe? (Not quite sure whether compiler can do anything useful with these Pcds, here? But I gather the general idea is stuff won't even be compiled, if it can be detected that Pcd disables it at compile time - as of course you know! As I say, not quite sure if this can apply here or not.)

PMheart: https://github.com/acidanthera/OpenCorePkg/commit/d780844c95f3c412f9c7c2e86c2fd493807a5086

mikebeaton: tyvm

PR #331, opened 2022-04-05
OcAppleKernelLib: Added `CustomPciSerialDevice` quirk

PMheart:
closes https://github.com/acidanthera/bugtracker/issues/1954

After merging:

- [ ]   Update Chagelog
- [ ]   Rebuild pdf
- [x]  Call `IsZeroBuffer` instead under `PatchSetApfsTrimTimeout` (off-topic)


Docs/Configuration.tex
```diff
@@ -2573,6 +2573,13 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
   on all systems but those explicitly dedicated to scientific or media calculations.
   Only certain Xeon models typically benefit from the patch.
 
+\item
+  \texttt{CustomPCISerialDevice}\\
+  \textbf{Type}: \texttt{plist\ boolean}\\
+  \textbf{Failsafe}: \texttt{false}\\
+  \textbf{Requirement}: 10.4\\
```
PMheart, line 2580: I am not sure what the minimum requirement is, though the lowest version makes sense to me.

Docs/SampleCustom.plist
```diff
@@ -1001,6 +1001,8 @@
 			<false/>
 			<key>AppleXcpmForceBoost</key>
 			<false/>
+			<key>CustomPCISerialDevice</key>
```
vit9696, line 1004: Pci to match others.

vit9696:
Please add the port getting/setting code, and add the original patch, while @joevt takes time to explain where to insert it.

Docs/SampleCustom.plist
PMheart: https://github.com/acidanthera/OpenCorePkg/commit/321a06d855987a17d4d23025cfa562011ece1599

PMheart: https://github.com/acidanthera/OpenCorePkg/commit/321a06d855987a17d4d23025cfa562011ece1599

joevt:
We are discussing a kernel patch that will make the kernel output to a specific PCIe serial port. Applying the patch will not have any affect unless boot-args is setup to make the kernel do serial port output (usually used by `kprintf`).

The patch is for PCI cards based on the 16x50 serial controller that have an I/O BAR base address register (such as the StarTech PEX2S953). For a built-in serial port, the I/O base is usually 0x3f8. The patch will change this to the I/O base of a specified PCIe card.

BaseSerialPortLib16550.c has code for getting the I/O address from the PCIe card that has path PcdSerialPciDeviceInfo.

Function GetSerialRegisterBase in BaseSerialPortLib16550.c does a small amount of validation and initialization to make serial output work. The `for BusNumber` loop goes through all the PCI bridges in the PcdSerialPciDeviceInfo path to make sure they have proper bus numbers, subordinate bus numbers, and I/O ranges. Each node in PcdSerialPciDeviceInfo specifies a device and function (starting from bus 0).

The last node in the PcdSerialPciDeviceInfo path is for the serial port PCIe device. Function GetSerialRegisterBase uses the first I/O BAR for the SerialRegisterBase.

Perhaps `CustomPciSerialDevice` should be a UEFI path that is separate from gEfiMdeModulePkgTokenSpaceGuid.PcdSerialPciDeviceInfo ? It's not really necessary that they be the same serial port (for example, you can have OpenCore go to one port and the kernel to go to another port). This way, you can get the device handle using the UEFI path, and use the PCI device protocol from the handle to access the PCI card. You can assume that all the PCI bridges and the PCI device are properly enumerated and setup (all the ranges are valid). Maybe verify that the specified card has an appropriate class code.

The final thing that GetSerialRegisterBase does is enable the I/O of the PCI device and the parent PCI bridges. I don't think we need to worry about parent PCI bridges? Just enable the I/O for the PCI device. I don't think we need to worry about the D0 state (at least in my setup; maybe more modern setups need this).

My patch (using Kernel -> Patch) replaces `66baf803` with `66ba0820` with mask fffff8ff which in assembly is:
`mov dx, 0x3f8` -> `mov dx, 0x2008` where 0x2008 is the I/O base for my PCIe card (it changes depending on the slot and other PCIe cards). The mask means the patch changes all values in the range 0x3f8..0x3ff to 0x2008..0x200f. I would describe the patch as changing the "base address" rather than the "address range". "base" implies that there is a range.

Since we are making a quirk, you can add a bit more safety to the search and replace. I/O addresses are read and written to using the `in` and `out` instructions. Therefore maybe you should ensure that each match is followed by `ec` or `ee` within 20 or so bytes (because there may be some instructions between `mov` and `in` or `out`). Add a debug statement if you don't find the in/out instruction. Add a debug statement to count the number of replacements.

In Hopper.app, you can search for matches using this: `66BAF[8-9a-f]03(..)*e[ec]` (where the `*` is a non-greedy capture (which one would usually write as `*?`)

Are there any kernels that don't use register dx for the port address, or register al for the value? The instructions are from `/pexpert/i386/pe_serial.c`.  They use functions `outb` and `inb`.  Those are defined in `i386/pio.h` as:
`__asm__ volatile("inb %w1, %b0" : "=a" (datum) : "Nd" (port));`
`__asm__ volatile("outb %b0, %w1" : : "a" (datum), "Nd" (port));`
I'm not sure what `Nd` means. I guess `a` is `al` since the only other value supported by `in` and `out` is `ax` which is a 16-bit value.

The reason for this patch:
`pe_serial.c` can do serial output to the following (ordered by preference):
1) PCH LPSS UART2: MMIO base: mmio_uart boot-arg, 0xFE036000, or 0xFE034000 (macOS 10.12.6)
2) Legacy: I/O base 0x3f8 (macOS 10.3.9)
3) PCIE_MMIO UART: MMIO base: pcie_mmio_uart boot-arg, or 0xFE410000 (macOS 10.14.1)

My serial PCIe card does not support the PCIE_MMIO UART method (and that method doesn't exist before macOS 10.14), and the legacy method doesn't have a boot-arg to change the I/O base. Therefore, this patch. This patch can't work if serial_init chooses option #1 but you can get around that by having boot-arg `mmio_uart=0`.

So, what CustomPciSerialDevice will do is redirect kernel legacy serial I/O to a PCIe card. For this to work, you also need to make sure that the Apple16X50PCI0 driver does not attach to the serial port because it will stop the serial output. For this I use a code-less kext as described in https://github.com/acidanthera/bugtracker/issues/1954

vit9696:
In my opinion, we can assume that the serial ports are the same for simplicity of the interface. For a quirk I am not sure that trying to match out/in will be easily doable with the compiler generating different code.

For MMIO I think we need to make our patch automatically replace the mmio address too.

All in all there should be decent docs in OC describing all this stuff.

PMheart:
Thanks @joevt and @vit9696.

With @vit9696's help, I implemented an initial draft of both the PMIO and MMIO patches.

As for the PMIO patch, I believe it is safe enough to match the instruction `mov dx, 0x3fx` (`0x66, 0xBA, 0xF8, 0x03`) by solving symbol `_serial_init`. This way the searching process is done only within this routine, and with Limit set to 1024 we won't search very farther from its beginning. So, I did not try to match `in` (`0xEC`) or `out` (`0xEE`).

For MMIO, there is one constant `0xFE0340xx` within `_serial_init`. With the same matching method talked above, this is easily done as well.

Actually, how about debug kernels? To be honest I did not check carefully, but at a glance `0xFE0340xx` did not seem to exist; or I think it is done otherwise (e.g. via mov dword ptrs)?

Under function `PatchSetPciSerialDeviceRegisterBase`, the mode of patching (i.e. MMIO vs PMIO) is determined based on the RegisterBase received from `GetSerialRegisterBase` function call.

Library/OcAppleKernelLib/CommonPatches.c
```diff
@@ -1064,6 +1064,126 @@ PatchIncreasePciBarSize (
...
+  .Skip        = 0,
+  .Limit       = 1024
+};
+
+STATIC
+CONST UINT8
+mSerialDeviceMmioFind[] = {
+  0x00, 0x40, 0x03, 0xFE  ///< mov whatever, 0xFE0340xx
+};
+STATIC_ASSERT (sizeof (mSerialDeviceMmioFind) == 4, "Unsupported mSerialDeviceMmioFind");
```
mhaeuser, line 1108: This looks odd, why must it be exactly 4? Maybe just ASSERT Find, Mask, and Replace are of equal length, whatever it is?

PMheart: https://github.com/acidanthera/OpenCorePkg/pull/331/commits/d42e4665b129677bdf8766b76e8153e37da8d0b3

joevt:
> In my opinion, we can assume that the serial ports are the same for simplicity of the interface.

Ok, so Serial -> Init is for OpenCore serial output.
CustomPciSerialDevice is for xnu serial output.
If Init or CustomPciSerialDevice are true, then call `SerialPortInitialize`.
If Init is not true, then don't do OpenCore serial output.
If CustomPciSerialDevice is not true, then don't patch the kernel.

> For a quirk I am not sure that trying to match out/in will be easily doable with the compiler generating different code.

My comments about `inb` and `outb` was to suggest methods of ensuring that `in` and `out` are always present in the same form near the bytes to be patched. I will write a script to check all the macOS kernels to get statistics.

> For MMIO I think we need to make our patch automatically replace the mmio address too.

Sounds like fun. macOS has boot-args for changing mmio address for the two MMIO options, but you think it would be better if the values came from the BAR registers of the device specified by `PcdSerialPciDeviceInfo`?

> For MMIO, there is one constant `0xFE0340xx` within `_serial_init` as of 10.13. With the same matching method talked above, this is easily done as well.

That's for the PCH LPSS UART2 option. It's not used if there's a UART at 0xFE0360xx so that should be the register to look for. It has existed in the xnu source code since at least macOS 10.12.6 (I did not check every version of 10.12). Is the PCH LPSS UART2 option usable by BaseSerialPortLib16550.c? Is the PCH LPSS UART2 a PCI device? It seems to have different registers than the PCIE_MMIO UART option - actually the registers are similar but the access width is 32 bits so maybe it can work.

For the PCIE_MMIO UART option, the address to modify is 0xFE4100xx.

> As for the PMIO patch, I believe it is safe enough to match the instruction `mov dx, 0x3fx` (`0x66, 0xBA, 0xF8, 0x03`) by solving symbol `_serial_init`. This way the searching process is done only starting from this routine, and with `Limit` set to `1024` we won't search very farther from the beginning. So, I did not try to match `in` (`0xEC`) or `out` (`0xEE`) for both simplicity and performance.

But there are other places in the kernel that do serial port output (hibernate? stuff)

What does PMIO stand for?

> Under function `PatchSetPciSerialDeviceRegisterBase`, the mode of patching (i.e. MMIO vs PMIO) is determined based on the size of `RegisterBase` received from `GetSerialRegisterBase` function call. Since we override `RegisterBase` in `Misc->Serial` section, does the order matter?

If you want to support both MMIO options (LPSS and PCIE), then you need to be able to differentiate between the two. Perform the LPSS patch if `PcdSerialUseMmio` is true and `PcdSerialRegisterAccessWidth` is 32 bit. Perform the PCIE_MMIO patch if `PcdSerialUseMmio` is true and `PcdSerialRegisterAccessWidth` is false.

For PMIO, check the `PcdSerialUseMmio` and `PcdSerialRegisterAccessWidth`. Don't perform the PMIO patch if `PcdSerialUseMmio` is false or `PcdSerialRegisterAccessWidth` is not 8 bit.

mhaeuser:
@joevt Why would we initialise serial when Init=false? This sounds borked. It’s not “OC serial output” (that should be determined by Debug->Target) but “init serial port”.

PMIO is port-mapped I/O.

Library/OcMainLib/OpenCoreKernelPatch.c
```diff
@@ -302,6 +314,16 @@ OcKernelApplyPatches (
       OcKernelApplyQuirk (KernelQuirkAppleXcpmForceBoost, CacheType, DarwinVersion, NULL, &KernelPatcher);
     }
 
+    if (Config->Kernel.Quirks.CustomPciSerialDevice) {
+      RegisterBase = GetSerialRegisterBase ();
```
vit9696, line 318: This code executes regardless of `Serial` → `Init`, meaning that config.plist values set in the `Misc` → `Serial` will be ignored here, but this quirk will still be applied. @mhaeuser, is that ok?

mhaeuser: @vit9696 There already is an ongoing discussion about when we need to init, need to resolve that first: https://github.com/acidanthera/OpenCorePkg/pull/331#issuecomment-1091118957

vit9696:
> @joevt Why would we initialise serial when Init=false? This sounds borked. It’s not “OC serial output” (that should be determined by Debug->Target) but “init serial port”.
> 
> PMIO is port-mapped I/O.

Well, this comment is correct. Serial → Init is literally "init serial port", it has nothing to do to the fact whether OC uses serial or not. Perhaps we should separate `Enable` and `Init` though, but in reality no PCI card would work without `Init`,

Library/OcMainLib/OpenCoreKernelPatch.c
PMheart: How about applying this patch in OpenCoreMisc.c instead? Is this possible?

joevt:
We need to at least call GetSerialRegisterBase to ensure that IO or MMIO is enabled for the PCI device and parent PCI bridges. SerialPortInitialize calls GetSerialRegisterBase.

If we're not doing serial output in OpenCore, then calling GetSerialRegisterBase should be sufficient, as xnu has its own code that performs initialization and set baud rate.

Should we override baud rate and clock since we are overriding base register?
pe_serial.c:
LEGACY_UART_CLOCK       1843200
DEFAULT_UART_BAUD_RATE 115200

vit9696:
If these patches are contributed, it is ok to me, but otherwise I would act based on the actual need to.

PMheart:
Regarding baud rate and clock in XNU - We can implement it. I would expect we update these values when performing Serial init. i.e.:
- Misc->Serial section updates the PCD values
- CustomPciSerialDevice kernel quirk patches XNU for the same thing

joevt:
I ran some statistics.
```
IFS=$'\n'
for themacos in $(
for thekernel in $((ls /Volumes/*/System/Library/Kernels/kernel; ls /Volumes/*/mach_kernel) 2> /dev/null); do
	thevolume=$(perl -pe 's|(/Volumes/[^/]*).*|\1|' <<< $thekernel)
	macosversion=$(/usr/libexec/PlistBuddy -c 'Print :ProductVersion' "$thevolume/System/Library/CoreServices/SystemVersion.plist")
	echo "$macosversion $thekernel"
done | sort -V
); do
	thekernel="${themacos#* }"
	perl -0777 -nE '
		$curpos=0;
		while ( /(?{$x=pos()})\x66\xba([\xf8-\xff])\x03/mg ) {
			$offset = ord($1) - 0xf8;
			if ( /(?{$y=pos()})([\xee\xec])/mg ) {
				$out = ord($1) == 0xee ? 1 : 0;
				print $out . " " . $offset . " " . ($y - $x - 4) . "\n";
			} else {
				print "x " . $offset . " x\n";
			}
			pos() = $x + 4;
		}
	' $thekernel > /tmp/matches.txt
	echo "$themacos"
	sort -V /tmp/matches.txt | uniq -c
	echo $(wc -l < /tmp/matches.txt)
done
```

It seems we need a different patch for Mac OS X versions 10.4 to 10.6. I'll have a look.
10.9.5 Mavericks has a suspicious gap of 46 bytes between a match and the `in` instruction. I'll double check the disassembly.
[counts.txt](https://github.com/acidanthera/OpenCorePkg/files/8447484/counts.txt)

PMheart:
Thanks. I took a quick glance at 10.4.11 XNU as well, and indeed a different patch is needed.

For 10.9, we can just change the max search distance (currently 32) to a higher value. e.g. 64.

joevt:
I don't think we can do a simple search/replace for 10.4.11. It has optimizations that make it difficult to find.
Look at this example:
```
003bd181 BBFB030000             mov        ebx, 0x3fb
003bd186 B880FFFFFF             mov        eax, 0xffffff80
003bd18b 89DA                   mov        edx, ebx
003bd18d EE                     out        dx, al
003bd18e B2F8                   mov        dl, 0xf8
003bd190 B85A000000             mov        eax, 0x5a
003bd195 EE                     out        dx, al
003bd196 EC                     in         al, dx
```
It moves 0x3fb into ebx which is used for the first `out`, but it replaces only the low byte with 0xf8 for the next `out` and `in`. I suppose you could just replace the entire function (but you have to replace all functions that use the same static globals).

PMheart:
This sounds complex. In my opinion, do not bother with 10.4 for now?

joevt:
> do not bother with 10.4 for now?

I'm ok with that.

The Mavericks 46 byte gap seems benign.
```
ffffff80006edb5b 66BAFD03               mov        dx, 0x3fd
ffffff80006edb5f 90                     nop

                                    loc_ffffff80006edb60:
ffffff80006edb60 C745FC00000000         mov        dword [rbp+var_4], 0x0       ; CODE XREF=_serial_putc+80
ffffff80006edb67 8B45FC                 mov        eax, dword [rbp+var_4]
ffffff80006edb6a 3D0F270000             cmp        eax, 0x270f
ffffff80006edb6f 7F1C                   jg         loc_ffffff80006edb8d

ffffff80006edb71 6666666666662E0F1F840000000000  nop        word [cs:rax+rax]

                                    loc_ffffff80006edb80:
ffffff80006edb80 FF45FC                 inc        dword [rbp+var_4]            ; CODE XREF=_serial_putc+75
ffffff80006edb83 8B45FC                 mov        eax, dword [rbp+var_4]
ffffff80006edb86 3D10270000             cmp        eax, 0x2710
ffffff80006edb8b 7CF3                   jl         loc_ffffff80006edb80

                                    loc_ffffff80006edb8d:
ffffff80006edb8d EC                     in         al, dx                       ; CODE XREF=_serial_putc+47
```

Attached is a list of file offsets for non-zero gaps.
[counts_offsets_nonzerogaps.txt](https://github.com/acidanthera/OpenCorePkg/files/8447615/counts_offsets_nonzerogaps.txt)

PMheart:
OK, thanks. I just increased the maximum search size to 64. Looks fine now?

If so, let's go back to the MMIO patch.

joevt:
Let me know what you would like me to check or test.

Library/OcAppleKernelLib/CommonPatches.c
```diff
@@ -1064,6 +1064,135 @@ PatchIncreasePciBarSize (
...
+
+        ++Walker;
+      }
+    }
+
+    //
+    // Patch PMIO.
+    //
+    if (FoundPmio) {
+      *WalkerPmio++ = (mSerialDevicePmioPort[0] & 0xF8U);
```
PMheart, line 1148: @joevt I am not sure whether we should `& 0xF8U` for replacement?

joevt: For PMIO:
For replacement, we want to preserve the 3 least significant bits.
I think we can assume the 3 least significant bits are always zero (this is a property of PMIO base registers?).
Therefore:
```
*WalkerPmio++ = (RegisterBase & 0xff) | (*WalkerPmio & 7);
*WalkerPmio = RegisterBase >> 8;
```

If you don't think the least significant bits are always zero, then:
```
UINT16 port = RegisterBase + (*WalkerPmio & 7);
*WalkerPmio++ = port & 0xff;
*WalkerPmio = port >> 8;
```

For MMIO:
I think we should not do kernel patches for MMIO. I think we should modify the boot-args value - or intercept `PE_parse_boot_argn` and return our calculated value for `mmio_uart` or `pcie_mmio_uart` if the user has indicated in config.plist that we should do so (with CustomPciSerialDevice) and has not provided those values in boot-args themselves.

PMheart: I have no idea to be honest, nor can I test. Could you please experiment and figure out which method is correct?

As for MMIO, I am ok without patching, until someone explicitly requests that.

We would also like to have this PR merged by next Monday’s release. Do you think we can do it? ^^

joevt: > EDIT - Yes, the 3x LSB must be zero indeed; see https://github.com/acidanthera/audk/blob/d2f79716cb67c0487433b159f882015773dd18b7/MdePkg/Library/BaseIoLibIntrinsic/IoLibMsc.c#L248 (thanks to @mhaeuser)

That's for 32-bit IO ports. And it's only checking the two least significant bits.
PMIO serial registers uses 8-bit IO ports which can have any value for the three least significant bits.

The question was about the three least significant bits for the IO BAR (the base for all the serial registers) https://en.wikipedia.org/wiki/PCI_configuration_space says that IO BAR uses all the bits except the two least significant bits. Therefore, I would use the `+` method instead of the `|` option.

But wait. https://wiki.osdev.org/PCI#Base_Address_Registers says the size of the BAR is equal to two to the power of the number of bits that don't change (except the info bits) when you write 1's to all the bits of the BAR. Since the serial port IO BAR has size 8 (for all the registers), the IO BAR must be a multiple of 8. Therefore the `|` method should also work.

So either method should work. I will test the current patch today.

PMheart: OK, since I do not have any device test, nor do others do most likely - I guess it will be only you who can tell whether the patch works.

By the way, I just figured out that my previous code could be simplified, which is done at https://github.com/acidanthera/OpenCorePkg/pull/331/commits/ff5d7863fe3af0a1bb5fb0767b494d85f83d34c3.

joevt: I tested https://github.com/acidanthera/OpenCorePkg/commit/ff5d7863fe3af0a1bb5fb0767b494d85f83d34c3
I disabled my xnu patch to make sure only the CustomPciSerialDevice quirk is used.

If I set Serial->Init to false, then OpenCore doesn't override the Pcd's, and CustomPciSerialDevice does a replacement of 0x3f8 with 0x3f8.

If I set Serial->Init to true, then I see serial output from OpenCore and xnu which means the CustomPciSerialDevice quirk works. But I want to be able to set Serial->Init to false to only have serial output in xnu if I'm only debugging xnu.

- Pcd's should always be overridden, regardless of Serial->Init. I think Serial->Init just needs to control calling SerialPortInitialize.
- CustomPciSerialDevice should not do PMIO patch if the base address is the same as the default 0x3f8.
- This message "Registering PCI serial device PMIO port 1016" should use hex (e.g. "0x3f8") for the base address number.

vit9696: In my opinion,  the current behaviour is correct. Leave Init to be true, remove Target serial bit, and you should only see XNU output.

joevt: > But I want to be able to set Serial->Init to false to only have serial output in xnu if I'm only debugging xnu.

Maybe I'm wrong about setting Serial->Init to false to do serial only in xnu.

The proper way to disable serial output in OpenCore is to unset `OcLog->Options & OC_LOG_SERIAL`. I don't like these types of flags. The individual flags are hidden (only listed in the documentation, not the .plist) and are difficult to check (need to convert to binary to see if the flag is already set) and to change (need to convert back to binary to change the flag). plist files are xml. xml are already verbose and bulky so you're not saving anything by having multiple flags encoded into a single inscrutable number. I would convert these flags into a `dict` but that's a project for another day and this was probably already considered when creating those multi-flag integers.

OcLog calls SerialPortWrite even if SerialPortInitialize wasn't called. This probably shouldn't happen ... unless the point of setting Serial->Init to false is to use the initialization that is performed by some other code outside OpenCore (such as PciSioSerialDxe) or in some other library used by OpenCore (such as OcDebugLibSerial).

OcDebugLibSerial calls SerialPortInitialize even if Serial->Init is false. This doesn't matter since the default DebugLib is OcDebugLibNull and this doesn't change unless you recompile OpenCore.

joevt: > In my opinion, the current behaviour is correct. Leave Init to be true, remove Target serial bit, and you should only see XNU output.

I'm ok with that. `OcLog->Options & OC_LOG_SERIAL` comes directly from Debug → Target.

PMheart: > CustomPciSerialDevice should not do PMIO patch if the base address is the same as the default 0x3f8.

> This message "Registering PCI serial device PMIO port 1016" should use hex (e.g. "0x3f8") for the base address number.

I fixed these two at https://github.com/acidanthera/OpenCorePkg/pull/331/commits/8b62364e43b758b29492ebd590468db808e5e449. Now that the patch works, let me go back to documents.

Docs/Configuration.tex
```diff
@@ -2573,6 +2573,17 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
   on all systems but those explicitly dedicated to scientific or media calculations.
   Only certain Xeon models typically benefit from the patch.
 
+\item
+  \texttt{CustomPciSerialDevice}\\
+  \textbf{Type}: \texttt{plist\ boolean}\\
+  \textbf{Failsafe}: \texttt{false}\\
+  \textbf{Requirement}: 10.7\\
```
PMheart, line 2580: Since 10.4-10.6 looks way more complex, I do not want to bother with those until requested.

PMheart: @joevt Maybe you know more details about this patch than I do. Would you please try writing the docs?

joevt: I don't know how to write OpenCore Docs.

Which parts do you need more explanation for?

**CustomPciSerialDevice**

This option causes the kernel (mostly pe_serial.c) to be patched to use the serial port specified by the Serial section. The patch changes all `in` and `out` accesses to 0x3f8-0x3ff (COM1 built-in serial port) to an address range described by the values in the Serial section.

There are two types of serial port devices each with a different calculation for the patched base address:
1) A PCI device specified by Serial→PciDeviceInfo. The base address for the address range is the value of the first I/O BAR (base address register) of the specified PCI device plus the value of Serial→RegisterBase (usually zero for a PCI device).
2) A non-PCI device (where Serial→PciDeviceInfo is a zero length path terminated by 0xff). The base address is the value of Serial→RegisterBase. For example, COM2 may exist at 0x2F8.

For the patch to be applied, the following must be true:
- Serial→Init must be true so that all the values in the Serial section (including PciDeviceInfo) are loaded into Pcd's.
- Serial→UseMmio must be false since the patch is only for devices controlled by port mapped I/O (PMIO).
- The calculated base address must not be the default 0x3f8.
The other values in the Serial section do not affect the kernel patch but will affect serial output from OpenCore when Debug→Target has bit 3 "Enable serial port logging" set.
- boot-args should have all the values required for the kernel to perform serial output. For example:
`debug=0x108 serial=3 serialbaud=115200`
- serialbaud in boot-args should match Serial→BaudRate if you're outputting to serial port in both OpenCore and the kernel.
- For kernel serial output to a non-default serial port device to work, you must use a method to stop a kext from matching the device. One method is to inject a codeless kext that matches the device using a higher IOProbeScore. A kext with code may be required if the normal matching methods cause the codeless kext to match multiple serial ports. For an example, search for `PCIeSerialDisable.kext/Contents/Info.plist` in https://github.com/acidanthera/bugtracker/issues/1954

PMheart: Thanks. I will try to write a more complete doc.

@vit9696 suggested we reduce module/section interdependency. Your points will be worth mentioning in the docs, but will not be how the patch works. Instead, we just apply it, though in most cases it does nothing (e.g. when it is an MMIO address which is greater than `UINT16_MAX`, it will not be applied; we have excluded 0x3F8 as well).

As for boot-args, can we actually set/append them automatically as we do for e.g. KASLR?

joevt: > Thanks. I will try to write a more complete doc.

What did I miss?

> @vit9696 suggested we reduce module/section interdependency. Your points will be worth mentioning in the docs, but will not be how the patch works.

What are the differences between how the patch works and what I described?

> Instead, we just apply it, though in most cases it does nothing (e.g. when it is an MMIO address which is greater than `UINT16_MAX`, it will not be applied; we have excluded 0x3F8 as well).

I don't think the PCI spec requires PMIO addresses to be limited to 16 bits - but the patches we're making are limited to 16 bits because the instructions we are replacing only load 16 bits. That's ok because x86 architecture can't access PMIO addresses beyond 64K and the firmware and OS will ensure that the I/O BARs are always only 16 bits.

> EDIT - Currently we only exclude 0x3F8. Shall we exclude 0x3F8-0x3FF as well?

No. 0x3F8 is the base address of the addresses we are replacing. We replace all occurrences of any address between the base address and the base address plus 7. If the base address matches the base address used by the kernel (0x3F8) then we don't do anything. If the base address is different than the base address used by the kernel (for example 0x3FA) then we would have to replace all addresses in the range 0x3F9..0x401. But the base address will never be between 0x3F9 and 0x3FF because it must be a multiple of the size.

> As for boot-args, can we actually set/append them automatically as we do for e.g. KASLR?

My idea was to patch `PE_parse_boot_arg` to hide the fact that we're overriding the boot-args separately from what is in the boot-args string.

Regarding the PMIO patches, I think we need to use the `+` operator instead of the `|` operator because we need to multiply the 3-bit register offset by `PcdSerialRegisterStride` like `SerialPortReadRegister` and `SerialPortWriteRegister` do.

`CommonPatches.c`:
```
STATIC
UINTN
mPmioRegisterBase = 0;    ///< To be set by PatchSetPciSerialDeviceRegisterBase()

SATIC
UINTN
mPmioRegisterStride = 1;    ///< To be set by PatchSetPciSerialDeviceRegisterBase()
```

`PatchSetPciSerialDeviceRegisterBase`:
```
...
  IN  UINTN  RegisterBase,
  IN  UINTN  RegisterStride
...
mPmioRegisterBase = RegisterBase;
mPmioRegisterStride = RegisterStride;
...
```

`PatchCustomPciSerialPmio`:
```
...
          UINTN Pmio = mPmioRegisterBase + (*WalkerPmio & 7U) * mPmioRegisterStride;
          WalkerPmio[0] = Pmio & 0xFFU;
          WalkerPmio[1] = Pmio >> 8U;
...
```

`PatchCustomPciSerialDevice`:
```
...
  if (mPmioRegisterBase != 0 && (mPmioRegisterBase != 0x3F8U || mPmioRegisterStride > 1) {
...
```

PMheart: > What did I miss?

No, rather I meant that the current doc written by me has almost every detail missing. I will handle this.

> What are the differences between how the patch works and what I described?

For instance, we should not check the value of Misc->Serial->Init to determine whether or not to enable the patch. Instead, we just apply it anyway.

Then regarding the patch, I will commit it later. Thanks for pointing it out.
As for patching `PE_parse_boot_arg`, that sounds crazy to be honest. I think we can just do it in a way that [OcAfterBootCompatLib](https://github.com/acidanthera/OpenCorePkg/tree/master/Library/OcAfterBootCompatLib) does, though I am not sure whether this is feasible. CC @vit9696

Also, in your original issue post, you talked about the boot-arg `msgbuf` - I think now the correct method is to use [DebugEnhancer](https://github.com/acidanthera/DebugEnhancer) instead.

joevt: > For instance, we should not check the value of Misc->Serial->Init to determine whether or not to enable the patch. Instead, we just apply it anyway.

True. What I meant was: if Serial->Init is not set to true, then the Pcd values don't get changed. If the Pcd values don't get changed, then they will have the default values. If they have the default values, then the patch won't get applied (because the default base address is 0x3f8 and the default stride is 1 and the default device path is 0xff and UseMMIO is default to false.

The PMIO patch should not be applied if RegisterAccessWidth is not 8 since the kernel does only 8 bit accesses to PMIO.

> Also, in your original issue post, you talked about the boot-arg `msgbuf` - I think now the correct method is to use [DebugEnhancer](https://github.com/acidanthera/DebugEnhancer) instead.

Thanks. I'll check it out... `msgbuf` is mentioned in Configuration.pdf. It is a boot-arg that is used in the kernel by bcd_init.c since 10.5.8 (but not 10.4.11 or earlier). The value is used by log_setsize. DebugEnhancer appears to use a value of 10MB for log_setsize. DebugEnhancer has two different patches for setting the size: one for when log_setsize exists, and one for when log_setsize does not exist. I don't know what versions of macOS the latter applies to (it should be commented in the code...)

Does DebugEnhancer require a boot-arg to enable it? It has a boot-arg to disable it `-dbgenhoff`. If DebugEnhancer is enabled, then using `msgbuf=` in boot-args has no effect. Maybe this should be mentioned in Configuration.pdf.

PMheart: Regarding `RegisterStride` - is there any function call to retrieve its value instead of reading it from Misc->Serial part?

Also, I think we should keep `mPmioRegisterBase` and `mPmioRegisterStride` UINT16 since this is what we are able to patch at most.

And we should also remove the `msgbuf` boot-args in docs, and recommend the use of DebugEnhancer instead since it is easier and as of 12.0 it is no longer possible to set this value without DebugEnhancer if I remember correctly. See https://github.com/acidanthera/bugtracker/issues/1916 CC @vit9696

joevt: > Regarding `RegisterStride` - is there any function call to retrieve its value instead of reading it from Misc->Serial part? Or maybe just call PcdGet functions?

I vote PcdGet. There's a function for RegisterBase because it requires device probing, enabling, and calculation. RegisterStride is much simpler.

> And we should also remove the `msgbuf` boot-args in docs, and recommend the use of DebugEnhancer instead since it is easier and as of 12.0 it is no longer possible to set this value without DebugEnhancer if I remember correctly. See [acidanthera/bugtracker#1916](https://github.com/acidanthera/bugtracker/issues/1916) CC @vit9696

Oh right. My xnu sources don't include macOS 12 yet so I didn't search it. Apple recently updated their open source repositories to include stuff from macOS 12 so I'll update that.

PMheart: Thanks. Regarding `RegisterStride` which is `UINT32`, how can we guarantee that the final patch (UINT16, 2 bytes) will not overflow? I don't think `>> 8U` is enough.

joevt: > Thanks. Regarding `RegisterStride` which is `UINT32`, how can we guarantee that the final patch (UINT16, 2 bytes) will not overflow? I don't think `>> 8U` is enough.

don't apply the PMIO patch if (UINTN)RegisterBase + 7 * RegisterStride > 0xffff

joevt: > OK. I updated your new patch. However, this time different values are replaced. Is it expected?

Different than what?

I think the log looks correct.

```
OCAK: Registering PCI serial device PMIO port 0x2008
OCAK: Registering PCI serial device register stride  0x4
OCAK: Matched PMIO serial register base <66 BA FF 03>
OCAK: Matched PMIO serial register base context out <EE>, patching register base
OCAK: Before register base patch <FF 03>
OCAK: After register base patch <24 20>
```

0x2008 + (0x3ff - 0x3f8)*4 = 0x2008 + 28 = 0x2008 + 0x1c = 0x2024.

PMheart: > Different than what?

Without stride (i.e. multiplication arithmetics) it should have always been replaced to 0x2008. But now that you said it is correct, it's all good.

joevt: Default stride is 1 so the new value for 0x3ff would be 0x2008 + 7 = 0x200f.
Stride cannot be 0.

PMheart: I see, then previously we were applying the patching with stride missing as if it was 0. Thanks for the fix.

joevt: I think previously were were applying the patch with stride assumed to be 1.
0x2008 | (0x3ff & 7) = 0x200f.

Docs/Sample.plist
```diff
@@ -1156,32 +1158,8 @@
 		</dict>
 		<key>Serial</key>
```
PMheart, line 1159: I removed the `Override` option here too. However, OcSerializeLib will complain now. Is it possible to make BOOLEAN optional as well?

PMheart, line 1159: I removed the `Override` option here too. However, OcSerializeLib will complain now. Is it possible to make BOOLEAN optional as well?

PMheart: https://github.com/acidanthera/OpenCorePkg/commit/8c965b6bd6813e3531f3d422edd16fdf97fca792

Docs/Configuration.tex
```diff
@@ -2573,6 +2573,32 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
...
+  As an alternative solution, a codeless kext \texttt{PCIeSerialDisable.kext} shown in the spoiler \texttt{PCIeSerialDisable.kext/Contents/Info.plist} at
+  \href{https://github.com/acidanthera/bugtracker/issues/1954}{acidanthera/bugtracker\#1954}, may also be used.
+
+  \emph{Note 2}: For this patch to be correctly applied, both \texttt{Init} and \texttt{Override}
+  must be enabled with all keys properly set in \texttt{Custom}, under section \texttt{Misc->Serial}.
+
+  \emph{Note 3}: Since this patch does not provide MMIO support, \texttt{UseMmio} under section \texttt{Misc->Serial->Custom} must be disabled.
+
+  \emph{Note 4}: The serial baud rate must be correctly set in both \texttt{BaudRate} under section \texttt{Misc->Serial->Custom} and via
+  \texttt{serialbaud=VALUE} boot argument, both of which should match against each other.
```
PMheart, line 2600: I will take a look whether we can set this boot argument automatically.

vit9696: Please do not. OpenCore should not interfere with such stuff.

Library/OcConfigurationLib/OcConfigurationLib.c
```diff
@@ -453,22 +455,29 @@ mMiscConfigurationSecuritySchema[] = {
   OC_SCHEMA_STRING_IN  ("Vault",                OC_GLOBAL_CONFIG, Misc.Security.Vault),
 };
 
+STATIC
+OC_SCHEMA
+mMiscConfigurationSerialCustomSchema[] = {
+  OC_SCHEMA_INTEGER_IN ("BaudRate",               OC_GLOBAL_CONFIG, Misc.Serial.Custom.BaudRate),
```
PMheart, line 461: Shall we replace all these with _OPT?

vit9696: No, only the dict itself. If the dict is specified, one must pass all.

PMheart:
@dakanji This makes sense. Thanks. Renamed.

Library/OcMainLib/OpenCoreKernelPatch.c
PMheart: UPDATE - Resolved by checking `Misc->Serial->Override`.

vit9696:
Looks good

PMheart:
@joevt @mhaeuser Do you have any other thoughts? If not, shall we merge?

joevt:
I have only the following two suggestions:

1) Regarding this:

>  The patch changes the PMIO register base address from the default \texttt{0x3F8-0x3FF} assumed by XNU to the actucal one used by the PCI serial card, such that the real PMIO address of the serial device will also be recognised by XNU.

I think this needs to be reworded so that base is not implied to be a range. Rather, base points to the beginning of a range.  Also, the patch also works for non PCI serial ports (these may appear as ACPI devices? I have to boot my hackintosh into macOS to be sure). Maybe something like this:

> The patch changes the PMIO register base address that the XNU kernel uses for serial input and output from that of the default built-in COM1 serial port \texttt{0x3F8} to the base address stored in the first IO BAR of a specified PCI device or to a specific base address (for example, \texttt{0x2F8} for COM2).

2) Regarding this note:

>   \emph{Note 4}: Since this patch does not provide MMIO support, \texttt{UseMmio} under section \texttt{Misc->Serial->Custom} must be disabled.

I think the user should know that this patch is for PMIO but they can use MMIO for OpenCore serial output and that there are boot-args to enable MMIO for XNU (but the addresses are not calculated automatically - they need to enter the correct address in boot-args themselves). Maybe this:

>   \emph{Note 4}: This patch is for PMIO support and is therefore not applied if \texttt{UseMmio} under section \texttt{Misc->Serial->Custom} is false. In that case, there are boot-args used in XNU's pe_serial.c that allow MMIO support.

PMheart:
1. I am ok with the change.

2. I do not think we should mention anything at source code level; plus I prefer having this at UseMmio instead.

dakanji:
> I do not think we should mention anything at source code level

Note that the docs already includes at least one source code level reference ... something about a Tianocore source file.
Seems better to provide a reference unless will be listing the args. On the other hand, what is to stop the source file from changing?

joevt:
We can just mention the other boot-args and then the user can search for the source file themselves to learn how they work and how they differ from each other and from the COM1 PMIO option and what version of macOS they work with. MMIO support exists but is not part of this patch so we can be vague about it. MMIO is fully supported for OpenCore but not XNU except for the boot-args which are sufficient for full capability.

>  there are boot-args `pcie_mmio_uart=` and `mmio_uart=` that allow the kernel to use MMIO for serial port access (in certain versions of macOS).

PMheart:
Thanks. I updated the descriptions at [Docs: Update quirk description](https://github.com/acidanthera/OpenCorePkg/pull/331/commits/c24b4bb873fb339e6fc7b318cd4fd8afe18e29f7).

PR #333, opened 2022-04-12
Utilities: Fixed buggy libFuzzer merge, which could skip new entries that increase coverage.

MikhailKrichanov:


User/Makefile
```diff
@@ -101,8 +101,8 @@ ifeq ($(FUZZ_MEM),)
 endif
 
 ifeq ($(COVERAGE),1)
-	CFLAGS  += -fprofile-arcs -ftest-coverage -D COVERAGE_TEST
-	LDFLAGS += --coverage
```
vit9696, line 105: This will not work on macOS, can we just not add `-Wl,-wrap,llvm_gcda_emit_arcs` for macOS?

vit9696, line 105: This will not work on macOS, can we just not add `-Wl,-wrap,llvm_gcda_emit_arcs` for macOS?

User/Makefile
```diff
@@ -101,8 +101,8 @@ ifeq ($(FUZZ_MEM),)
 endif
 
 ifeq ($(COVERAGE),1)
-	CFLAGS  += -fprofile-arcs -ftest-coverage -D COVERAGE_TEST
```
vit9696, line 104: What is the difference between these two flags and the new one? Is it supported by gcc?

vit9696, line 104: What is the difference between these two flags and the new one?

MikhailKrichanov: --coverage is just a shorter synonym for those two options and yes, it is supported by gcc.
https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html

PR #338, opened 2022-04-20
OcAfterBootCompatLib: Add `DisableExternalAppleGpu` quirk to disable external GPUs on real Macs

PMheart:
#### Help Wanted
This does not work as reported at https://github.com/acidanthera/bugtracker/issues/928#issuecomment-1105446248. At this stage debug printing won't work either, making it harder to trace the correct Status code. One possible strategy suggested by @vit9696 is to brute-force it by [`DirectResetCold`](https://github.com/acidanthera/OpenCorePkg/blob/ef2db45050c4aed6aa2e93d7c00df45706ab4e13/Library/OcDeviceMiscLib/DirectReset.c) (i.e. check if Status == something, then `DirectResetCold`). However, this takes time. If someone wants to help with this process, I'd really appreciate.

closes https://github.com/acidanthera/bugtracker/issues/928

TODO:
- [ ] Rebuild pdf after merging


khronokernel:
### MacBookPro8,2 Testing
Tested using a MacBookPro8,2 (2011 15") unit with commit 98c923c305fd4a5881bfdd47b3e0d9d9e997f580, OpenCore will stall out after executing `if (BootCompat->Settings.DisableExternalAppleGpu) {}` code block. This is seen after `End RandomSeed`.

Setting the block to only set the variable still stalls the system in the same way:

```c
  if (BootCompat->Settings.DisableExternalAppleGpu) {
    DisabledGpu = 1U;
    Status = gRT->SetVariable (
      L"gpu-power-prefs",
      &gApplePersonalizationVariableGuid,
      EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_RUNTIME_ACCESS | EFI_VARIABLE_NON_VOLATILE,
      sizeof (DisabledGpu),
      &DisabledGpu
      );
  }
```

Image reference for stall:
Screen Shot 2022-04-09 at 3 27 10 PM

mikebeaton:
@PMheart Can put exit status in another var, assuming it is even exiting? And can put something else in another var before setting, and something other else in that var if setting exits, to check that?

mikebeaton:
@khronokernel - probably a stupid suggestion, but what happens if you remove the EFI_VARIABLE_BOOTSERVICE_ACCESS bit from the mask sent to SetVariable?

mikebeaton:
This should be replaced by https://github.com/acidanthera/OpenCorePkg/tree/unprotect if testing works out.

mikebeaton:
Having gone round the houses, I'm now back to thinking that this commit should have been the right way to do it all along, except that it doesn't work! I'll try to PM yourself, pmheart, and Vit to try to work out what the heck we could try next.

mikebeaton:
@PMheart - Hi! Is it worth keeping this PR open, for now, given that there remains an open [bugtracker issue](https://github.com/acidanthera/bugtracker/issues/928) on this, plus my own [failed attempt](https://github.com/acidanthera/OpenCorePkg/pull/390) to do something similar, throwing up further issues, it seems.

PR #342, opened 2022-04-27
Update SSDT-SBUS-MCHC.dsl

Antigo1:


andreszerocross:
Not all ACPI use PC00 as default, most of them are PCI0 not PC00.

PC00 just for Intel 6xx chipset

khronokernel:
ACPI Source files are simply for reference, each user will need to adjust and validate their configuration to the ACPI tables present in the system's firmware. These are not grab and go files

PR #343, opened 2022-04-27
Code refactoring

dreamwhite:
As discussed in issue acidanthera/bugtracker#1997:

- Improved f-strings support
- Single line statements
- Proper file writing statements


dreamwhite:
As macOS Monterey 12.3.1 comes bundled with Python3.10, you'll have to change the shebang to use python3.10

AppleModels/update_generated.py
```diff
@@ -119,32 +117,29 @@ def validate_products(db, dbpd):
...
   Export yaml database to MacInfoLib format.
   TODO: use jinja2?
-  """
+  '''
 
   with open(path, 'w') as fh:
-    print('// DO NOT EDIT! This is an autogenerated file.', file=fh)
-    print('#include "MacInfoInternal.h"', file=fh)
-    print('CONST MAC_INFO_INTERNAL_ENTRY gMacInfoModels[] = {', file=fh)
+    fh.write('// DO NOT EDIT! This is an autogenerated file.')
```
vit9696, line 131: Does this add newlines?

AppleModels/update_generated.py
```diff
@@ -1,12 +1,10 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python3.10
```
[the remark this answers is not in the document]
dreamwhite: Well, since I'm using match-case syntax, that was introduced from Python 3.10, I've changed the shebang 😅

AppleModels/update_generated.py
dreamwhite: Oh, I forgot to add the newline character. I'll update the code as soon as possible

dreamwhite: Commit 5aa2d4ec should have fixed what you wanted

AppleModels/update_generated.py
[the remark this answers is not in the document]
dreamwhite: The following files need Python3.10 to correctly work (as already explained before):

- `Debug/Scripts/common_uefi.py`
- `Debug/Scripts/gdb_uefi.py`
- `Debug/Scripts/lldb_uefi.py`
- `Utilities/macrecovery/macrecovery.py`

khronokernel:
> As macOS Monterey 12.3.1 comes bundled with Python3.10

To clarify, this is not true. macOS's `python3` is just a hollow binary that reroutes to Xcode Tools, users need to manually install them.

Since many may therefore be using python.org binaries (instead of installing the rest of the bloat in Xcode Tools), I would advise against raising the requirement to python 3.10 unless absolutely necessary. Python 3.7 will be at end of life in June 2023, 3.8 and 3.9 are much longer.

If the only benefit is switch case in Python 3.10, I believe it would be best to retain at least the last major release instead (ie. 3.9) and leave the old switch case code as-is.

dreamwhite:
Commit cc5c283 should have fixed as you asked

vit9696:
Python codestyle does not use tabs from what I know, but rather 4 spaces. To make the changes in this commit useful and future-prone, please introduce prospector support in the CI: https://prospector.landscape.io/en/master/

AppleModels/update_generated.py
vit9696: How is this change better than what we had though? Makes me feel it is a useless change that makes things less readable.

dreamwhite: Mmmh, you're right. I'm gonna revert to the old write statements

PMheart:
I introduced prospector at https://github.com/acidanthera/ocbuild/tree/prospector, and the PR https://github.com/acidanthera/ocbuild/pull/17. Please take a look and try to integrate it here by updating workflows.

dreamwhite:
> I introduced prospector at https://github.com/acidanthera/ocbuild/tree/prospector, and the PR [acidanthera/ocbuild#17](https://github.com/acidanthera/ocbuild/pull/17). Please take a look and try to integrate it here by updating workflows.

Thank you so much for replying me and sorry for my late reply but I'm kinda busy in this period.
Anyways, how should I exactly update workflows? I've never used prospector before, but from what I'm seeing [here](https://github.com/acidanthera/ocbuild/pull/17/commits/62e676ae78a99f5fa9cc8efbbdb888085892e291), I have to run prospector for each Python script/folder that contains `.py` scripts. Am I wrong?
If so, shouldn't I wait for acidanthera/ocbuild#17 PR to be merged onto master branch, and then update this PR?

PMheart:
Hello! Sorry for the late response as well. And yes, https://github.com/acidanthera/ocbuild/pull/17 is merged. You may refer to https://github.com/acidanthera/ocbuild/blob/572c780a542c1181463dd1f16b404593e2925527/.github/workflows/main.yml#L28 and create a similar workflow here in OpenCorePkg.

dreamwhite:
> Hello! Sorry for the late response as well. And yes, [acidanthera/ocbuild#17](https://github.com/acidanthera/ocbuild/pull/17) is merged. You may refer to https://github.com/acidanthera/ocbuild/blob/572c780a542c1181463dd1f16b404593e2925527/.github/workflows/main.yml#L28 and create a similar workflow here in OpenCorePkg.

Alright. Should I adapt the workflow to be executed on each script, or just pass the whole project folder, with a view of future additional Python3 scripts?

PMheart:
The second is better I believe.

PMheart:
I just introduced prospector, and I believe that there will be more things to fix.

dreamwhite:
Sorry for the late reply. About "There will be more things to fix" you mean fixing the warnings that prospector raised?

PMheart:
Hello! No worries for the late response. And yes, that's what I meant. Though, it's very weird that now prospector exited immediately after downloading the profile. I will look into it.

dreamwhite:
> Hello! No worries for the late response. And yes, that's what I meant. ~Though, it's very weird that now prospector exited immediately after downloading the profile. I will look into it.~
> 
> UPDATE: The error was reported by prospector, so it was correct.
> 
> UPDATE 2: Now it is probably time to either fix the code at our side, or disable the errors/warnings in https://github.com/acidanthera/ocbuild/blob/master/prospector/profile.yml.

Okay so after sorting all the raised errors I think we can disable the following warnings:

- pycodestile:
    - E302 / expected 2 blank lines, found 1 (col 1)
    - E221 / multiple spaces before operator (check [here](https://github.com/acidanthera/OpenCorePkg/blob/59fd52487970a66def8b7fac71b10859902f3f55/AppleModels/update_products.py#L37); I think we should exclude this kind of warning, since it's for better readability, but it's up to your choice)
    - W191 / indentation contains tabs (check [here](https://github.com/acidanthera/OpenCorePkg/blob/59fd52487970a66def8b7fac71b10859902f3f55/AppleModels/update_products.py#L53))
    - E128 / same as before

I'll update this comment asap

PMheart:
In my opinion, better to follow E302; and please convert all tabs to spaces to obey W191 and/or E128.

vit9696:
I do not think we should disable any of these warnings. They are common to all production python code.

dreamwhite:
> In my opinion, better to follow E302; and please convert all tabs to spaces to obey W191 and/or E128.

Okay, E302 warnings should be fixed. I'll wait for the workflow to be run, so I can check again the prospector artifacts and fix the other issues.

dreamwhite:
I think warning `pycodestyle: E128 / continuation line under-indented for visual indent (col 17)` can be ignored, since for better readability. Let me know your thoughts

dreamwhite:
I've updated just the first file (AppleModels/update_generated.py) to fix the first errors. If it's fine, then I'll proceed with the other files

AppleModels/update_generated.py
```diff
@@ -1,318 +1,296 @@
...
+        if len(info['SystemSerialNumber']) == 12:
+            models = [info['BoardProduct']] if not isinstance(info['BoardProduct'], list) else info['BoardProduct']
 
-      for model in models:
-        if info['MaximumOSVersion'] is None:
-          l[model] = 'latest'
-        else:
-          l[model] = info['MaximumOSVersion']
+            for model in models:
+                l[model] = 'latest' if info['MaximumOSVersion'] is None else info['MaximumOSVersion']
```
vit9696, line 280: if not info['MaximumOSVersion']?

vit9696:
Well, I am not sure I like print -> .write change, it feels like it makes files less readable to be honest.

AppleModels/update_generated.py
dreamwhite: Oops! I missed it. Gonna fix it

dreamwhite:
> Well, I am not sure I like print -> .write change, it feels like it makes files less readable to be honest. Why did you do this?

Well, there's not practical difference. I thought it was useless opting for `print(..., file=fh)` since newlines aren't explicit as with `fh.write`. I temporarily reverted `fh.write` statements with `print(..., file=fh)`. Let me know if it's fine or not

dreamwhite:
Okay so, about `AppleModels/update_generated.py` script there are still the following warnings:

```
pylint: redefined-outer-name / Redefining name 'db' from outer scope (line 286) (col 4)
AppleModels/update_generated.py
  Line: 34
    pylint: redefined-outer-name / Redefining name 'db' from outer scope (line 286) (col 4)
    pylint: invalid-name / Variable name "db" doesn't conform to snake_case naming style (col 4)
  Line: 36
    pylint: unused-variable / Unused variable 'dirs' (col 14)
  Line: 39
    pylint: unspecified-encoding / Using open without explicitly specifying an encoding (col 17)
    pylint: invalid-name / Variable name "fh" doesn't conform to snake_case naming style (col 36)
  Line: 41
    pylint: invalid-name / Variable name "r" doesn't conform to snake_case naming style (col 20)
  Line: 46
    pylint: invalid-name / Variable name "e" doesn't conform to snake_case naming style (col 16)
  Line: 58
    pylint: redefined-outer-name / Redefining name 'db' from outer scope (line 286) (col 20)
    pylint: invalid-name / Argument name "db" doesn't conform to snake_case naming style (col 20)
  Line: 64
    pylint: invalid-name / Variable name "pp" doesn't conform to snake_case naming style (col 8)
  Line: 67
    pylint: invalid-name / Variable name "p" doesn't conform to snake_case naming style (col 12)
  Line: 71
    pycodestyle: E701 / multiple statements on one line (colon) (col 29)
  Line: 79
    pycodestyle: E701 / multiple statements on one line (colon) (col 29)
  Line: 85
    pylint: redefined-outer-name / Redefining name 'db' from outer scope (line 286) (col 22)
    pylint: invalid-name / Argument name "db" doesn't conform to snake_case naming style (col 22)
    pylint: too-many-branches / Too many branches (14/12)
  Line: 121
    pylint: consider-using-dict-items / Consider iterating with .items() (col 8)
  Line: 129
    pylint: redefined-outer-name / Redefining name 'db' from outer scope (line 286) (col 25)
    pylint: invalid-name / Argument name "db" doesn't conform to snake_case naming style (col 25)
  Line: 135
    pylint: unspecified-encoding / Using open without explicitly specifying an encoding (col 9)
    pylint: invalid-name / Variable name "fh" doesn't conform to snake_case naming style (col 28)
  Line: 147
    pylint: consider-using-f-string / Formatting a regular string which could be a f-string (col 18)
  Line: 148
    pycodestyle: E128 / continuation line under-indented for visual indent (col 17)
Same pycodestyle warning till Line 171
Line: 203
    pylint: redefined-outer-name / Redefining name 'db' from outer scope (line 286) (col 24)
    pylint: invalid-name / Argument name "db" doesn't conform to snake_case naming style (col 24)
  Line: 209
    pylint: unspecified-encoding / Using open without explicitly specifying an encoding (col 9)
    pylint: invalid-name / Variable name "fh" doesn't conform to snake_case naming style (col 28)
  Line: 272
    pylint: redefined-outer-name / Redefining name 'db' from outer scope (line 286) (col 22)
    pylint: invalid-name / Argument name "db" doesn't conform to snake_case naming style (col 22)
  Line: 273
    pycodestyle: W293 / blank line contains whitespace (col 1)
  Line: 274
    pycodestyle: E741 / ambiguous variable name 'l' (col 5)
    pylint: invalid-name / Variable name "l" doesn't conform to snake_case naming style (col 4)
  Line: 280
    pycodestyle: W291 / trailing whitespace (col 98)
  Line: 282
    pylint: unspecified-encoding / Using open without explicitly specifying an encoding (col 9)
    pylint: invalid-name / Variable name "fh" doesn't conform to snake_case naming style (col 30)
  Line: 285
    pycodestyle: E305 / expected 2 blank lines after class or function definition, found 1 (col 1)
```

I'd like to suppress `E128` warning, since we use continuation line for better readability (check below screenshot)

image

Regarding line 147:

```
            sb_model = info.get('AppleModelId')
            sb_model = f"'{sb_model.lower()}'" if sb_model else 'NULL'
```

we can optimize it with 

```
sb_model = f"'{info.get('AppleModelId').lower()}"' if info.get('AppleModelId') else 'NULL'
```

BUT the readability is definitely worse. Let me know your thoughts on this

Regarding line 122:

```
if len(to_add) > 0:
```

I'd rather rewrite it as `if len(to_add.items()) > 0`

Regarding line 282:

```    pylint: unspecified-encoding / Using open without explicitly specifying an encoding (col 9)```

warning can be suppressed by explicitly specifying an encoding (such as `UTF-8`)

As [Python Docs](https://docs.python.org/3.9/library/functions.html#open) says, we should pass a `encoding='UTF-8` parameter in `open` statement.

Let me know your thoughts on that.

Also, since I don't wanna clog this PR thread with lot of those warning that I'll fix/suppress, may I ask a way to contact one of you devs?
Greetings

**Comment by vit9696** [comment]
Regarding E128, indenting one space right will not make any difference. Do not think it makes sense to disable the warning.

Regarding line 147, it is about `print`, not the line you mentioned.

Regarding line 122, it is about `for key, value in dict.items()`, not what you mean.

Regarding UTF-8 it is ok to specify it as an argument.

It is ok to post things here for easier tracking, but you could ping us through matrix if you want.

**Comment by dreamwhite** [comment]
Thank you so much for your rely. May I ask you your matrix username as well as the server (like matrix.org, idk)

**Comment by vit9696** [comment]
Can send you a e-mail/PM somewhere.

**Comment by dreamwhite** [comment]
> Can send you a e-mail/PM somewhere.

Sure, I have telegram (@dreamwhitedev) and an e-mail address (ivancafiero@gmail.com).

**Comment by dreamwhite** [comment]
Hi there, sorry for the late reply but it has been a very busy week.
I fixed pretty every warning inside `update_generated.py` script except for the f-string one, since some variables are formatted differently.
We could leave it for the moment and proceed with the other scripts. Let me know

**Comment by dreamwhite** [comment]
image

Regarding instead `update_products.py` instead, there are those warnings that can be easily suppressed, since all the local variables are used

**Comment by dreamwhite** [comment]
I'd like to disable the following pylint warnings:

- `wildcard-import`
- `too-many-public-methods`
- `undefined-variable`

**Comment by dreamwhite** [comment]
I'm attaching my profile.yml:
[profile.yml.zip](https://github.com/acidanthera/OpenCorePkg/files/8748330/profile.yml.zip)

**Comment by PMheart** [comment]
Hello! I am sorry for the late response as well; have been extremely busy over this week too. And I think we should remove all undefined variables, so that `undefined-variable` does not have to be disabled. I am ok with the other changes.

**Comment by dreamwhite** [comment]
image

Okay so basically the reason why I added `undefined-variable` is because of `Debug/Scripts/*.py`. I haven't found yet a way to test them :/ How can I do it?

**Comment by PMheart** [comment]
Hmm, ok. I think there should be an argument for prospector to include it, but I am not sure. If it cannot be done this way, then I am ok with `undefined-variable`.

**Comment by dreamwhite** [comment]
About sheel scripts, is `ERROR: Model U9B is used in DataBase but not present in Products!` an error cause by missing the Model in Products? Eventually how can we fix it?

**Comment by vit9696** [comment]
Just do not change AppleModels/Products.zjson in my opinion.

**Comment by dreamwhite** [comment]
@PMheart at this point of the code restyling I'd say that it's kinda finished. What do you think about it? Is it ready to go in master branch or it needs more documentation?

**Comment by PMheart** [comment]
Hmm, CI still fails. Before merging we should at least fix the errors, or disable some checks?

**Comment by vit9696** [comment]
I disagree with disabling the following warnings:

- redefined-outer-name
- invalid-name
- consider-using-f-string
- wild-card-import (it is very dangerous as it can transitively import other methods, at worst one can add local comments to suppress particular warnings)
- undefined-variable (can be suppressed with a comment for a local file/string)
- no-self-use (need particular examples)
- simplifiable-if-expression
- expression-not-assigned
- global-statements

At least without particular examples, and why the warning is harmless.

**Comment by dreamwhite** [comment]
Okay so:

About

- `redefined-outer-name` it actually warns since it's both defined in `if __name__ == '__main__':` and line 33. At this point I think it can actually be ditched or converted to a global variable.
- `invalid-name` is actually due to variable names, such as `db`,`r`,`fh` and many others. I'll rename them as soon as possible if required
- `consider-using-f-string` (line 149 of `AppleModels/update_generated.py`) is due to string formatting. What worried me by a lot is like `.FirmwareFeatures = 0x%XULL,\n'`. Is `ULL` actually a string or what? If it's a string then I won't have any kind of problems fixing it

- `wild-card-import`

What kind of local comments would you like to add? In any case I think we can replace the import with something like: `from common_uefi import UefiMisc`.

- `undefined-variable` is actually blamed both by `gdb` and `ida_bytes` libraries that for obvious reasons GitHub Actions can't test. I don't know if you planned to adapt `workflows.yml` to also use `gdb`.

(P.S. I wrongly type `_` in `macrecovery.py` line 296. Added to my fix list

- `no-self-use` is actually due to the following functions:

- `gdb_uefi.py`:

```python
...
 def lookup_type(self, typename):
        return gdb.lookup_type(typename)

    #
    # Returns gdb.Type for a pointer to a type.
    #

    def ptype(self, typename):
        return gdb.lookup_type(typename).pointer()

    #
    # Computes CRC32 on an array of data.
    #

    def crc32(self, data):
        return binascii.crc32(data) & 0xFFFFFFFF

...
 def get_sym_cmd(self, file, orgbase, sections, macho, fallack_base):
        cmd = f'add-symbol-file {file}'

        # Fallback case, no sections, just load .text.
        if not sections.get('.text') or not sections.get('.data'):
            cmd += f' 0x{fallack_base:x}'
            return cmd

        cmd += f" 0x{int(orgbase):x}{sections['.text']}"

        if not macho or not os.path.exists(file):
            # Another fallback, try to load data at least.
            cmd += f" -s .data 0x{int(orgbase) + sections['.data']:x}"
            return cmd

        # 1. Parse Mach-O.
        # FIXME: We should not rely on otool really.
        commands = subprocess.check_output(['otool', '-l', file])
        try:
            lines = commands.decode('utf-8').split('\n')
        except Exception:
            lines = commands.split('\n')
        in_sect = False
        machsections = {}
        for line in lines:
            line = line.strip()
            if line.startswith('Section'):
                in_sect = True
                sectname = None
                segname = None
            elif in_sect:
                if line.startswith('sectname'):
                    sectname = line.split()[1]
                elif line.startswith('segname'):
                    segname = line.split()[1]
                elif line.startswith('addr'):
                    machsections[segname + '.' + sectname] = int(line.split()[1], base=16)
                    in_sect = False

        # 2. Convert section names to gdb sections.
        mapping = {
            '__TEXT.__cstring':         '.cstring',
            '__TEXT.__const':           '.const',
            '__TEXT.__ustring':         '__TEXT.__ustring',
            '__DATA.__const':           '.const_data',
            '__DATA.__data':            '.data',
            '__DATA.__bss':             '.bss',
            '__DATA.__common':          '__DATA.__common',
            # FIXME: These should not be loadable, but gdb still loads them :/
            # '__DWARF.__apple_names':    '__DWARF.__apple_names',
            # '__DWARF.__apple_namespac': '__DWARF.__apple_namespac',
            # '__DWARF.__apple_types':    '__DWARF.__apple_types',
            # '__DWARF.__apple_objc':     '__DWARF.__apple_objc',
        }

        # 3. Rebase.
        for entry in mapping.items():
            if machsections.get(entry):
                cmd += f' -s {mapping[entry]} 0x{int(orgbase) + machsections[entry]:x}'

        return cmd

...

def usage(self):
        print('Usage: reload-uefi [-o] [/path/to/GdbSyms.dll]')
```

```python
...
def crc32(self, data):
        return binascii.crc32(data) & 0xFFFFFFFF
...
    def set_field(self, value, field_name, data):
        member = value.GetChildMemberWithName(field_name)
        data = lldb.SBData.CreateDataFromInt(data, size=member.GetByteSize())
        error = lldb.SBError()
        member.SetData(data, error)
...
 def get_sym_cmd(self, filename, orgbase, *_):
        if filename.endswith('.pdb'):
            dll_file = filename.replace('.pdb', '.dll')
            module_cmd = f'target modules add -s {filename} {dll_file}'
        else:
            dll_file = filename
            module_cmd = f'target modules add {filename}'
        map_cmd = f'target modules load -f {dll_file} -s 0x{orgbase:X}'
        return(module_cmd, map_cmd)
...
def get_short_help(self):
        return 'Usage: reload-uefi [-o] [/path/to/GdbSyms.dll]'

    def get_long_help(self):
        return 'Usage: reload-uefi [-o] [/path/to/GdbSyms.dll]'

    def usage(self):
        print(self.get_short_help())
```

I never experienced before this kind of warning, but I'll deep investigate on it and try to find the wisest way to fix it. I'll let you know

- `simplifiable-if-expression` is probably due to the short-if syntax itself. Example:

```python
    def pe_is_64(self, pe_headers):
        magic = pe_headers.GetValueForExpressionPath('.Pe32.OptionalHeader.Magic').GetValueAsUnsigned()
        return True if magic == self.PE32PLUS_MAGIC else False
```

pylint actually suggest to replace the if-expression with `test` but I don't what the heck is even `test`. Still to deep investigate on it

- `expression-not-assigned`. Example:

```python
            syms.append(self.get_sym_cmd(real_sym, orgbase, sections, macho, base)) if real_sym else print(f'No symbol file {sym_name}')

```

I'd suppress this warning since it actually appends data to `syms`. Do you agree on suppressing it?

- `global-statements`. Example:

```python
g_target_definition = None


def get_target_definition(triple):
    global g_target_definition
```

I'm actually confused about how `g_target_definition` variable is used: firstly it's `None` type, then it's a `dict` type.
At this point I'd rewrite the code like:

```python
def get_target_definition(triple):
    g_target_definition = {}
    offset = 0
    for reg_info in x86_64_register_infos:
        reg_name = reg_info['name']

        # Only fill in the offset if there is no 'slice' in the register
        # info
        if 'slice' not in reg_info and 'composite' not in reg_info:
            reg_info['offset'] = offset
            offset += reg_info['bitsize'] // 8

        # Set the GCC/DWARF register number for this register if it has one
        reg_num = get_reg_num(name_to_gcc_dwarf_regnum, reg_name)
        if reg_num != LLDB_INVALID_REGNUM:
            reg_info['gcc'] = reg_num
            reg_info['dwarf'] = reg_num

        # Set the generic register number for this register if it has one
        reg_num = get_reg_num(name_to_generic_regnum, reg_name)
        if reg_num != LLDB_INVALID_REGNUM:
            reg_info['generic'] = reg_num

        # Set the GDB register number for this register if it has one
        reg_num = get_reg_num(name_to_gdb_regnum, reg_name)
        if reg_num != LLDB_INVALID_REGNUM:
            reg_info['gdb'] = reg_num

    g_target_definition['sets'] = [
        'General Purpose Registers',
        'Floating Point Registers']
    g_target_definition['registers'] = x86_64_register_infos
    g_target_definition[
        'host-info'] = {'triple': triple, 'endian': eByteOrderLittle}
    g_target_definition['g-packet-size'] = offset
    return g_target_definition
```
Do you agree?

# TL;DR

- `redefined-outer-name` can be easily fixed by removing the outer definition since it's just used in the script itself
- `invalid-name` need to get your approval on renaming the variables and eventually what's the preferred naming style
- `consider-using-f-string` if `ULL` is just a string and nothing more I can fix it easily
- `wild-card-import` can be fixed by importing just the needed classes/methods
- `undefined-variable` due to missing libraries (`gdb` and `ida_bytes`)
- `no-self-use` needs more investigation on it
- `simplifiable-if-expression` probably not going to do nothing. `short-if` syntax is causing this stuff. Eventually can revert it to normal `if` syntax
- `expression-not-assigned` same as `simplifiable-if-expression`
- `global-statements` can easily be fixed

dreamwhite:
> pylint actually suggest to replace the if-expression with `test` but I don't what the heck is even `test`. Still to deep investigate on it

Founded the fix:

just replace the `return True if variable == value else False` with `return variable == value`

vit9696:
Thank you!

- I think it is ok to disable invalid-name, too many useless examples to fix.
- `ULL` is indeed just a string. I.e. it is 0x HEX_VALUE ULL to print 0x12345ULL. ULL is an unsigned long long numeric suffix from C.
- disable undefined-variable within the files affected (with comments).
- no-self-use could probably be disabled within particular files, since it is few (with comments).
- expression-not-assigned is caused by unreadable code, you should rewrite it in multiple lines.

The rest I believe you answered yourself ^_^

dreamwhite:
Thank you so much for your reply. Actually I've never used prospector before so I'm asking you: how can I disable warnings within a certain file? What kind of comment should I write?

vit9696:
You can also add such comments after the problematic line, e.g. `python line # pylint: …` or within the block. This allows for more granular disablement. See pylint docs for more details.

dreamwhite:
Alright, I should have fixed stuff. I don't like the way I added the ignore warnings comments, but it works. If you have any advice let me know :)

Debug/Scripts/gdb_uefi.py
```diff
@@ -42,6 +42,8 @@
 
 
 class ReloadUefi(gdb.Command):
+    # pylint: disable=undefined-variable
```
vit9696, line 45: I believe you are best to disable undefined-variable and no-self-use warnings globally on this file.

Debug/Scripts/lldb_uefi.py
```diff
@@ -75,6 +76,7 @@ def typed_ptr(self, typename, address):
     #
 
     def crc32(self, data):
+        # pylint: disable=no-self-use
```
vit9696, line 79: Same here.

Debug/Scripts/x86_64_target_definition.py
```diff
@@ -178,6 +180,7 @@
 }
 
 name_to_generic_regnum = {
+    # pylint: disable=undefined-variable
```
vit9696, line 183: Same here.

Utilities/macserial/cccdump.py
```diff
@@ -8,13 +8,15 @@
 
 
 def readaddr(ea):
+    # pylint: disable=undefined-variable
```
vit9696, line 11: Same here.

dreamwhite:
Alright, is it fine?

vit9696:
Thank you! Please also provide a PR for ocbuild to disable the warnings, so that CI passes.

vit9696:
Please also revert AppleModels/Products.zjson changes.

dreamwhite:
Alright, I:

- reverted `AppleModels/Products.zjson` (I wrongly deleted it during my tests. Added the one from master branch)
- created a PR on acidanthera/ocbuild as you suggested for passing GitHub CI
- fetched latest changes from acidanthera/OpenCorePkg master branch with uncrustify commits

dreamwhite:
Seems like `Library/OcMacInfoLib/AutoGenerated.c` file is modified: (sorry for the formatting) [link](https://github.com/dreamwhite/OpenCorePkg/runs/7326333650?check_suite_focus=true#step:6:67)

dreamwhite:
Here I am, both `Shell Scripts` and `Python Scripts` workflows pass. [Proof](https://github.com/dreamwhite/OpenCorePkg/actions/runs/2665911923)

 I temporarily changed `analyze.yml` to point at my repo. Will revert it when acidanthera/ocbuild#18 will get merged

dhinakg (approved):
LGTM

.github/workflows/analyze.yml
```diff
@@ -59,6 +59,30 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install Dependencies
+        run: |
+          python3 -m pip install prospector
+
+      - name: Run prospector
+        run: |
+          /usr/bin/curl -L https://raw.githubusercontent.com/dreamwhite/ocbuild/master/prospector/profile.yml -o profile.yml || exit 1
```
vit9696, line 76: Merged into ocbuild, please update the link. Thank you!

Debug/Scripts/x86_64_target_definition.py
```diff
@@ -196,11 +198,6 @@ def get_reg_num(reg_num_dict, reg_name):
     return LLDB_INVALID_REGNUM
 
 
-def get_reg_num(reg_num_dict, reg_name):
```
vit9696, line 199: Why was that removed, sorry?

Debug/Scripts/x86_64_target_definition.py
```diff
@@ -731,53 +728,50 @@ def get_reg_num(reg_num_dict, reg_name):
...
 
-def get_dynamic_setting(target, setting_name):
+def get_dynamic_setting(_, setting_name):
     if setting_name == 'gdb-server-target-definition':
         triple = os.getenv('EFI_TRIPLE')
         if not triple:
             print('Cannot find requested triple!\nHint: set EFI_TRIPLE variable.')
             sys.exit(1)
         return get_target_definition(triple)
+    return None
```
vit9696, line 777: Should not be None.

Utilities/macserial/cccdump.py
```diff
@@ -1,96 +1,99 @@
+# pylint: disable=undefined-variable,import-error,too-many-boolean-expressions
```
vit9696, line 1: That code is for python 2 actually. Though maybe now it is ok to upgrade, IDA supports python 3 today.

Debug/Scripts/x86_64_target_definition.py
dreamwhite: Weird! I think I missed it but in my master branch, `get_reg_num` is there: [source](https://github.com/dreamwhite/OpenCorePkg/blob/master/Debug/Scripts/x86_64_target_definition.py)

Utilities/macserial/cccdump.py
dreamwhite: You mean the whole script, right? It should be Py3 compliant, despite I couldn't test it (`ida_bytes` library missing on my system). How can I check it? changed stuff is just `print` statements, `l` variable name changed to `length` and `is None` syntax where needed.
Is there anything else I can do to test it? 
Sorry but I never used that script and have no clue on how to properly test it 😅

Debug/Scripts/x86_64_target_definition.py
dreamwhite: The issue there is that prospector blames about this function not returning anything if the `if` code isn't executed. What should it return if `setting_name != 'gdb-server-target-definition'`?

Utilities/macserial/cccdump.py
vit9696: Leave as is, I guess I will fix it if I need it again.

dreamwhite: Oh sure. Can I mark this conversation as solved then?

dreamwhite:
@PMheart sorry for tagging you. Since acidanthera/ocbuild#18 got merged, can I safely delete my fork or is it still required?

PMheart:
@dreamwhite Yes. Now you can just use the link on master.

Debug/Scripts/x86_64_target_definition.py
vit9696: Well, it is clearly deleted in the PR though.

dreamwhite: Ouch, I'm kinda confused at this point. PR says the function is not there, but the master branch has it 😅

dreamwhite:
Hi there, sorry for bumping this PR. Not my intention stressing you devs. Is there any news or additional edit should I apply?

dreamwhite:
> @dreamwhite, this patch breaks gdb debugging. Was this tested?

Hi, sorry for the long delay but I was on holiday and internet connection sucked. May I ask you what part of gdb debugging is broken?

vit9696:
I believe I fixed it (at least mostly) in a subsequent commit, but the one described here: https://github.com/acidanthera/OpenCorePkg/tree/master/Debug

The scripts did not even load properly when e.g. calling `reload-uefi`, let alone breakpoints and value pretty printing. Same for lldb.

PR #345, opened 2022-04-30
Add Facebook Hackintosh group

Core-i99:
This PR adds the Facebook Hackintosh group to the OpenCore discussion forums.

A couple of rules from this group: (to make clear no macOS distro's or anything like that are supported)
- No support for distros (Niresh, Hackintosh Zone, HackintoshShop and Tonymacx86)
- Posts about those macOS distros and their tools will be removed.
- No support for prebuild EFI's
- Posts in any other language than English will be removed, to make every post understandable for everyone.

These rules are in an announcement/pinned post  everyone can see it once they joined the group.

As of the time of writing this group has 86.4K members.

(And yes, there's a spelling fault in the link to this group. The actual name of this group is correct)


Core-i99:
Why is this closed?

Andrey1970AppleLife:
We don't need social media.

PR #346, opened 2022-05-02
OcAudio: Convert from int ids to string ids

mikebeaton:
I've had a chance to do a bit of work on https://github.com/acidanthera/bugtracker/issues/1777.

This is the first part, upgrading OC_AUDIO_PROTOCOL so that sound files are specified by string rather than int, to make it possible to implement system actions with support for audio assist in BootEntryProtocol next.

### DIFF
--- a/Include/Apple/Protocol/AppleVoiceOver.h
+++ b/Include/Apple/Protocol/AppleVoiceOver.h
@@ -41,6 +41,7 @@ typedef struct APPLE_VOICE_OVER_AUDIO_PROTOCOL_ APPLE_VOICE_OVER_AUDIO_PROTOCOL;
   Files marked with * are only present on BridgeOS.
 **/
 typedef enum {
+  AppleVoiceOverAudioFileIndexLocalisedMin           = 0x01, ///< First valid localised file
   AppleVoiceOverAudioFileVoiceOverOn                 = 0x01, ///< VoiceOverOn
   AppleVoiceOverAudioFileVoiceOverOff                = 0x02, ///< VoiceOverOff
   AppleVoiceOverAudioFileUsername                    = 0x03, ///< Username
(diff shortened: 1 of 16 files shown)


Include/Apple/Protocol/AppleVoiceOver.h
```diff
@@ -48,13 +48,33 @@ typedef enum {
...
   AppleVoiceOverAudioFileVoiceOverBoot2              = 0x3C, ///< VoiceOver_Boot (*)
   AppleVoiceOverAudioFileClick                       = 0x3D, ///< Click (*)
   AppleVoiceOverAudioFileBeep                        = 0x3E, ///< Beep
   AppleVoiceOverAudioFileMax                         = 0x3F,
 } APPLE_VOICE_OVER_AUDIO_FILE;
 
+/**
+  Is localised?
+**/
+#define AppleVoiceOverAudioFileIsLocalised(File)  ((File) < AppleVoiceOverAudioFileIndexNonLocalised)
```
vit9696, line 62: Better to stick to CAPS for all macros still (e.g. EFI_ERROR).

Library/OcAudioLib/OcAudioVoiceOver.c
```diff
@@ -27,6 +27,68 @@
...
+  APPLE_VOICE_OVER_AUDIO_FILE_CLICK,
+  OC_VOICE_OVER_AUDIO_BASE_TYPE_APPLE
+};
+STATIC APPLE_VOICE_OVER_FILE_MAP  mAppleVoiceOverAudioFileBeep = {
+  APPLE_VOICE_OVER_AUDIO_FILE_BEEP,
+  OC_VOICE_OVER_AUDIO_BASE_TYPE_APPLE
+};
+
+CONST APPLE_VOICE_OVER_FILE_MAP  *gAppleVoiceOverAudioFiles[AppleVoiceOverAudioFileMax] = {
+  [AppleVoiceOverAudioFileVoiceOverOn]                 = &mAppleVoiceOverAudioFileVoiceOverOn,
```
vit9696, line 79: Any big reason not to have an array of structures rather than pointers to structures?

Library/OcBootManagementLib/BootAudio.c
```diff
@@ -74,65 +75,61 @@ OcPlayAudioFile (
...
 
   if (Context->OcAudio != NULL) {
-    Status = Context->OcAudio->PlayFile (Context->OcAudio, File, 0, FALSE, TRUE);
+    Status = Context->OcAudio->PlayFile (Context->OcAudio, BasePath, BaseType, TRUE, 0, FALSE, TRUE);
   }
 
   if (Fallback && EFI_ERROR (Status)) {
-    switch (File) {
-      case AppleVoiceOverAudioFileBeep:
+    if (!AsciiStrCmp (BaseType, OC_VOICE_OVER_AUDIO_BASE_TYPE_APPLE)) {
```
vit9696, line 82: I think we stick to == 0.

Library/OcBootManagementLib/BootAudio.c
```diff
@@ -178,50 +175,65 @@ OcPlayAudioEntry (
...
+
+  if ((Entry->EntryIndex > 0) && (Entry->EntryIndex <= 9 + 26)) {
+    Status = OcAsciiSafeSPrint (
+               BasePath,
+               sizeof (BasePath),
+               "%a%c",
+               Entry->EntryIndex > 9 ? "Letter" : "",
+               "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ"[Entry->EntryIndex]
+               );
+    ASSERT_EFI_ERROR (Status);
```
vit9696, line 189: Both assert and if sound weird. Feels like it should be not checked at all or asserted as all the data is predictable and trusted.

Library/OcMainLib/OpenCoreUefiAudio.c
```diff
@@ -325,77 +147,40 @@ OcAudioAcquireFile (
...
-               TmpPath,
-               sizeof (TmpPath),
-               &BaseType,
-               &Localised
-               );
 
-  if (BasePath == NULL) {
-    DEBUG ((DEBUG_INFO, "OC: Unknown Wave %d\n", File));
-    return EFI_NOT_FOUND;
+  if ((BasePath == NULL) || (BaseType == NULL) || (Buffer == NULL) || (*Buffer == NULL)) {
```
vit9696, line 154: I think we do not overuse braces in such situations.

Library/OcAudioLib/OcAudioVoiceOver.c
mikebeaton: Only saving a few bytes. Switch to array of structures?

Include/Apple/Protocol/AppleVoiceOver.h
mikebeaton: Changed in upcoming commit.

Library/OcMainLib/OpenCoreUefiAudio.c
mikebeaton: @vit9696 - uncrustify is currently requiring these, sorry, so left them in!

Library/OcAudioLib/OcAudioVoiceOver.c
mikebeaton: Changed.

Library/OcBootManagementLib/BootAudio.c
mikebeaton: Changed.

Library/OcBootManagementLib/BootAudio.c
mikebeaton: Hope the revised version with static calculation of required string size looks okay? (Should maybe, defensively, still have ASSERT?)

Library/OcAudioLib/OcAudioVoiceOver.c
```diff
@@ -27,6 +27,56 @@
 
 #include "OcAudioInternal.h"
 
+//
+// Convert from Apple file id to file base name and type.
+//
+CONST APPLE_VOICE_OVER_FILE_MAP  gAppleVoiceOverAudioFiles[AppleVoiceOverAudioFileMax] = {
```
vit9696, line 33: To save space we should rather use either two tables for localised and not or subtract 0x30 from localised indices. I like the idea with two tables more.

mikebeaton: @vit9696 - Okay, I think remaining resolved.

PR #347, opened 2022-05-05
BootEntryProtocol: Add system action and hotkey support; implement Reset NVRAM and Toggle SIP as drivers.

mikebeaton:
- Addresses the rest of https://github.com/acidanthera/bugtracker/issues/1777 .
 - Optionally allows resetting NVRAM without destroying BIOS boot entries (as mentioned occasionally in Issues).
   - I've tried using this, and I quite like it - if you select it, you boot automatically back into the OC boot menu, but with the audio volume level back at default and the initial boot item forgotten (along with 'everything' else, i.e. everything safe, too), i.e. 'just like a real mac'; it also preserves BIOS boot entries for OpenShell, Windows, etc., if you need them (which I do, in fact, as I sometimes need to boot into Windows directly, to be allowed to get Dell updates based on my system id).
   - Still, I have added a warning about using this option in docs.
 - Supports native NVRAM reset, which I don't think is widely known, but anyway has been discovered (or rediscovered) recently by Syncretic, [here](https://forums.macrumors.com/search/3416102/?q=resetnvram&t=post&c[thread]=2132317&c[users]=Syncretic&o=relevance).
   - It works, not just on MacPros. :)
 - Supports user specified `csr-active-config` value for Toggle SIP
 - Adds optional `Enabled` and `Disabled` flavours for `Toggle SIP` (allows theme designers to optionally provide distinct icons)


Library/OcBootManagementLib/VariableManagement.c
```diff
@@ -89,8 +90,11 @@ IsDeletableVariable (
     //
     // Only erase boot and driver entries for BDS
     // I.e. BootOrder, Boot####, DriverOrder, Driver####
+    // Preserve Boot#### entries except BootNext when PreserveBoot is TRUE.
     //
-    if (  (  (StrnCmp (Name, L"Boot", StrLen (L"Boot")) == 0)
+    if (  (StrCmp (Name, L"BootNext") == 0)
+       || (  !PreserveBoot
+          && (StrnCmp (Name, L"Boot", StrLen (L"Boot")) == 0)
```
mhaeuser, line 97: Probably for another patchiest, but we should replace all such `StrLen` calls with `L_STR_LEN`

Library/OcBootManagementLib/VariableManagement.c
```diff
@@ -450,21 +446,28 @@ OcDeleteVariables (
   UINTN                         BootOptionSize;
   UINT16                        BootOptionIndex;
 
-  DEBUG ((DEBUG_INFO, "OCB: NVRAM cleanup...\n"));
+  DEBUG ((DEBUG_INFO, "OCB: NVRAM cleanup %u...\n", PreserveBoot));
```
mhaeuser, line 449: Doesn't this need to be %d because BOOLEAN is promoted to int?

Library/OcBootManagementLib/VariableManagement.c
```diff
@@ -450,21 +446,28 @@ OcDeleteVariables (
...
+  } else {
+    BootProtectSize = sizeof (BootProtect);
+    Status          = gRT->GetVariable (
+                             OC_BOOT_PROTECT_VARIABLE_NAME,
+                             &gOcVendorVariableGuid,
+                             NULL,
+                             &BootProtectSize,
+                             &BootProtect
+                             );
+    if (EFI_ERROR (Status)) {
```
mhaeuser, line 468: This also needs to check `BootProtectSize` retained its value, I think.

Platform/ResetNvramEntry/ResetNvramEntry.c
```diff
@@ -0,0 +1,168 @@
...
+  IN           OC_PICKER_CONTEXT  *PickerContext,
+  IN     CONST EFI_HANDLE         Device,
+  OUT       OC_PICKER_ENTRY       **Entries,
+  OUT       UINTN                 *NumEntries
+  )
+{
+  //
+  // Custom entries only.
+  //
+  if (Device != NULL) {
```
mhaeuser, line 54: Hmm, UEFI has the convention to NULL-check all used parameters in protocol calls. However, as it's our own protocol, we could allow this. I never really understood that convention anyway. Cc @vit9696

Platform/ResetNvramEntry/ResetNvramEntry.c
```diff
@@ -0,0 +1,168 @@
...
+// a variable we have access to, but we want to confirm that the system works.
+//
+STATIC
+EFI_STATUS
+InternalSystemActionResetNvram (
+  IN     VOID  *Config
+  )
+{
+  ASSERT (Config == &mPreserveBoot);
+  return OcResetNvram (*((BOOLEAN *)Config));
```
mhaeuser, line 86: Considering the ASSERT, maybe just use `mPreserveBoot`?

Platform/ResetNvramEntry/ResetNvramEntry.c
```diff
@@ -0,0 +1,168 @@
...
+    DEBUG ((DEBUG_WARN, "BEP: ResetNvram %s is ignored due to %s!\n", L"--preserve-boot", L"--native"));
+  }
+
+  Status = gBS->InstallMultipleProtocolInterfaces (
+                  &ImageHandle,
+                  &gOcBootEntryProtocolGuid,
+                  &mResetNvramBootEntryProtocol,
+                  NULL
+                  );
+  ASSERT_EFI_ERROR (Status);
```
mhaeuser, line 165: We _try_ not to ASSERT on such stuff, his may validly fail

Platform/ToggleSipEntry/ToggleSipEntry.c
```diff
@@ -0,0 +1,172 @@
...
+
+  Status = EFI_NOT_FOUND;
+
+  //
+  // Is CHAR_NULL terminated string present?
+  //
+  if (  (LoadedImage->LoadOptions != NULL)
+     && ((LoadedImage->LoadOptionsSize % 2) == 0)
+     && ((LoadedImage->LoadOptionsSize / 2) > 0)
+     && (((CHAR16 *)LoadedImage->LoadOptions)[(LoadedImage->LoadOptionsSize / 2) - 1] == CHAR_NULL)
```
mhaeuser, line 140: Hard for me to check right now, but isn't such logic used elsewhere too? Could try to remember to move it to a library

Platform/ResetNvramEntry/ResetNvramEntry.c
mhaeuser: Well, that was not the point, it was about the arguments that do _not_ get NULL-checked.

Platform/ResetNvramEntry/ResetNvramEntry.c
mikebeaton: Yeees. The ASSERT is just to check that the approach is working, and the deref meant to indicate how something would typically use this (if there were multiple callbacks, presumably the best a method like this could do would be to ASSERT if NULL, I guess).

Docs/Sample.plist
```diff
@@ -1129,8 +1129,6 @@
 			<false/>
```
vit9696, line 1129: Do we still implement CMD+OPT+P+R in OpenCore rather than in the driver?

Platform/ResetNvramEntry/ResetNvramEntry.c
vit9696: I think we should follow suit and return EFI_INVALID_PARAMETER for all non-optional NULL pointer arguments.

mikebeaton: So, if I mark this OPTIONAL, that's all I need to do here, right?

Docs/Sample.plist
mikebeaton: Okay... you're fine with an additional BEP feature to implement I guess arbitrary hotkeys?

Include/Acidanthera/Library/OcBootManagementLib.h
```diff
@@ -530,40 +541,56 @@ typedef struct {
...
-  BOOLEAN        RealPath;
+  BOOLEAN                  RealPath;
+  //
+  // System action. Boot Entry Protocol only. Optional.
+  //
+  OC_BOOT_SYSTEM_ACTION    SystemAction;
+  //
+  // Optional config data for system action. Boot Entry Protocol only.
+  //
+  VOID                     *ActionConfig;
```
mikebeaton, line 585: @vit9696 @mhaeuser - I think I have a minor issue, round here. It is reasonable for a BEP driver to want to either allocate this memory, or not. Currently, I'm doing nothing to free it, if it's allocated. What's the best pattern to follow, here? I think it's reasonably EDK-II-ish to require that this memory is always allocated, on the pool if non-NULL? (In which case, I think perhaps I should use the pre-existing LoadOptions and LoadOptionsSize to store this, and as the parameters for SYSTEM_ACTION?)

Docs/Configuration.tex
```diff
@@ -6824,6 +6802,81 @@ \subsubsection{Additional information}
...
+
+The following configuration options may be specified in the \texttt{Arguments} section for this driver:
+
+\begin{itemize}
+  \item \texttt{-{}-preserve-boot} - Boolean flag, enabled if present. \medskip
+
+  If enabled, BIOS boot entries are not cleared during NVRAM reset. This option should
+  be used with caution, as some boot problems can be fixed by clearing these entries. \medskip
+
+  \item \texttt{-{}-native} - Boolean flag, enabled if present. \medskip
```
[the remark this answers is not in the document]
mikebeaton: Hmm. At first I thought this difference in naming comes from different background assumptions - I got into OpenCore pretty late, running it in on real unsupported Macs, while you built it, obviously got into it very early(!), and running on non-Apple hardware. So I think it's 'native', because I'm thinking about Aapl hardware... But then again, large parts of OpenCore emulate parts of Apple early boot which aren't in non-Apple hardware (&/or replace/modify similar things which are there). So it still seems pretty reasonable to say that this feature triggers the native version of something which OpenCore otherwise 'emulates' (does with it's own version). That said, and if you're still not convinced, ofc I'm happy to change the naming!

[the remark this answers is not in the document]
dakanji: I can't point at specific examples right now, but I seem to get the impression that OC generally uses `apple` for what you describe and that this will fit better with existing instances. Also, `native` could be misunderstood by a casual user as being a "native" OpenCore item/implementation as compared to a third party item/implementation.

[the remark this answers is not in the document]
mikebeaton: Yeah, fair point.

Library/OcBootManagementLib/VariableManagement.c
mikebeaton: Raised separate issue. Can do after this.

mikebeaton:
Hi @vit9696, @mhaeuser - Just updated, including hotkey support, if you could re-review at your convenience.

Platform/ResetNvramEntry/ResetNvramEntry.c
```diff
@@ -0,0 +1,194 @@
...
+  //
+  // Custom entries only.
+  //
+  if (Device != NULL) {
+    return EFI_NOT_FOUND;
+  }
+
+  DEBUG ((DEBUG_INFO, "BEP: Adding Reset NVRAM entry, preserve boot %u, apple %u\n", mPreserveBoot, mUseApple));
+
+  *Entries    = mResetNvramBootEntries;
```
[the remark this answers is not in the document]
mikebeaton: No, should be okay. OC_PICKER_ENTRY entries are created by this driver in GetBootEntries, then passed back to it to free, if needed, in FreeBootEntries immediately after conversion to OC_BOOT_ENTRY.

PR #357, opened 2022-06-11
Support AvoidRuntimeDefrag for macOS 13 DP1

mhaeuser:
macOS 13 Developer Beta 1 introduced a new driver, bootrt.efi, which
now carries the kernel call gate. It communicates with efiboot via an
info buffer passed via LoadOptions. Patch it on load with our hook to
run our kernel entry code.

### DIFF
--- a/Include/Acidanthera/Library/OcDevicePathLib.h
+++ b/Include/Acidanthera/Library/OcDevicePathLib.h
@@ -40,6 +40,18 @@ AppendFileNameDevicePath (
   );
 
 /**
+  Locate the terminating node inside the device path.
+
+  @param[in] DevicePath  The device path used in the search.
+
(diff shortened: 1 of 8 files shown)


Lorys89:
compiled and tried and now makes me install on laptop Dell icl, Great job, with avoidruntime off didn't work

apple-ihack-geek:
what is "AvoidRuntimeDefrag"? I tried searching the guide and googling around online, but can't seem to find any explanation... 
I know defragging is bad from SSDs, but I also have a bunch of HDDs in my system and those should be defragged  right?

Core-i99:
OC has very good documentation
https://github.com/acidanthera/OpenCorePkg/blob/master/Docs/Configuration.pdf

PR #359, opened 2022-06-19
OcAppleKernelLib: Make quirk Patcher parameter optional

PMheart:
Patcher can be NULL when the kext is not found; this should not lead to assertion.

### DIFF
--- a/Library/OcAppleKernelLib/CommonPatches.c
+++ b/Library/OcAppleKernelLib/CommonPatches.c
@@ -51,7 +51,7 @@ CONST UINTN
 STATIC
 EFI_STATUS
 PatchAppleCpuPmCfgLock (
-  IN OUT PATCHER_CONTEXT  *Patcher,
+  IN OUT PATCHER_CONTEXT  *Patcher OPTIONAL,
   IN     UINT32           KernelVersion
   )
 {
@@ -65,7 +65,10 @@ PatchAppleCpuPmCfgLock (
     return EFI_SUCCESS;
   }
 
-  ASSERT (Patcher != NULL);
+  if (Patcher == NULL) {
+    DEBUG ((DEBUG_INFO, "OCAK: Patcher not found under for %a on %u\n", __func__, KernelVersion));
+    return EFI_NOT_FOUND;
+  }
 
   Count     = 0;
   Walker    = (UINT8 *)MachoGetMachHeader (&Patcher->MachContext);
@@ -222,7 +225,7 @@ PATCHER_GENERIC_PATCH
 STATIC
 EFI_STATUS
 PatchAppleXcpmCfgLock (
-  IN OUT PATCHER_CONTEXT  *Patcher,
+  IN OUT PATCHER_CONTEXT  *Patcher OPTIONAL,
   IN     UINT32           KernelVersion
   )
 {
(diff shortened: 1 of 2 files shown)


Library/OcAppleKernelLib/CommonPatches.c
```diff
@@ -242,6 +248,11 @@ PatchAppleXcpmCfgLock (
     return EFI_SUCCESS;
   }
 
+  if (Patcher == NULL) {
```
vit9696, line 251: Makes no sense for kernel patches. Kernel can never be null. Same for certain other patches below.

PMheart: https://github.com/acidanthera/OpenCorePkg/commit/6535f1433e0d62b9b524cbd0120bde443bfa46e4

Library/OcAppleKernelLib/CommonPatches.c
```diff
@@ -242,6 +248,11 @@ PatchAppleXcpmCfgLock (
     return EFI_SUCCESS;
   }
 
+  //
+  // This is a kernel patch, so Patcher cannot be NULL.
+  //
+  ASSERT (Patcher != NULL);
```
vit9696, line 254: Makes little sense to move it here. If it is a kernel patch, then Patcher must not be NULL even on not supported releases.

PMheart: [Fixed.](https://github.com/acidanthera/OpenCorePkg/pull/359/commits/9d76078058c197100bb369e3849387a18d51a6e6)

Library/OcAppleKernelLib/CommonPatches.c
```diff
@@ -60,13 +60,21 @@ PatchAppleCpuPmCfgLock (
   UINT8  *WalkerEnd;
   UINT8  *WalkerTmp;
 
+  if (Patcher == NULL) {
```
vit9696, line 63: Please move that below. Not found looks like an error, but we should rather see skipping. Similar to other places.

PMheart: https://github.com/acidanthera/OpenCorePkg/pull/359/commits/6ebf9c352c2ab60022bbe4717299dae6459de247

Library/OcAppleKernelLib/CommonPatches.c
```diff
@@ -60,12 +60,20 @@ PatchAppleCpuPmCfgLock (
...
+  //
   if (OcMatchDarwinVersion (KernelVersion, KERNEL_VERSION_VENTURA_MIN, 0)) {
     DEBUG ((DEBUG_INFO, "OCAK: Skipping AppleCpuPmCfgLock patch on %u\n", KernelVersion));
     return EFI_SUCCESS;
   }
 
-  ASSERT (Patcher != NULL);
+  if (Patcher == NULL) {
+    DEBUG ((DEBUG_INFO, "OCAK: Skipping %a on NULL Patcher on %u\n", __func__, KernelVersion));
+    return EFI_SUCCESS;
```
vit9696, line 75: Why is this EFI_SUCCESS? If the kext is missing and we did not expect it, I would assume it is an error. Should be EFI_NOT_FOUND. Otherwise looks good. Please merge.

PMheart: https://github.com/acidanthera/OpenCorePkg/pull/359/commits/3e35e5d8c8445c819fd7af4bf8949d444d2ddcdd

PR #360, opened 2022-06-21
ApfsTrimTimeout: replace ambiguous "it"

Artoria2e5:
The passage reads like 0 turns off trim timeout, but it actually turns off trim based on the context. dortania/bugtracker#192


PMheart:
Thanks! I will rebuild pdf myself.

PR #361, opened 2022-06-22
remove pointless stuff from CreateVault sign.command

justinkb:
not sure what the point of all this code this commit removes was.

guess the author didn't realize the pubkey is an output parameter, not an input one

### DIFF
--- a/Utilities/CreateVault/sign.command
+++ b/Utilities/CreateVault/sign.command
@@ -6,11 +6,11 @@ abort() {
 }
 
 cleanup() {
-  echo "Cleaning up keys"
+  echo "Cleaning up key"
   rm -rf "${KeyPath}"
 }
 
-if [ ! -x /usr/bin/dirname ] || [ ! -x /bin/chmod ] || [ ! -x /bin/mkdir ] || [ ! -x /usr/bin/openssl ] || [ ! -x /bin/rm ] || [ ! -x /usr/bin/strings ] || [ ! -x /usr/bin/grep ] || [ ! -x /usr/bin/cut ] || [ ! -x /bin/dd ] || [ ! -x /usr/bin/uuidgen ] ; then
+if [ ! -x /usr/bin/dirname ] || [ ! -x /bin/chmod ] || [ ! -x /bin/mkdir ] || [ ! -x /bin/rm ] || [ ! -x /usr/bin/strings ] || [ ! -x /usr/bin/grep ] || [ ! -x /usr/bin/cut ] || [ ! -x /bin/dd ] || [ ! -x /usr/bin/uuidgen ] ; then
   abort "Unix environment is broken!"
 fi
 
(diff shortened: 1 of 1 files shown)


vit9696:
The changes make no sense to me. I commented inline. Perhaps you can provide more details what you want to do, but the current suggestion makes the script essentially useless.

justinkb:
RsaTool, when used in sign mode, takes 3 further parameters other than the "-sign" parameter. They are, in order, 1) the location of the file you want to sign (input), 2) the location of the signature file (output) and 3) the location of the pubkey (crucially, this is an output parameter, refer to https://github.com/acidanthera/OpenCorePkg/blob/02b365e22a542f333e4775a68eb40090d740f054/Utilities/RsaTool/RsaTool.c#L187 and https://github.com/acidanthera/OpenCorePkg/blob/02b365e22a542f333e4775a68eb40090d740f054/Utilities/RsaTool/RsaTool.c#L197)

In sign mode, the tool generates a key itself and writes it to the chosen location. I have no idea who designed it like this, but that's what it seems to do. It is not really a useful general purpose method, but makes sense for this vaulting purpose, I suppose since the key is irrelevant except as a one-time usage for signing the vault.plist. Now, afterwards, the create_vault script simply copies it into the opencore.efi binary as normal before it being signed for EFI image verification.

I feel like I'm missing something since you closed this so quickly, tho. I can't see any inline comments btw

justinkb:
I stand by my commit. You have to look past what it looks like the original code (the stuff I'm deleting) is doing. That entire original code block never made any sense in all the years it was in the repository. It's generating a CA certificate, a private key and a public key, but none of them have any effect on the vault.sig or opencore.efi products, since RsaTool doesn't use the private key at all (it generates its own keypair) and it simply overwrites the just-generated public key with its own generated public key.

justinkb:
@vit9696 are you planning on taking another look at this with my comments in mind? I have a few other commits for CreateVault lined up (making the code portable between macOS and Linux)

justinkb:
I should add the question if RsaTool is actually used for anything else in this project. If not, the whole tool can be scrapped from the repo, since openssl can do everything (cryptographically speaking, that is) the script needs to do by itself (which would make it even easier for linux users to use the script)

vit9696:
Yes, you are actually right. RsaTool does not take the private key, so it cannot do signing with a known key even in theory. Thank you for finding this.

As for getting rid of openssl, the format we use was not supported by an old OpenSSL used in macOS. Is it supported now? If yes, which version?

There is also an important part of using RsaTool here: private key is not written to disk.

justinkb:
That's a good point about the private key. Probably better to keep RsaTool
then. I will submit the PR for the Linux port stuff in a bit, when I've
cleaned it up just a bit. I've made it so it still produces the exact same
plist files as the original version, without relying on plistbuddy, on both
macos and linux. The PR will also have a commit dealing with memtest things
(so the create vault script ignores memtest log files and benchmark
reports, we don't wanna hash those)

1alessandro1:
> making the code portable between macOS and Linux

@justinkb when you can, please take a look at https://github.com/acidanthera/bugtracker/issues/1999

PR #362, opened 2022-06-27
CreateVault linux portability

justinkb:
Some notes:

- the plist output functions produce the exact same plist outputs as PlistBuddy as far as I've been able to tell.
- I'm piping the find output through sort with LC_COLLATE=POSIX to have it behave the same on linux as it does on macos. (doesn't matter functionally since OpenCore doesn't care about the order, but it's just nice to have the exact same vault.plists produced on all platforms)
- have to use awk instead of cut in sign.command since binutils (on linux) "strings -t d" outputs numbers that are padded with spaces on the left, whereas that isn't the case on macos apparently.
- the memtest86 things aren't really related, but seem like no brainers to include anyway
- it is still the user's responsibility to compile RsaTool (from Utilities/RsaTool) for their platform and place it in the directory along these scripts, maybe there should be a README in the CreateVault directory with this information?

### DIFF
--- a/Utilities/CreateVault/create_vault.sh
+++ b/Utilities/CreateVault/create_vault.sh
@@ -74,7 +74,7 @@ write_header vault.plist
   \( ! -iname "vault.*" \) \
   \( ! -iname "MemTest86.log" \) \
   \( ! -iname "MemTest86-Report-*.html" \) \
-  \( ! -iname "OpenCore.efi" \) | env LC_COLLATE=POSIX sort | while read -r fname; do
+  \( ! -iname "OpenCore.efi" \) | env LC_COLLATE=POSIX /usr/bin/sort | while read -r fname; do
   fname="${fname#"./"}"
   wname="${fname//\//\\\\}"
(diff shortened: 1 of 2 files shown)


Utilities/CreateVault/create_vault.sh
```diff
@@ -17,54 +17,80 @@ if [ ! -d "${OCPath}" ]; then
...
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>Files</key>
+	<dict>
+EOF
+}
+
+write_file_name_and_hash() {
+  {
```
PMheart, line 44: Is this pair of braces really needed?

justinkb: I actually didn't know that was possible, the { cmd1; cmd2 } >> thing. Shellcheck suggested it all, pretty neat!

vit9696:
> it is still the user's responsibility to compile RsaTool (from Utilities/RsaTool) for their platform and place it in the directory along these scripts, maybe there should be a README in the CreateVault directory with this information?

Sounds like a good idea.

PR #363, opened 2022-06-28
OCVAR: Refactor NVRAM code, add OpenVariableRuntimeDxe

mikebeaton:
Apologies this PR has taken quite so long to produce!

 - ~There are a few minor code tidy-ups in other files as well (all more or less directly related to the PR); so not quite as many files changed strictly for the PR as it seems - I hope this is okay and reasonable to review~ [EDIT: Now committed separately.]
 - All the files in `Platform/OpenVariableRuntimeDxe` are exact clones of the files with the same name in `UDK/MdeModulePkg/Universal/Variable/RuntimeDxe`, except for changes in the following two only:
   - `VariableDxe.c`
   - `VariableRuntimeDxe.inf`
 - The refactoring of the NVRAM code into one library is I hope a good idea anyway, but is mainly to clean up the dependencies involved in linking some of that code with Duet now
 - Some other minor comments/questions in TODOs in the code

Assuming the code looks vaguely plausible, it would be good to find out how it works for Duet users (i.e. Reset NVRAM and CTRL+Enter support, and the new `Launchd.command` script, which should hopefully allow correct automatic installation of at least Monterey and Big Sur OTA updates).


Application/OpenCore/OpenCore.c
```diff
@@ -135,6 +136,7 @@ OcMain (
 
   DEBUG ((DEBUG_INFO, "OC: OcLoadNvramSupport...\n"));
   OcLoadNvramSupport (Storage, &mOpenCoreConfiguration);
+  OcReportVersion (&mOpenCoreConfiguration);
```
vit9696, line 139: Why was it moved here?

Docs/Configuration.tex
```diff
@@ -567,6 +571,12 @@ \subsection{Directory Structure}\label{directory-structure}
 \item
   \texttt{nvram.plist} \\
   OpenCore variable import file.
+\item
+  \texttt{nvram.fallback} \\
+  OpenCore variable import fallback file.
+\item
+  \texttt{nvram.used} \\
+  Renamed  previous OpenCore variable import file after switch to fallback file.
```
vit9696, line 579: Extra space.

Docs/Sample.plist
```diff
@@ -1271,36 +1271,21 @@
...
-				<string>SystemAudioVolumeSaved</string>
-				<string>bluetoothActiveControllerInfo</string>
-				<string>bluetoothInternalControllerInfo</string>
-				<string>flagstate</string>
-				<string>fmm-computer-name</string>
-				<string>fmm-mobileme-token-FMM</string>
-				<string>fmm-mobileme-token-FMM-BridgeHasAccount</string>
-				<string>nvda_drv</string>
-				<string>prev-lang:kbd</string>
+				<string>*</string>
```
vit9696, line 1284: This looks rather dangerous to me. Better to keep what we had before.

Docs/SampleCustom.plist
```diff
@@ -1298,36 +1298,21 @@
 				<string>ForceDisplayRotationInEFI</string>
 			</array>
 		</dict>
-		<key>LegacyEnable</key>
-		<false/>
 		<key>LegacyOverwrite</key>
 		<false/>
 		<key>LegacySchema</key>
 		<dict>
+			<key>36C28AB5-6566-4C50-9EBD-CBB920F83843</key>
```
vit9696, line 1305: Same here.

Include/Acidanthera/Library/OcConfigurationLib.h
```diff
@@ -474,11 +474,13 @@ OC_DECLARE (OC_NVRAM_LEGACY_ENTRY)
   OC_MAP (OC_STRING, OC_NVRAM_LEGACY_ENTRY, _, __)
 OC_DECLARE (OC_NVRAM_LEGACY_MAP)
 
+//
+// Modifications to OC_NVRAM_CONFIG require updating OC_VARIABLE_RUNTIME_PROTOCOL_REVISION.
```
vit9696, line 478: Do we pass this structure to the driver now?

Library/OcBootManagementLib/BootEntryInfo.c
```diff
@@ -698,13 +698,15 @@ InternalDescribeBootEntry (
     BootEntry->Name = AllocateCopyPool (sizeof (L"Windows"), L"Windows");
   }
 
+  // TODO: Should macOS installer have own OC_BOOT_ENTRY_TYPE (plus own voiceover file?)?
```
vit9696, line 701: Think yes…

Library/OcBootManagementLib/BootEntryManagement.c
```diff
@@ -2380,6 +2381,10 @@ OcLoadBootEntry (
              &DmgLoadContext
              );
   if (!EFI_ERROR (Status)) {
+    if (BootEntry->IsAppleInstaller) {
```
vit9696, line 2384: Could we provide a more descriptive comment here with reasoning?

Library/OcMainLib/OpenCorePlatform.c
```diff
@@ -951,6 +951,7 @@ OcGetLegacySecureBootECID (
   }
 
   CopyMem (ApECID, &Uuid, sizeof (*ApECID));
+  DEBUG ((DEBUG_INFO, "OC: ApECID %ld from system-id\n", *ApECID));
```
vit9696, line 954: if I remember correctly, it should be %Ld for a 64-bit number. I am also not positive whether it should be her though.

Library/OcMainLib/OpenCoreUefi.c
```diff
@@ -117,19 +120,30 @@ OcLoadDrivers (
...
 
-    DEBUG ((
-      DEBUG_INFO,
-      "OC: Driver %a at %u (%a) is %a\n",
-      DriverFileName,
-      Index,
-      DriverComment,
-      SkipDriver ? "skipped!" : "being loaded..."
-      ));
+    if (!LoadEarly || !SkipDriver) {
```
vit9696, line 137: Why that particular if is used?

Library/OcVariableLib/LegacyVariable.c
```diff
@@ -0,0 +1,182 @@
...
+  //
+  // It is not really required to support boot var routing with emulated NVRAM (since there are
+  // no firmware NVRAM boot vars used outside of OpenCore to avoid trashing), but having working
+  // support is more convenient when switching back and forth between emulated and non-emulated
+  // NVRAM, i.e. one less thing to have to remember to switch, since with this code everything
+  // works as expected with or without RequestBootVarRouting. (Without it, boot entries do not
+  // restore to the right place when RequestBootVarRouting is enabled.)
+  // OpenRuntime.efi must be loaded early, but after OpenVariableRuntimeDxe.efi, for this to work.
+  //
+  if (Config->Uefi.Quirks.RequestBootVarRouting) {
```
vit9696, line 74: Ugh, that's a bit too tight. Can we avoid accessing config data outside of the main lib?

Library/OcVariableLib/VariableManagement.c
```diff
@@ -167,6 +208,7 @@ DeleteVariables (
 
   //
   // Assume we have not failed yet.
+  // TODO: Noting that critical failure can happen exactly once, across all variables, (why) is it needed?
```
vit9696, line 211: Critical failure represents a second NVRAM failure. Why did you add this comment?

Library/OcVariableLib/VariableManagement.c
```diff
@@ -178,7 +220,7 @@ DeleteVariables (
       TmpBuffer = AllocateZeroPool (RequestedSize);
       if (TmpBuffer != NULL) {
         if (Buffer != NULL) {
-          CopyMem (TmpBuffer, Buffer, BufferSize);
+          CopyMem (TmpBuffer, Buffer, BufferSize); ///< TODO: Is this copy required?
```
vit9696, line 223: Why not?

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
```diff
@@ -0,0 +1,621 @@
...
+               );
+
+    if (EFI_ERROR (Status)) {
+      continue;
+    }
+
+    VariableMap = NvramStorage.Add.Values[GuidIndex];
+
+    //
+    // TODO: Is OC LegacyOverwrite setting still needed, given that this code
```
vit9696, line 185: Hrm, why would it be empty?

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
```diff
@@ -0,0 +1,621 @@
...
+    }
+  } while (Status == EFI_BUFFER_TOO_SMALL);
+
+  if (EFI_ERROR (Status)) {
+    SaveContext->Status = Status;
+    return OcProcessVariableAbort;
+  }
+
+  //
+  // Only save non-volatile variables; also, match launchd script and only save
```
vit9696, line 293: Ugh, cannot we use OcXmlLib there?

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
```diff
@@ -0,0 +1,621 @@
...
+  FreePool (FileBuffer);
+
+  return Status;
+}
+
+STATIC
+OC_VARIABLE_RUNTIME_PROTOCOL
+  mOcVariableRuntimeProtocol = {
+  OC_VARIABLE_RUNTIME_PROTOCOL_REVISION,
+  (OC_VARIABLE_RUNTIME_PROTOCOL_LOAD_NVRAM)LoadNvram,
```
vit9696, line 602: There should not be needed a cast. Why?

OpenCorePkg.dsc
```diff
@@ -272,6 +275,22 @@
   OpenCorePkg/Platform/OpenPartitionDxe/PartitionDxe.inf
   OpenCorePkg/Platform/OpenRuntime/OpenRuntime.inf
   OpenCorePkg/Platform/OpenUsbKbDxe/UsbKbDxe.inf
+  OpenCorePkg/Platform/OpenVariableRuntimeDxe/VariableRuntimeDxe.inf {
+    <PcdsFixedAtBuild>
+      gEfiMdeModulePkgTokenSpaceGuid.PcdEmuVariableNvModeEnable|TRUE
+      gEfiMdeModulePkgTokenSpaceGuid.PcdMaxVariableSize|0x10000
+      gEfiMdeModulePkgTokenSpaceGuid.PcdVariableStoreSize|0x10000
```
vit9696, line 282: Is not 64K way too little for emulated NVRAM? Thought we increased that for OpenDuet.

vit9696 (commented):
Have not checked the driver and the script yet.

OpenCorePkg.dsc
dakanji: Microsoft, effectively the gatekeeper, says the min variable store size for SecureBoot should be 120kb (See Page 7 here: https://uefi.org/sites/default/files/resources/UEFI_Plugfest_Security_Microsoft_Fall_2016.pdf). The MacOS requirements presumably then need to be added to this.

For reference, running QueryVariableInfo on my basic Dell inspiron laptop from 2015 shows that it has a 384kb store (393216 bytes). It would be nice to run this on a post-2012 Mac to get a feel for what Apple is doing with those as they apparently went significantly higher than 64kb as from the Trashcan.

Docs/Configuration.tex
```diff
@@ -6954,6 +6940,119 @@ \subsubsection{Configuration}
   some other systems. \medskip
 \end{itemize}
 
+\subsection{OpenVariableRuntimeDxe}\label{emunvram}
+
+Provides in-memory emulated NVRAM implementation. This can be useful on systems with fragile
+(e.g. MacPro5,1, see discussion linked from this \href{https://forums.macrumors.com/threads/opencore-on-the-mac-pro.2207814/page-425?post=30945127#post-30945127}{forum post})
```
dakanji, line 6946: You could tighten the link here as follows:
https://forums.macrumors.com/posts/30945127

mikebeaton: They're definitely not always tightened elsewhere. I didn't know that format, I'll make the change. Thanks.

OpenCorePkg.dsc
mikebeaton: This is the value in OpenDuet currently, I just copied it. I'm happy to change it to whatever, in both, ofc.

FWIW, maybe of note that we keep far from the whole contents of NVRAM in the file we save and restore, so I suspect this is in practice enough. (Of course, we have to allow for enough space for whatever the OS puts in there while running, even though much/some of it may be discarded on each reboot - i.e. only what is a) saved by `nvramdump` and then b) kept by `LegacySchema` is preserved across reboots.)

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
mikebeaton: Hi. A cast is not needed. But I noticed that a few libraries in EDK-II use this convention. The point (at least as I've understood it, and why I've added it) it is to allow a search for `OC_VARIABLE_RUNTIME_PROTOCOL_LOAD_NVRAM` to find `LoadNvram` (and same for the others). Otherwise both `LoadNvram`, and its use in constructing this protocol, only _implicitly_ reference the type (i.e. no global search for the type will find it, even though its signature has to match the type, in fact).

Library/OcVariableLib/VariableManagement.c
mikebeaton: Why? You re-read the entire variable again into the new buffer no? (The copy makes it look as if it will just read the rest of the variable, beyond what was read before into the too-small buffer; but the call to read the variable isn't even passed enough info (i.e. the old size as well as the new size) for this to be possible, I think.)

Docs/SampleCustom.plist
mikebeaton: Sorry - I meant to revert these. Will do. Although see my comments in the docs. In fact it is *far* easier in actual use to leave these as `*`, for two reasons:

 1. You can set test vars of arbitrary names see whether things are working as you expect
   - (Also 1a you can examine the contents of the saved file and see everything that was present to save, not just what you allowed it to save, though this is more for debugging the driver itself)
 2. Exactly what macOS saves as installer vars changes on each version (AFAICT), and so it is possible that an installer will fail just because we only include here what we thought was required, based on a previous installer.

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
mhaeuser: That „convention“ is dangerous because it may hide type mismatches, including calling convention. It’s usually best to do the opposite of EDK II. :)

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
mikebeaton: When we've initialised emulated NVRAM, before we anything into it from our file, it must be empty, right?  But I guess not - things in the emulated NVRAM driver itself (in the case of OC-only use) and in the rest of OpenDuet (in the case of OpenDuet) could already have populated certain things, right?

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
mikebeaton: You mean as opposed to me currently manually building up the XML 'by hand' in a string buffer?

If so, then the answer is that currently OcXmlLib has no write support. (Even if it was considered preferable to add that, that is another fairly major project...! :-// )

Library/OcBootManagementLib/BootEntryInfo.c
mikebeaton: I think yes, too, but it's another bunch of fiddly changes... not entirely sure if it would happen for Monday or similar timescale.

Library/OcMainLib/OpenCorePlatform.c
mikebeaton: I'm not sure there is any other way to tell what's finally used? (When I wanted to know this, for sure, at some point, I didn't think I could otherwise tell?)

Library/OcVariableLib/VariableManagement.c
mikebeaton: I guess I'm just asking whether it wouldn't make more sense to allow two retries per variable, or else to not bother allowing two retries at all? I guess I'm just not quite sure of the reasoning (or the known example where it helps, quite likely) where allowing it to retry after failing once, on one variable, is desirable, but then if it fails again, possibly on another variable, it is then desirable to stop the scan. But happy to leave as is!

Docs/Sample.plist
mikebeaton: See below.

Library/OcBootManagementLib/BootEntryManagement.c
mikebeaton: Will add. (Obviously the reasoning per se is scattered around the docs and comments in the PR (!) - but I will add a specific brief comment here.)

Library/OcMainLib/OpenCoreUefi.c
mikebeaton: Good question ... I was thinking of removing the skipped lines entirely - as there can be lots, and they perhaps don't add much (that isn't at least implicitly present anyway!) - but reasoned they can be useful all the same. But with an early and late phase of loading, you get two lots of lots of skipped lines.

I preferred to only show what is loaded in the early phase, not what is skipped, and to show what is skipped in the later/normal phase (which would include anything loaded early, at that point).

Note that if I do include these here, these come in early log (while logging is still being buffered, and when OC is writing log info to whatever default console device is present) - which has another issue for me:

 - I need as few lines as possible in this stage of early log, or my MBP10,2 (still!) hangs after the firmware update to 429.0.0.0.0 (as I already reported to you), so unfortunately for me, having a load of skipped lines here also renders by MBP10,2 more or less unbootable with OC.

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
mikebeaton: (Noting that in the case of OC previously, there was another use entirely, which was when reading from `nvram.plist`, but writing over non-emulated, normal NVRAM. But this use no longer exists.)

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
mikebeaton: Yes, I already spotted that that was the countervailing reason _not_ to use it. It is a shame to have no way to get both advantages, but I don't think there is one. (Unless using `///<` comment???) I'll remove it.

mhaeuser: Unsure, I haven‘t ever felt a need for that. I‘d search for the protocol over the function type. But I don‘t mind a comment (but do mind the cast).

Application/OpenCore/OpenCore.c
vit9696: I personally strongly dislike the idea of having config code anywhere outside of OcMainLib. Can we avoid this?

Library/OcBootManagementLib/BootEntryInfo.c
vit9696: Can postpone, yes.

Library/OcMainLib/OpenCorePlatform.c
vit9696: Fair. Just change the print format then.

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
mikebeaton: I'll just remove. Obviously it's findable - as you say, by searching for the protocol - it's just an extra step or two. (Steps which I find myself doing, quite often!!) (But unless it's done consistently everywhere, having it sort of findable another way in some places is probably misleading, anyway.)

Docs/Configuration.tex
dakanji: It must just be that I was at some point thinking all such external links _SHOULD_ be explictly written and not hyperlinked.

The same format can be used for the Applelife.ru link further below btw as they use the same software as MacRumors.
Both the long and short format do not currently work for AppleLife ... server is down or SSL cert is invalid/revoked.

mikebeaton: I just checked again, and it's almost exclusively done in the way I've done it, in the main `Configuration.pdf` file (which fits with the original intended usage of hyperlinks, btw). So if you believe it should be done another way, I think that's a separate battle for you to fight in a separate issue! ;-)

But thanks for the short link, equally for AppleLife, and yes I noticed that was down (at least for us to access? not sure).

Application/OpenCore/OpenCore.c
mikebeaton: So just to be sure I understand, you do mean you're against the idea of moving what I've moved (NVRAM config code), into a separate NVRAM lib?

In that case (pending being back at Mac/hack and actually trying it) I think we could move _some_ of this back into OcMainLib.

But the essential issue is that the new emulated NVRAM protocol now has to do some of what the normal OC NVRAM config does (some of the same code, for the same reasons). So _that_ code cannot be in OcMainLib, without having to then link OcMainLib into OpenDuet - which creates a horrible mess (with TONS of otherwise non-needed stuff having to be linked into OpenDuet (only for most of it to be discarded)), as far as I can see.

mikebeaton: @vit9696 - Okay I have tried to address this as a new commit on the PR (i.e. trying to make it more or less convenient so that the new commit itself will show the changes just made, while the whole PR will still show the changes against the current OC code - I guess that's obvious!).

The main change is:
 - I have ensured that no OC global config is accessed outside of OcMainLib now ... except LegacySchema/OC_NVRAM_LEGACY_MAP
 - I still can't see how to avoid that. (Although I guess we could define that structure not in the config, but then include it in the config - but I think that wouldn't be a great improvement, if any.)

Also other minor fixes of other issues mentioned below, for which I'll comment or close the relevant conversations.

Docs/SampleCustom.plist
mikebeaton: I've reverted these, and left some comments about this in the documentation.

Library/OcVariableLib/VariableManagement.c
mikebeaton: Doh! Sorry.

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
mikebeaton: I'm assuming that my new comment added here (basically covering my second comment above) captures why this is still useful.

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
mikebeaton: Removed.

Library/OcBootManagementLib/BootEntryManagement.c
mikebeaton: Added.

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
mikebeaton: Can we postpone this one for now, too?

Include/Acidanthera/Library/OcConfigurationLib.h
mikebeaton: With the current commit, I now only pass OC_NVRAM_LEGACY_MAP (which as far as I can see is required), not all of OC_NVRAM_CONFIG.

mikebeaton:
@vit9696 @mhaeuser As mentioned, I have cleaned up this PR to contain only what is directly relevant; other very minor changes, plus the ASCII/Unicode constant change in the string libs (which were worth having, but were adding some noise to this already fairly large PR) are now committed separately, below this, live on master.

Library/OcVariableLib/VariableManagement.c
mikebeaton: I'm still not quite seeing what I'm missing, in this case. Clearly, stopping on NVRAM failure - either immediately or eventually - is sensible. I only didn't fully understand, why stop after two failures, possibly on two different variables? (E.g. Why not n failures, where n != 2? Or n per variable? Or just 1?) But I'm definitely not saying I see anything 'wrong' or dangerous with the code as is! (So by 'why is this needed' I just meant, 'why not just n=1'?)

mhaeuser: I’m not sure either, honestly. If it was at least forcing distinct variables, you could argue „twice per one variable could mean something about the variable is off, twice for different variables must mean the storage is off“, but that doesn’t seem to be what’s happening. Maybe this just is „good enough“?

Library/OcVariableRuntimeLib/OcVariableRuntimeLib.c
vit9696: Ok.

OpenCorePkg.dsc
vit9696: Let's keep it as is and then increase as needed.

vit9696:
In my opinion, time to merge this. Thank you!

PR #366, opened 2022-07-07
OcLog: Allow optional unsafe fast log

mikebeaton:
@vit9696 @mhaeuser Is something like this acceptable? I can add docs and changelog if so.

@mhaeuser I tried incrementally overwriting a pre-created file, but of the two slightly different approaches I've tried for that, it's either no faster than rewriting the whole thing, or else somewhat faster, but still quite variable as @joevt [reported](https://github.com/acidanthera/bugtracker/issues/2074), and much slower than the basic, simple, fast approach on at least some runs.

Btw there is a clear and pleasant improvement in DEBUG boot speed (as expected really) even on a reasonably fast machine when using UNSAFE.


Library/OcLogAggregatorLib/OcLog.c
```diff
@@ -363,24 +366,45 @@ InternalLogAddEntry (
     // Write to internal buffer.
     //
 
+    STATIC_ASSERT (OC_LOG_BUFFER_SIZE <= MAX_UINT32, "log buffer size must fit within 32 bits");
+    OldOffset = (UINT32)AsciiStrLen (Private->AsciiBuffer);
```
mhaeuser, line 370: Any clue how long this takes accumulated? Would it be beneficial to record this to `Private` instead of re-determining it?

Library/OcLogAggregatorLib/OcLog.c
```diff
@@ -363,24 +366,45 @@ InternalLogAddEntry (
...
+
     Status = AsciiStrCatS (Private->AsciiBuffer, Private->AsciiBufferSize, Private->TimingTxt);
     if (!EFI_ERROR (Status)) {
       Status = AsciiStrCatS (Private->AsciiBuffer, Private->AsciiBufferSize, Private->LineBuffer);
     }
 
+    NewOffset = (UINT32)AsciiStrLen (Private->AsciiBuffer);
```
mhaeuser, line 377: This can be moved down into the nested ifs, no?

Include/Acidanthera/Protocol/OcLog.h
```diff
@@ -137,6 +138,7 @@ struct OC_LOG_PROTOCOL_ {
   UINTN                  DisplayLevel;  ///< The error level visible onscreen.
   UINTN                  HaltLevel;     ///< The error level causing CPU dead loop.
   EFI_FILE_PROTOCOL      *FileSystem;   ///< Log file system root, not owned.
+  EFI_FILE_PROTOCOL      *LogFile;      ///< Log file, owned. Unsafe logging only.
```
mhaeuser, line 141: Please move it all the way down, this is unsafe. In the Revision-model, instances of lesser revision must be prefixes of those of a higher revision. I.e., you can only append (or use previously reserved data), to preserve backwards-compatibility. This is of course not critical here, but I also see no good reason to not do it.

Library/OcLogAggregatorLib/OcLog.c
mikebeaton: Definitely < 10% of the time we're saving by making this change overall. But yes, can put to private and see.

mikebeaton: This made no measurable difference at all (change from 3-4ms per log line to 3-4ms per log line... as opposed to 20-40ms with safe write), but I've moved it to private in upcoming revision anyway.

mikebeaton:
@mhaeuser - I've made the suggested changes. Could you check that they look sane, including in the case where the settings of a pre-existing OC_LOG_PROTOCOL instance can get updated in `OcConfigureLogProtocol`? (I am a little uneasy about whether this makes sense yet, with the current changes.)

Library/OcLogAggregatorLib/OcLog.c
```diff
@@ -363,24 +366,51 @@ InternalLogAddEntry (
...
+          WrittenSize = WriteSize;
+          OcLog->UnsafeLogFile->Write (OcLog->UnsafeLogFile, &WrittenSize, &Private->AsciiBuffer[OldAsciiBufferOffset]);
+          OcLog->UnsafeLogFile->Flush (OcLog->UnsafeLogFile);
+          if (WriteSize != WrittenSize) {
+            DEBUG ((
+              DEBUG_VERBOSE,
+              "OCL: Log write truncated %u to %u\n",
+              WriteSize,
+              WrittenSize
+              ));
```
mhaeuser, line 400: Can this recurse?

mikebeaton: Hmm, I guess not, there are multiple verbose debug statements in OcSetFileData as used in the safe path.

Docs/Configuration.tex
```diff
@@ -3888,6 +3889,12 @@ \subsection{Debug Properties}\label{miscdebugprops}
   avoid frequent use of this option when dealing with flash drives as large I/O
   amounts may speed up memory wear and render the flash drive unusable quicker.
 
+  \textbf{Warning}: It is possible to enable fast file logging, which works as indicated
```
PMheart, line 3892: Better change it to `Warning 2` to match our other `Note`s.

mikebeaton: Will do

PR #368, opened 2022-07-11
OcAppleKernelLib: Fix user after free

PMheart:
closes https://github.com/acidanthera/bugtracker/issues/2082

This PR fixes the cases of use after free in OcAppleKernelLib.

CC and thanks @mhaeuser


vit9696:
Looks good to me.

PR #371, opened 2022-07-14
RsaTool: Build and link against static LibreSSL on macOS

dhinakg:
Instead of trying to link against Apple's libcrypto which is intended for internal use and is now in the dyld shared cache, build LibreSSL as a static library and link against it instead.

This script will only run if the following three conditions are met:

1. `HAS_OPENSSL_BUILD` is set (to enable RsaTool compilation)
2. The host is macOS (the script does not support other operating systems; it is assumed that OpenSSL is available on these systems)
3. `libressl/lib/libcrypto.a` is missing (will only recompile if it is missing)

Binary has been tested to run on 10.11 and 12.0. Additional testing would be appreciated. [Sample binary for testing.](https://github.com/acidanthera/OpenCorePkg/files/9112788/RsaTool.zip)

As a side note, this also allows Apple Silicon to build OpenCore outside of Rosetta now.


Utilities/RsaTool/build_libressl.sh
```diff
@@ -0,0 +1,137 @@
...
+  # If we are building on arm64 (Apple Silicon), these extra options are required to ensure x86_64 builds.
+  EXTRA_OPTS=(--host=arm-apple-darwin --build=x86_64-apple-darwin)
+  CFLAGS+=" --target=x86_64-apple-darwin"
+  LDFLAGS+=" --target=x86_64-apple-darwin"
+else
+  EXTRA_OPTS=()
+fi
+
+# Monkeypatch to disable strtonum for <11.0 support
+"${SED}" -i '' -E 's/strsep strtonum/strsep/g' configure || ret=$?
```
PMheart, line 98: If I remember correctly, `-i ''` is only required by macOS.

Hm, since this is only for macOS, could be fine

PMheart:
You might also just `doSomething || abort` instead of checking `ret`?

Utilities/RsaTool/build_libressl.sh
dhinakg: BSD sed needs `-i ''` for in-place, and the only supported OS is macOS so no compatibility problems to deal with

dhinakg:
> You might also just `doSomething || abort` instead of checking `ret`?

Good point, thanks, changed in bc308d9

vit9696:
Need to please shellcheck and uncrustify, otherwise looks about right.

dhinakg:
ShellCheck fixed in 83bdd87, Uncrustify fixed in master

PR #373, opened 2022-07-19
OCAB: Implement macOS 10.4/10.5 relocation block support

Goldfish64:
Implements macOS 10.4 and 10.5 support for relocation blocks while using the AllowRelocationBlock booter quirk.

This is required due to XNU in 10.4 and 10.5 having a _TEXT base address of 0x111000, different from 0x200000 in 10.6 and later. SetVirtualAddressMap is also always invoked by EfiBoot even when using a relocation block.

### DIFF
--- a/Library/OcAfterBootCompatLib/RelocationBlock.c
+++ b/Library/OcAfterBootCompatLib/RelocationBlock.c
@@ -311,7 +311,7 @@ AppleRelocationRebase (
   // On macOS 10.6 and newer, we need to adjust it like the others.
   //
   if (!BootCompat->KernelState.RelocationBlockLegacy) {
-    *BA->SystemTableP  -= RelocDiff;
+    *BA->SystemTableP -= RelocDiff;
   } else {
     *BA->MemoryMapSize -= (*BA->MemoryMapDescriptorSize * 4);
   }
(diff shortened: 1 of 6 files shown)


Library/OcAfterBootCompatLib/RelocationBlock.c
```diff
@@ -290,9 +332,43 @@ AppleRelocationRebase (
...
+    DescLargestAddress = Desc->PhysicalStart;
+    NumEntries         = MemoryMapSize / DescriptorSize;
+
+    //
+    // Locate end of valid memory map. It is assumed that the entries are in order from
+    // lowest to highest address, performed by RebuildAppleMemoryMap booter quirk.
+    //
+    for (Index = 0; Index < NumEntries; ++Index) {
+      if (Desc->PhysicalStart >= DescLargestAddress) {
+        DescLargestAddress = Desc->PhysicalStart;
```
mhaeuser, line 358: The `DescLargestAddress` name and the control flow format make it look like it looks for a maximum, but that's not the case. I would invert the if-condition (move else branch up), move this outside the branch (it will `break`), and rename the variable to `PrevDescAddress`. I'm not sure this is safe though, some firmwares do this OOTB. What is the exact issue?

mhaeuser (changes requested):

vit9696:
LGTM.

PR #374, opened 2022-07-24
OpenCanopy: Avoid invalid animation state when returning to menu after starting entry with animated label

mikebeaton:
@mhaeuser - sent pm explaining reasoning behind this and possible alternative/additional options for related changes.


mikebeaton:
Okay, redoing as per comments - tyvm

PR #380, opened 2022-08-06
ocpasswordgen: Hide password characters on input

savvamitrofanov:
This changes should hides characters on password input. In this case I use getch from conio.h library on Windows and also implemented getch for unix using unistd and termios standard libraries.

### DIFF
--- a/Utilities/ocpasswordgen/ocpasswordgen.c
+++ b/Utilities/ocpasswordgen/ocpasswordgen.c
@@ -4,10 +4,47 @@
 **/
 
 #include <stdio.h>
+#ifdef _WIN32
+#include <conio.h>
+#else
+#include <termios.h>
+#include <unistd.h>
+#endif
 
 #include <Library/DebugLib.h>
 #include <Library/OcCryptoLib.h>
 #include <UserPseudoRandom.h>
+
+//
+// Signal Interrupt Control-C symbol
+//
+#define CHAR_END_OF_TEXT 3
+#define CHAR_DELETE      127
+
+#ifndef _WIN32
+/**
+ * getch unix-compatible implementation
+ *
+ * @return integer with ch codenumber
+ */
+int
+getch (
+  void
+  )
+{
+    int Char;
+    struct termios OldTtyAttrs, NewTtyAttrs;
+
+    tcgetattr(STDIN_FILENO, &OldTtyAttrs);
+    NewTtyAttrs = OldTtyAttrs;
+    NewTtyAttrs.c_lflag = NewTtyAttrs.c_lflag & ~(ICANON|ECHO);
+    tcsetattr(STDIN_FILENO, TCSANOW, &NewTtyAttrs);
+    Char = getchar();
+    tcsetattr(STDIN_FILENO, TCSANOW, &OldTtyAttrs);
+
+    return Char;
+}
+#endif
 
 int
 ENTRY_POINT (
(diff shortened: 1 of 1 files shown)


Utilities/ocpasswordgen/ocpasswordgen.c
```diff
@@ -21,13 +58,34 @@ ENTRY_POINT (
...
+    if (Char == EOF || Char == CHAR_LINEFEED) {
       break;
     }
+    if (Char == CHAR_CARRIAGE_RETURN) {
+      continue;
+    }
+    if (Char == CHAR_BACKSPACE || Char == CHAR_DELETE) {
+      if (PasswordLen > 0) {
+        DEBUG ((DEBUG_ERROR, "\b \b"));
+        PasswordLen--;
```
PMheart, line 79: `--PasswordLen;`

Utilities/ocpasswordgen/ocpasswordgen.c
```diff
@@ -21,13 +58,34 @@ ENTRY_POINT (
...
+    fflush(stdin);
+    Char = getch ();
+    // Handle interrupt signal for conio
+    if (Char == CHAR_END_OF_TEXT) {
+      exit(EXIT_SUCCESS);
+    }
+    if (Char == EOF || Char == CHAR_LINEFEED) {
       break;
     }
+    if (Char == CHAR_CARRIAGE_RETURN) {
```
PMheart, line 73: else if? and same below.

Utilities/ocpasswordgen/ocpasswordgen.c
```diff
@@ -21,13 +58,34 @@ ENTRY_POINT (
   UINT8   Index;
   UINT8   PasswordHash[SHA512_DIGEST_SIZE];
 
-  DEBUG ((DEBUG_ERROR, "Please enter your password: "));
+  DEBUG ((DEBUG_ERROR, "Please enter your password: \n"));
```
vit9696, line 61: In my opinion, better to switch output to printf here and leave it with printf below. DEBUG goes through a lot of intermediate buffers, and may even be disabled. Also, that extra space before `\n` is not needed.

Utilities/ocpasswordgen/ocpasswordgen.c
vit9696: No, since break does not return control.

Utilities/ocpasswordgen/ocpasswordgen.c
```diff
@@ -21,13 +58,34 @@ ENTRY_POINT (
...
+    if (Char == CHAR_BACKSPACE || Char == CHAR_DELETE) {
+      if (PasswordLen > 0) {
+        DEBUG ((DEBUG_ERROR, "\b \b"));
+        PasswordLen--;
+      }
+      continue;
+    }
+
+    if (Char == ' ') {
+      DEBUG ((DEBUG_ERROR, " "));
```
vit9696, line 85: I do not get it why spaces are handled differently. Password can contain spaces, and it should just display as a normal `*` in this case.

Utilities/ocpasswordgen/ocpasswordgen.c
```diff
@@ -21,13 +58,34 @@ ENTRY_POINT (
...
+  DEBUG ((DEBUG_ERROR, "Please enter your password: \n"));
 
   for (PasswordLen = 0; PasswordLen < OC_PASSWORD_MAX_LEN; ++PasswordLen) {
-    Char = getchar ();
-    if ((Char == EOF) || (Char == '\n')) {
+    fflush(stdin);
+    Char = getch ();
+    // Handle interrupt signal for conio
+    if (Char == CHAR_END_OF_TEXT) {
+      exit(EXIT_SUCCESS);
```
vit9696, line 68: Why is that success? The password was not generated, in my opinion, that's `EXIT_FAILURE`.

Utilities/ocpasswordgen/ocpasswordgen.c
```diff
@@ -4,11 +4,48 @@
 **/
 
 #include <stdio.h>
+#ifdef _WIN32
+#include <conio.h>
```
vit9696, line 8: Need to test this on Windows before merging.

Utilities/ocpasswordgen/ocpasswordgen.c
PMheart: Shall we use printf for all userspace tools as well?

vit9696: I believe not, only here it is critical.

Utilities/ocpasswordgen/ocpasswordgen.c
savvamitrofanov: Yep, it should be EXIT_FAILURE

Utilities/ocpasswordgen/ocpasswordgen.c
savvamitrofanov: I tested it before creating pull request, it works

Utilities/ocpasswordgen/ocpasswordgen.c
savvamitrofanov: Definitely yes! Would be better don't print password chars at all to hide password length as we discussed before.

Utilities/ocpasswordgen/ocpasswordgen.c
savvamitrofanov: Thx. I'll apply uncrustify

PR #382, opened 2022-08-10
Rename FullNvramAccess to RawNvramAccess

dakanji:
See discussion here: https://github.com/acidanthera/OpenCorePkg/commit/b3f2033d200bb8bf114baad861bc7a1a51ed8259#comments

Proceeded with PR as it appeared the reluctance expressed was more of a reluctance to implement the change as opposed to a view that the proposed rename was of no value or, in fact, detrimental.

Excuse me if that interpretation is wrong! (EDIT: Now better understand reticence seeing how finicky Uncrustify is!) 

PS: Docs say contributors should not build the PDFs ... so they will need building by a team member if merged.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,7 @@ OpenCore Changelog
 ==================
 #### v0.8.4
 - Added checks for `Driver` -> `LoadEarly` in ocvalidate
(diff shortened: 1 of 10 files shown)


dakanji:
Noted

PR #384, opened 2022-08-18
Build: Add network stack drivers into build

savvamitrofanov:
This commit adds basic set of NetworkPkg drivers with HTTP boot support.
Also adds missing DxePcdLib into HiiDatabaseDxe.


Docs/Sample.plist
```diff
@@ -1611,6 +1611,150 @@
 				<key>Path</key>
 				<string>Ext4Dxe.efi</string>
 			</dict>
+			<dict>
```
PMheart, line 1614: In my opinion, only put these drivers in SampleCustom.plist.

build_oc.tool
```diff
@@ -158,11 +158,20 @@ package() {
     cp "${arch}/Shell.efi" "${dstdir}/${arch}/EFI/OC/Tools/OpenShell.efi" || exit 1
 
     efiDrivers=(
+      "ArpDxe.efi"
```
PMheart, line 161: Instead of alphabetical order, shall we classify by authors (e.g. ours vs external)?

OpenCorePkg.dsc
```diff
@@ -167,7 +184,10 @@
   MdeModulePkg/Bus/Isa/Ps2MouseDxe/Ps2MouseDxe.inf
   MdeModulePkg/Bus/Isa/Ps2KeyboardDxe/Ps2KeyboardDxe.inf
   MdeModulePkg/Bus/Usb/UsbMouseDxe/UsbMouseDxe.inf
-  MdeModulePkg/Universal/HiiDatabaseDxe/HiiDatabaseDxe.inf
+  MdeModulePkg/Universal/HiiDatabaseDxe/HiiDatabaseDxe.inf {
+    <LibraryClasses>
+      PcdLib|MdePkg/Library/DxePcdLib/DxePcdLib.inf
```
vit9696, line 189: DxePcdLib requires pcd protocols, which legacy firmware may not have.

Docs/Sample.plist
savvamitrofanov: Yes, I agree, this is not a common set of drivers, I'll remove them from Sample.plist

build_oc.tool
savvamitrofanov: I think this will complicate maintaining this in the future, so I think we should leave it as it is

PR #385, opened 2022-09-06
OcAppleKernelLib: Support for PageableKC/SysKC injection

flagersgit:
In Ventura, Apple has decided to remove kext binaries from the root volume and require KDK installation in order to generate a new PageableKC/SysKC via `kmutil`. 

The Auxiliary KC remains able to be created without the need of KDK installation, however this is not viable for use cases where kexts in the SysKC must be removed or supplanted.

This is particularly problematic for projects such as OpenCore Legacy Patcher where automation is key and the KDK download is locked behind Apple ID authentication.

Such a feature as this, aside from solving the above problem, opens the door to "vanilla" injection of kexts which have linkage requirements for other kexts which are in the SysKC (e.g. IOGraphicsFamily for a hypothetical fully-featured framebuffer driver from MacHyperVSupport).

While this feature is meant to run in primarily non-UEFI environments<sup>1</sup>, it ideally should also be functional in a UEFI<sup>2</sup> (given a SysKC separately), even if not exposed through the configuration.

## Design Goals
* Functional from UEFI, User-space, and XNU/kernel-space itself (compiled as an external module).
* If being used from kernel-space, it ideally must be able to use the booted kernel cache<sup>3</sup>
---
1. (continued)...non-UEFI environments, such as user space or kernel space itself by hooking OSKext's and other related routines.
2. It would not be practical to use this in UEFI as the kexts which reside, or need to reside, in the SysKC are typically much larger (e.g. containing firmware blobs, etc.) and would not fit on the ESP. 
There is no mechanism in UEFI to link an APFS Preboot volume folder (where the BootKC resides) to its paired System volume (where the SysKC resides).
Further, exposing a modified KC in its entirety to XNU's OSKext would be cumbersome and likely waste memory.
3. This is so that one may "chain" injections. For example, downgrading IOSurface in BootKC from OpenCore by blocking with Exclude strategy and re-injecting, then from kernel-space ([with whichever out-of-scope component may do this](https://github.com/moraea/Veneer/)) downgrading IOAcceleratorFamily by the same procedure.


.gitignore
```diff
@@ -87,3 +87,4 @@ Utilities/TestProcessKernel/OC/*
 Utilities/TestProcessKernel/ProcessKernel
 Utilities/TestNtfsDxe/TestNtfsDxe
 Utilities/TestExt4Dxe/TestExt4Dxe
+Utilities/TestPageableInject/PageableInject
```
PMheart, line 90: Missing newline; same below.

Utilities/TestPageableInject/Makefile
```diff
@@ -0,0 +1,39 @@
...
+	infback.o \
+	inffast.o \
+	inflate.o \
+	inftrees.o \
+	trees.o \
+	uncompr.o \
+	zlib_uefi.o \
+	zutil.o
+VPATH   = ../../Library/OcAppleKernelLib:$\
+	../../Library/OcCompressionLib/lzss:$\
```
PMheart, line 36: Indentation.

PR #389, opened 2022-10-25
OcCpuLib: Cache InternalCalculateARTFrequencyIntel failure to avoid log loop

mikebeaton:
I think this is correct and with no unwanted side effects.

Note previously, by definition, the call to InternalCalculateARTFrequencyIntel inside OcGetTSCFrequency was capable of returning a zero result (otherwise it would never drop through to the other options), however it never cached a zero result and always recalculated it (and therefore always logged while doing so).

The updated code follows exactly the pattern already used in InternalCalculateTSCFromApplePlatformInfo.

### DIFF
--- a/Library/OcCpuLib/FrequencyDetect.c
+++ b/Library/OcCpuLib/FrequencyDetect.c
@@ -447,8 +447,9 @@ InternalCalculateARTFrequencyIntel (
   // this frequency on module entry to initialise a TimerLib instance, and at
   // a later point in time to gather CPU information.
   //
-  STATIC UINT64  ARTFrequency        = 0;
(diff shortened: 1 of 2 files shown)


Library/OcCpuLib/FrequencyDetect.c
```diff
@@ -463,11 +464,14 @@ InternalCalculateARTFrequencyIntel (
   UINT8                          Model;
 
   if (Recalculate) {
+    ObtainedFreqs       = FALSE;
```
vit9696, line 467: Perhaps rename to ObtainedARTFreq? Otherwise looks good to me.

PR #390, opened 2022-10-25
OpenCoreNvram: Add UnprotectAppleNvram patches

mikebeaton:
Hi @PMheart - I forced pushed and I think trashed your comments by mistake - apologies. I've already addressed 1 of the 2 (?) of them anyway, and I'll retype the other.

When it's a PR comments are kept even through force pushes etc., so I'm making it a PR now.

### DIFF
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -8728,6 +8728,18 @@ for additional options.
   \emph{Note}: This quirk is useful in cases where unsuccessful drive detection
   results in an absence of boot entries.
 
+\item
+  \texttt{UnprotectAppleNvram}\\
+  \textbf{Type}: \texttt{plist\ boolean}\\
+  \textbf{Failsafe}: \texttt{false}\\
+  \textbf{Description}: Unprotect blocked NVRAM variables on recent Apple firmware.
+
+  Newer Apple firmware wraps NVRAM access in various security checks. This quirk applies patches
+  which disable these checks, allowing OpenCore to set variables such as
+  \texttt{FA4CE28D-B62F-4C99-9CC3-6815686E30F9:gpu-power-prefs}, which are otherwise blocked.
+
+  \emph{Note}: This quirk lowers system security and should not be used unless specifically needed.
+
 \end{enumerate}
 
 \subsection{ReservedMemory Properties}\label{uefirsvdprops}
--- a/Library/OcMainLib/OpenCoreNvram.c
+++ b/Library/OcMainLib/OpenCoreNvram.c
@@ -203,12 +203,117 @@ OcAddNvram (
   }
 }
 
+#if defined (MDE_CPU_X64)
+// test rax, rax
+// jz ...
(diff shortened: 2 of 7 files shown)


Docs/Configuration.tex
```diff
@@ -8728,6 +8728,18 @@ \subsection{Quirks Properties}\label{uefiquirkprops}
...
+  \texttt{UnprotectAppleNvram}\\
+  \textbf{Type}: \texttt{plist\ boolean}\\
+  \textbf{Failsafe}: \texttt{false}\\
+  \textbf{Description}: Unprotect blocked NVRAM variables on recent Apple firmware.
+
+  Newer Apple firmware wraps NVRAM access in various security checks. This quirk applies patches
+  which disable these checks, allowing OpenCore to set variables such as
+  \texttt{FA4CE28D-B62F-4C99-9CC3-6815686E30F9:gpu-power-prefs}, which are otherwise blocked.
+
+  \emph{Note}: This quirk lowers system security and should not be used unless specifically needed.
```
mikebeaton, line 8741: @pmheart - changed - ty.

Library/OcMainLib/OpenCoreNvram.c
```diff
@@ -203,12 +203,117 @@ OcAddNvram (
...
+STATIC CONST UINT8  CheckSecurityListGetNext[] = {
+  0x48, 0x85, 0xDB, 0x74
+};
+
+// xor rbx, rbx
+// jz ...
+STATIC CONST UINT8  DontCheckSecurityListGetNext[] = {
+  0x48, 0x31, 0xDB, 0x74
+};
+#endif
```
mikebeaton, line 230: @PMheart commented that these might be too short.

My reply was that I am replacing within short functions (see search lengths below) and I know that I am  searching exactly from the start of the function. Therefore I thought these were likely to be specific enough, and more likely to work on variants of the functions. I could add more bytes, but thought it might link them to too specific a version of the functions, since it has to work on firmware for several different Macs.

As I said (in a lost comment, due to force push not within PR - it's now a PR!), I am not certain of the best approach, and I wondered if @vit9696 (or @khronokernel ??) or anyone else had any time to confirm if this looks okay or not? Many thanks in advance if so.

mikebeaton: Btw, the basic structure of the functions is:
 - get a pointer to a linked list
 - if not NULL, loop round the linked list calling installed security check methods, aborting if check fails
 - do the thing we want
So I'm just forcing the early check to treat the list of security functions as empty.

mikebeaton: (Probably should add, I think I'll find out more if this current version fails or crashes on some macs, rather than just failing to match as would become more likely if I add more.)

mikebeaton:
Due to severe problems with this strategy on some machines (allows you to set var values which brick the machine) I am abandoning this approach for now.

PR #394, opened 2022-11-11
[macrecovery] Create `com.apple.recovery.boot`

dreamwhite:
I know this sounds stupid but since the only possible way to use the Internet Recovery Install method is `macrecovery.py`, I thought it would be great that the script creates the folder `com.apple.recovery.boot` where the downloaded image and chunklist goes.

Proof of work:

image

image

image

### DIFF
--- a/Utilities/macrecovery/macrecovery.py
+++ b/Utilities/macrecovery/macrecovery.py
@@ -461,7 +461,7 @@ def main():
                         ' "selfcheck" checks whether MLB serial validation is possible, "verify" performs'
                         ' MLB serial verification, "guess" tries to find suitable mac model for MLB.')
     parser.add_argument('-o', '--outdir', type=str, default='com.apple.recovery.boot',
-                        help='customise output directory for downloading, defaults to current directory')
+                        help='customise output directory for downloading, defaults to com.apple.recovery.boot')
     parser.add_argument('-n', '--basename', type=str, default='',
                         help='customise base name for downloading, defaults to remote name')
     parser.add_argument('-b', '--board-id', type=str, default=RECENT_MAC,
--- a/Utilities/macrecovery/macrecovery.py
+++ b/Utilities/macrecovery/macrecovery.py
@@ -214,7 +214,7 @@ def save_image(url, sess, filename='', directory=''):
     if filename.find('/') >= 0 or filename == '':
         raise RuntimeError('Invalid save path ' + filename)
 
-    print(f'Saving {url} to {filename}...')
+    print(f'Saving {url} to {DOWNLOAD_DIR}/{filename}...')
 
     with open(os.path.join(directory, filename), 'wb') as fh:
         response = run_query(url, headers, raw=True)
--- a/Utilities/macrecovery/macrecovery.py
+++ b/Utilities/macrecovery/macrecovery.py
@@ -24,7 +24,7 @@ except ImportError:
     from urlparse import urlparse
 
 SELF_DIR = os.path.dirname(os.path.realpath(__file__))
-COM_APPLE_RECOVERY_BOOT = 'com.apple.recovery.boot'
+DOWNLOAD_DIR = 'com.apple.recovery.boot'
 
 RECENT_MAC = 'Mac-7BA5B2D9E42DDD94'
 MLB_ZERO = '00000000000000000'
@@ -205,9 +205,9 @@ def save_image(url, sess, filename='', directory=''):
         'Cookie': '='.join(['AssetToken', sess])
     }
 
-    if not os.path.exists(COM_APPLE_RECOVERY_BOOT):
-        os.mkdir(COM_APPLE_RECOVERY_BOOT)
-    directory = COM_APPLE_RECOVERY_BOOT
+    if not os.path.exists(DOWNLOAD_DIR):
+        os.mkdir(DOWNLOAD_DIR)
+    directory = DOWNLOAD_DIR
 
     if filename == '':
         filename = os.path.basename(purl.path)
--- a/Utilities/macrecovery/macrecovery.py
+++ b/Utilities/macrecovery/macrecovery.py
@@ -204,6 +204,10 @@ def save_image(url, sess, filename='', directory=''):
         'Cookie': '='.join(['AssetToken', sess])
     }
 
+    if not os.path.exists('com.apple.recovery.boot'):
+        os.mkdir('com.apple.recovery.boot')
+    directory = 'com.apple.recovery.boot'
+
     if filename == '':
         filename = os.path.basename(purl.path)
     if filename.find('/') >= 0 or filename == '':
--- a/Utilities/macrecovery/macrecovery.py
+++ b/Utilities/macrecovery/macrecovery.py
@@ -24,7 +24,6 @@ except ImportError:
     from urlparse import urlparse
 
 SELF_DIR = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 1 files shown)


Utilities/macrecovery/macrecovery.py
```diff
@@ -204,6 +204,10 @@ def save_image(url, sess, filename='', directory=''):
         'Cookie': '='.join(['AssetToken', sess])
     }
 
+    if not os.path.exists('com.apple.recovery.boot'):
+        os.mkdir('com.apple.recovery.boot')
+    directory = 'com.apple.recovery.boot'
```
PMheart, line 209: Declare this variable first and then reuse it?

dreamwhite: Ohh, you mean declaring once `com.apple.recovery.boot` as a `str` variable and then reuse it?

PMheart: Yes.

dreamwhite: Can you please check latest commit?

Utilities/macrecovery/macrecovery.py
```diff
@@ -24,6 +24,7 @@
     from urlparse import urlparse
 
 SELF_DIR = os.path.dirname(os.path.realpath(__file__))
+COM_APPLE_RECOVERY_BOOT = 'com.apple.recovery.boot'
```
dhinakg, line 27: `DOWNLOAD_DIR` may be a more appropriate variable name.

dhinakg (commented):
Maybe adjust https://github.com/dreamwhite/OpenCorePkg/blob/edb2b71ccfe6e673ab4b5d1b34e59acea8050894/Utilities/macrecovery/macrecovery.py#L217 to include the directory name

Utilities/macrecovery/macrecovery.py
dreamwhite: Alright, done

dreamwhite:
> Maybe adjust https://github.com/dreamwhite/OpenCorePkg/blob/edb2b71ccfe6e673ab4b5d1b34e59acea8050894/Utilities/macrecovery/macrecovery.py#L217 to include the directory name

Alright, done

Utilities/macrecovery/macrecovery.py
```diff
@@ -204,12 +205,16 @@ def save_image(url, sess, filename='', directory=''):
         'Cookie': '='.join(['AssetToken', sess])
     }
 
+    if not os.path.exists(DOWNLOAD_DIR):
```
vit9696, line 208: This change is inadequate. `directory` is passed as an argument, and you always override it. Perhaps you should change the function prototype to `directory=DOWNLOAD_DIR` instead of `directory=''`, and then create `directory` if it does not exist here.

vit9696 (changes requested):

Utilities/macrecovery/macrecovery.py
dreamwhite: Alright. I made another change in the latest commit, since `save_image` function gets called with a `directory` override.

I dropped such override, since by default `directory` value is `com.apple.recovery.boot` in the function prototype, despite I don't think that's the best way. Suggestions are welcome

Utilities/macrecovery/macrecovery.py
```diff
@@ -280,9 +283,9 @@ def action_download(args):
         print(info)
     print(f'Downloading {info[INFO_PRODUCT]}...')
     dmgname = '' if args.basename == '' else args.basename + '.dmg'
-    dmgpath = save_image(info[INFO_IMAGE_LINK], info[INFO_IMAGE_SESS], dmgname, args.outdir)
```
vit9696, line 283: And now you broke output directory handling in macrecovery arguments. Please edit:

```py
    parser.add_argument('-o', '--outdir', type=str, default=os.getcwd(),
                        help='customise output directory for downloading, defaults to current directory')
```

to

```
    parser.add_argument('-o', '--outdir', type=str, default='com.apple.recovery.boot',
                        help='customise output directory for downloading, defaults to com.apple.recovery.boot')
```

dreamwhite: Hi, sorry for the inconvenience. Can you please check latest commit?

Utilities/macrecovery/macrecovery.py
```diff
@@ -457,7 +460,7 @@ def main():
                         help='Action to perform: "download" - performs recovery downloading,'
                         ' "selfcheck" checks whether MLB serial validation is possible, "verify" performs'
                         ' MLB serial verification, "guess" tries to find suitable mac model for MLB.')
-    parser.add_argument('-o', '--outdir', type=str, default=os.getcwd(),
+    parser.add_argument('-o', '--outdir', type=str, default='com.apple.recovery.boot',
                         help='customise output directory for downloading, defaults to current directory')
```
vit9696: It no longer defaults to current directory, thus this should be renamed.

Utilities/macrecovery/macrecovery.py
```diff
@@ -195,7 +195,7 @@ def get_image_info(session, bid, mlb=MLB_ZERO, diag=False, os_type='default', ci
     return info
 
 
-def save_image(url, sess, filename='', directory=''):
+def save_image(url, sess, filename='', directory='com.apple.recovery.boot'):
```
vit9696, line 198: I guess this change can be reverted. Otherwise good.

Utilities/macrecovery/macrecovery.py
dreamwhite: Yeah sorry, commit dreamwhite/OpenCorePkg@24d83a3 has the fixed help message

Utilities/macrecovery/macrecovery.py
dreamwhite: Currently testing without the override of the function prototype. `com.apple.recovery.boot` directory gets created and `BaseSystem.dmg` is downloading. A few mins and I'll confirm if `BaseSystem.chunklist` as well as the chunklist verification process completes, and I'll push the revert commit for the function prototype

dreamwhite:
Given that users must know before-hand what they're downloading, is it possible to guess the current downloaded macOS version? It would be useful as `macrecovery.py` script can write onto `com.apple.recovery.boot/.contentDetails` the name of the recovery (e.g. `Recovery 12.6.1`)

vit9696:
Merged, thank you!

As for version detection, this sounds nice. You can try to prototype this, I do not mind merging such a PR.

dreamwhite:
> As for version detection, this sounds nice. You can try to prototype this, I do not mind merging such a PR.

I think the most doable way is the following:

- Use `hdiutil mount -plist com.apple.recovery.boot/BaseSystem.dmg`
- Parse the output of the previous command using `plistlib` and extract the mount point (usually it's `/Volumes/macOS Base\ System` but I wanna make it as dynamic as possible (e.g. multiple `macOS Base System` mounted, therefore the path won't be static but it'll contain `-1/2/3 etc` or whatever)
- Use `/usr/libexec/PlistBuddy -c "Print "DTPlatformVersion {mountpoint}/Install\ macOS*.app/Contents/Info.plist` to extract the version of the recovery environment (I don't know why but even downloading the latest Monterey image gets `12.5` version and installs `12.6.1`. Probably an Apple bug, but need further research)
- Write a file named `.contentDetails` under `com.apple.recovery.boot` (can use both Python or even Bash by appending using `>> com.apple.recovery.boot/.contentDetails` 

Let me know your thoughts on that and if there's any hole in this

PR #395, opened 2022-11-16
Hide Picker when waking from macOS Hibernation

khronokernel:
Allows for seamless wake from hibernation, and avoids situation where someone may accidentally change the boot option when waking. Tests were conducted on a genuine MacBookPro11,3

Notes:

* Changes ShowPicker logic to be a string, takes following inputs:
  * `Always` - Show Picker regardless (previously `ShowPicker=True`).
  * `Never` - Never Show Picker (previously `ShowPicker=False`).
  * `SkipOnHibernateWake` - Hide Picker if machine is waking from macOS Hibernation.
    * Picker will still be shown if invoked via PollAppleHotKeys' Option/Esc keys.
    * Explicit setting to avoid accidental boot loops from botched Hibernation Configurations (Primarily PCs) where user may not realize or doesn't have PollAppleHotKeys enabled.
* Currently only detects macOS Hibernation, Linux and Windows Hibernation detection is unsupported at this time.

Reference: 
* https://github.com/acidanthera/bugtracker/issues/1722

### DIFF
--- a/Library/OcMainLib/OpenCoreMisc.c
+++ b/Library/OcMainLib/OpenCoreMisc.c
@@ -935,8 +935,8 @@ OcMiscBoot (
      && (AsciiStrCmp (ShowPicker, "SkipOnHibernateWake") != 0)
      && (AsciiStrCmp (ShowPicker, "Never") != 0))
   {
-    DEBUG ((DEBUG_WARN, "OC: Unknown ShowPicker: %a, using Default\n", ShowPicker));
-    ShowPicker = "Default";
+    DEBUG ((DEBUG_WARN, "OC: Unknown ShowPicker: %a, using Always\n", ShowPicker));
+    ShowPicker = "Always";
   }
 
   if (!EFI_ERROR (Status)) {
--- a/Library/OcMainLib/OpenCoreMisc.c
+++ b/Library/OcMainLib/OpenCoreMisc.c
@@ -931,7 +931,7 @@ OcMiscBoot (
              );
 
   ShowPicker = OC_BLOB_GET (&Config->Misc.Boot.ShowPicker);
-  if (  (AsciiStrCmp (ShowPicker, "Default") != 0)
+  if (  (AsciiStrCmp (ShowPicker, "ShowAlways") != 0)
      && (AsciiStrCmp (ShowPicker, "SkipOnHibernateWake") != 0)
      && (AsciiStrCmp (ShowPicker, "Hide") != 0))
   {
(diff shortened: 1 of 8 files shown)


Library/OcMainLib/OpenCoreMisc.c
```diff
@@ -930,6 +930,8 @@ OcMiscBoot (
              );
   if (!EFI_ERROR (Status)) {
     Context->PickerCommand = OcPickerBootAppleRecovery;
+  } else if (OcIsAppleHibernateWake () && Config->Misc.Boot.HidePickerHibernate) {
+    Context->PickerCommand = FALSE;
```
vit9696, line 934: Hmmm, but what if hibernate boot entry was not found?

mhaeuser: @vit9696 This is at the early stage where there's just information from the config parsed and no actual handling has taken place yet. Thus, the idea of this looks just fine. If the hibernate boot entry is not found, it should invoke the same handling we already have in place for when `ShowPicker` is false.

@khronokernel However, this is still broken:
1. ``FALSE`` is not a valid value for ``PickerCommand``.
2. As `OcIsAppleHibernateWake()` performs "actual work", it should be on the right-hand side.
3. This is special handling for when `ShowPicker` is enabled. Thus, either the conditions could be appended to the `ShowPicker` handling as such:
```
else if (Config->Misc.Boot.ShowPicker && (!Config->Misc.Boot.HidePickerHibernate || !OcIsAppleHibernateWake ())) {
  Context->PickerCommand = OcPickerShowPicker;
}
```
... or it could be handled more explicitly inside:
```
else if (Config->Misc.Boot.ShowPicker) {
  if (!Config->Misc.Boot.HidePickerHibernate || !OcIsAppleHibernateWake ()) {
    Context->PickerCommand = OcPickerShowPicker;
  } else {
    Context->PickerCommand = OcPickerDefault;
  }
}
```
5. There is no concept for the interactive side of things. Say a macOS installation is hibernated and the user manually enters the picker with no memory of hibernating, then this option isn't of much help. I think Apple completely blocks entering the picker, though that seems tricky regarding potential boot loops. This could be solved in many ways, such as a special hotkey to force-enter the picker, or at least having a visual indicator of hibernation, such as in the boot entry title and icon. @vit9696 Preferences?

mhaeuser: @vit9696 @khronokernel Actually, if this remains to be a special mutator for `ShowPicker`, maybe it should be something like `ShowPicker = Never | Normal | Force` (the naming is bad, but you get the idea)?

khronokernel: Issues 1 through 4 should be resolved now, waiting on comments for 5

vit9696: 5 is indeed a problem, but I do not see a good way to resolve it. A visual indicator would be tough to add, and a special hotkey may be another problem for boards with unreliable keyboard drivers. In my opinion, merge and address it only if there is too much trouble.

PR #397, opened 2022-11-21
Add ChassisType configuration for PlatformInfo->Generic

khronokernel:
Reference:

* https://github.com/acidanthera/bugtracker/issues/2162

### DIFF
--- a/Docs/Sample.plist
+++ b/Docs/Sample.plist
@@ -1320,6 +1320,8 @@
 		<dict>
 			<key>AdviseFeatures</key>
 			<false/>
+			<key>ChassisType</key>
+			<integer>0</integer>
 			<key>MLB</key>
 			<string>M0000000000000001</string>
 			<key>MaxBIOSVersion</key>
--- a/Docs/SampleCustom.plist
+++ b/Docs/SampleCustom.plist
@@ -1378,6 +1378,8 @@
 		<dict>
 			<key>AdviseFeatures</key>
 			<false/>
+			<key>ChassisType</key>
+			<integer>0</integer>
 			<key>MLB</key>
 			<string></string>
 			<key>MaxBIOSVersion</key>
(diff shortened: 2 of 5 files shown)


PMheart:
Documentation in Configuration.tex/pdf is missing, otherwise looks good to me.

vit9696:
This feels way too much for `Generic`, to be honest. I would say we should use `Custom` configuration in such situations.

hectorskypl:
> @khronokernel, I agree with @vit9696. Close it.

@Andrey1970AppleLife, @vit9696, please not close stuff which is still broken and which is not resolved.
I think @khronokernel did that for debug purpose for a while, it really need further investigation rather just say "use custom SMBIOS" and close issue, I used it but what? It still ain't work...

I request if is possible to reopen my issue, cause it still ain't fixed.

hectorskypl:
> "use custom SMBIOS" and close issue, I used it but what? It still ain't work...
> 
> > 
> 
> What value you used? We don't know value for Rack. And I already responded you [here](https://github.com/acidanthera/bugtracker/issues/2162#issuecomment-1322100553) about dmidump.

I used that what said to try @khronokernel so 23, I thought it will be fine, then I decided also dig further going to stops which khronokernel did, until I get to dead end where I was not able to find anything more.

But even when we know value for Tower enclosure why in system profiler it's ain't displayed as Tower or why in IOReg it's also not present like in original IOReg dump which Khronokernel used to show me as example?
To be exact its in: IOReg:/efi/platform there should be also injected chassis-type, but OpenCore not does it, that why I try research it.

I just try understand where this thing might be overall injected... that thing called "enclosure"

Andrey1970AppleLife:
>   why in IOReg it's also not present like in original IOReg dump which Khronokernel used to show me as example? To be exact its in: IOReg:/efi/platform there should be also injected chassis-type, but OpenCore not does it, that why I try research it.
> 

CC @vit9696 , OC bug?

vit9696:
Let's better move to the issue and discuss the problem there. This PR can stay for now till we figure out what needs to be done.

PR #398, opened 2022-11-21
OcBootManagementLib: Enable Apple picker on non-EFI graphics cards

mikebeaton:


Library/OcBootManagementLib/OcBootManagementLib.c
```diff
@@ -241,21 +256,29 @@ OcRunBootPicker (
...
       //
-      if (Context->HideAuxiliary || (Context->PickerCommand == OcPickerProtocolHotKey)) {
-        DEBUG ((DEBUG_INFO, "OCB: System has no boot entries, showing picker with auxiliary\n"));
+      if (Context->HideAuxiliary || (Context->PickerCommand == OcPickerProtocolHotKey) || IsApplePickerSelection) {
         Context->PickerCommand = OcPickerShowPicker;
         Context->HideAuxiliary = FALSE;
+        if (IsApplePickerSelection) {
+          DEBUG ((DEBUG_WARN, "OCB: Apple Picker returned no entry valid under OC, falling back to builtin\n"));
+          Context->PickerMode = OcPickerModeBuiltin;
+          gST->ConOut->ClearScreen (gST->ConOut);
```
mikebeaton, line 269: Apologies, but after putting up a hopefully quite useful PR, I am going to ask a really stupid question! This ClearScreen does not seem to do anything - i.e. the builtin picker text comes up over the still present final Apple picker screen - even though the ClearScreen not far below it definitely does do something to a graphics mode screen (hence the recent change to not call it when in graphics mode). What am I missing, if anyone can save me some time by pointing out the no-doubt obvious?

Library/OcBootManagementLib/OcBootManagementLib.c
```diff
@@ -455,3 +479,50 @@ OcRunFirmwareApplication (
...
+      APPLE_FIRMWARE_USER_INTERFACE_PROTOCOL_REVISION
+      ));
+    Status = EFI_UNSUPPORTED;
+  }
+
+  if (!EFI_ERROR (Status)) {
+    //
+    // Location of relevant byte variable within loaded driver.
+    //
+    aGopAlreadyConnected = (VOID *)((UINT8 *)FirmwareUI + sizeof (APPLE_FIRMWARE_USER_INTERFACE_PROTOCOL));
```
mikebeaton, line 514: Modifying this byte flag is definitely valid for the specific 144.0.0.0.0 version of the firmware I was testing, but I guess it could be more robust (e.g. the same version of the protocol functionally doesn't have to have the same memory layout) to write a small patch to the `ConnectGop` method which this affects.

mikebeaton:
I'd be okay to merge this now.

1. BootKicker restarts after Apple picker has finished (which will have set BootNext, always, and BootOrder, if CTRL+Enter). This is cleaner than returning, as BootKicker will often be used to break out of OC boot var redirection. If we do not restart immediately, we are likely to leave pending and perhaps unexpected changes remaining in the non redirected Boot vars.

2. I have used a GOP clear screen to resolve the issue that returning to the builtin picker results in it displaying over the remains of the Apple picker (which is possibly unexpected given that the builtin picker internally already returns to text mode and clears screen, but definitely true on the OC settings which I have been using to test this, with most things overridden by OC).

Item 1 is pointing out a change, but I believe it is correct. Item 2 I believe is required.

In actual use, this is pretty seamless now. We do still have some long delays on entering and exiting the Apple picker, in some cases. These seem to be highly dependent on OC text/graphics rendering settings (e.g. with some settings the Apple picker can be lightning fast), but I have not yet fully investigated this.

PR #403, opened 2022-11-29
Include EnableForAll

CrisHotpatch:
ACPI->Quirks->EnableForAll
Booter->Quirks->EnableForAll

### DIFF
--- a/Application/OpenCore/OpenCore.c
+++ b/Application/OpenCore/OpenCore.c
@@ -91,7 +91,23 @@ OcStartImage (
 {
   EFI_STATUS                       Status;
   EFI_CONSOLE_CONTROL_SCREEN_MODE  OldMode;
+  CHAR16                           *DevicePathText;
 
+    if (Chosen->DevicePath != NULL) {
+      DevicePathText = ConvertDevicePathToText (Chosen->DevicePath, FALSE, FALSE);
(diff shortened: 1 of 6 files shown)


dreamwhite:
I think that this type of change request has been already investigated by the devs. For more info read [here](https://www.insanelymac.com/forum/topic/350754-opencore-general-discussion/page/19/#comment-2675604)

Please explain why such changes would be helpful. Your description is just a summary of the changes without any further explanation.

And please do not make the false assumption that OpenCore isn't capable of booting Windows or other OSs just because the end-user hasn't properly read `Configuration.pdf`, especially the `12.3 Windows support` paragraph.

kernel-dev:
If I completely understood the whole purpose behind your commit:

If you are worried about OpenCore injecting SSDTs for Windows that are designed for macOS, I would advise you to look at the `_OSI` method check as defined per the [ACPI specification here](https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/05_ACPI_Software_Programming_Model/ACPI_Software_Programming_Model.html?highlight=_osi#osi-operating-system-interfaces). There is no real reason that OpenCore _should not_ allow for SSDT injections on Windows. If a developer wishes to write an SSDT that is only meant to target macOS, they can do so by making an `_OSI` check, and **not** by OpenCore defaulting to only allow for SSDT injections for macOS, or an internal quirk to enable patching for all platforms.

However, if I had misinterpreted your message's meaning, then please feel free to correct me.

PR #404, opened 2022-11-30
Minor Tweaks to Docs

dakanji:
Some minor tweaks for clarity

Mainly:
- Changelog.md: Reworded to better convey intent given that there are no MP51 specific GPUs as was implied
- Configuration.tex: Reworded for clarity on VM and better outline support scope on Apple Hardware
  - Previously _suggested_ support only goes as far as MP51 but MP31 is effectively the  boundary


mikebeaton:
Apologies no ping or request for review, in fact this change is specific to MacPro5,1 (or, at least, to 144.0.0.0.0 firmware) so the revised wording still might not match my original intent with the wording.

PR #406, opened 2022-12-01
[macrecovery] Guess downloaded recovery version

dreamwhite:
The aim of this PR is to improve my previous PR #394.
More specifically, with the proposed changes, `macrecovery` can write onto `com.apple.recovery.boot/.contentDetails` the version of the downloaded OS, obtained with the following steps:

1. mount and get the mount point of the downloaded `BaseSystem.dmg` available in `com.apple.recovery.boot` folder (by using `hdiutil mount -plist com.apple.recovery.boot/BaseSystem.dmg`)
2. get the `DTPlatformVersion` in `Install macOS XXXXX.app/Contents/Info.plist` key
3. write onto `com.apple.recovery.boot/.contentDetails` 

Please note that this may slow the execution times of the script itself, mainly due to `hdiutil`. Don't know if there are different ways of mounting a DMG in macOS using the command line, but that's what I used back in the past. I'm opened to suggestions

Kind regards
dreamwhite

### DIFF
--- a/Utilities/macrecovery/macrecovery.py
+++ b/Utilities/macrecovery/macrecovery.py
@@ -12,6 +12,7 @@ import hashlib
 import json
 import linecache
 import os
+import plistlib
 import random
 import struct
 import sys
@@ -288,6 +289,7 @@ def action_download(args):
     cnkpath = save_image(info[INFO_SIGN_LINK], info[INFO_SIGN_SESS], cnkname, args.outdir)
     try:
         verify_image(dmgpath, cnkpath)
+        get_basesystem_version(args)
         return 0
     except Exception as err:
         if isinstance(err, AssertionError) and str(err) == '':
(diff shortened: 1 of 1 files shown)


Utilities/macrecovery/macrecovery.py
```diff
@@ -453,6 +455,29 @@ def action_guess(args):
     print(f'UNKNOWN: Failed to determine supported models for MLB {mlb}!')
     return None
 
+def get_basesystem_version(args):
+    print('Extracting recovery version...')
+    os.system(f'hdiutil mount -plist {args.outdir}/BaseSystem.dmg > mount.plist')
```
williambj1, line 460: This breaks compatibility for Windows and Linux.

dreamwhite: O_O I completely forgot about it. Should I use a conditional like `if platform.system() == 'Darwin'` and execute the whole code, or whatever?

Andrey1970AppleLife:
The majority the user use this utility not from macOS.

dreamwhite:
> The majority the user use this utility not from macOS.

Yeah I noticed that. What about adding a condition like `if os is macos, then get the version from the dmg`?

PR #410, opened 2022-12-18
Introduce Docker setup for Dev Containers and CI

mhaeuser:
Docker allows to set up a portable development environment in a container. This can be useful to test Linux toolchains on macOS. Introduce such environments for EDK II and documentation development.

The current implementation supports both Dev Environments by Docker Desktop and Dev Containers by Visual Studio Code. To reduce the maintenance burden, use these environments for the Linux-based GitHub Actions as well. Non-x86 hosts, including Apple Silicon, are supported via GCC cross-compilation.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,70 +59,65 @@ jobs:
           tag: ${{ github.ref }}
           file_glob: true
 
-  build-linux-clangpdb-gcc5:
-    name: Linux CLANGPDB/GCC5
+  build-linux-clangpdb:
+    name: Linux CLANGPDB
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
 
-      - name: Install Dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get install nasm uuid-dev libssl-dev iasl
-          wget https://apt.llvm.org/llvm.sh
-          chmod +x llvm.sh
-          sudo ./llvm.sh 13
-          echo "/usr/lib/llvm-13/bin" >> $GITHUB_PATH
-
-      - name: CI Bootstrap
-        run: |
-          src=$(/usr/bin/curl -Lfs https://raw.githubusercontent.com/acidanthera/ocbuild/master/ci-bootstrap.sh) && eval "$src" || exit 1
-
-      - run: ./build_duet.tool
-      - run: ./build_oc.tool
+      - name: ./build_duet.tool
+        run: docker compose run -e TOOLCHAINS=CLANGPDB build-duet
+      - name: ./build_oc.tool
+        run: docker compose run -e TOOLCHAINS=CLANGPDB build-oc
 
       - name: Upload to Artifacts
         uses: actions/upload-artifact@v3
         with:
-          name: Linux CLANGPDB-GCC5 Artifacts
+          name: Linux CLANGPDB Artifacts
+          path: Binaries/*.zip
+
+  build-linux-gcc5:
+    name: Linux GCC5
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: ./build_duet.tool
+        run: docker compose run -e TOOLCHAINS=GCC5 build-duet
+      - name: ./build_oc.tool
+        run: docker compose run -e TOOLCHAINS=GCC5 build-oc
+
+      - name: Upload to Artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: Linux GCC5 Artifacts
           path: Binaries/*.zip
 
   build-linux-clangdwarf:
     name: Linux CLANGDWARF
     runs-on: ubuntu-22.04
-    env:
-      TOOLCHAINS: CLANGDWARF
     steps:
       - uses: actions/checkout@v3
 
-      - name: Install Dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get install nasm uuid-dev iasl doxygen texlive texlive-latex-extra
-          wget https://apt.llvm.org/llvm.sh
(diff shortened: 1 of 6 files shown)


Dockerfiles/oc-dev-docs/Dockerfile
```diff
@@ -0,0 +1,7 @@
+ARG OC_DEV_DOCS_BASE=ubuntu:22.04
+FROM $OC_DEV_DOCS_BASE
+
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && \
```
PMheart, line 5: Maybe replace with simply apt?

mhaeuser: Why?
1) apt is for end-users, while apt-get is for backends (this is a backend).
2) The existing CI code uses apt-get, not apt.

dreamwhite: I'm not a Linux expert nor I wanna judge your choices, but I wanna point out that:

> apt is for end-users, while apt-get is for backends (this is a backend).

Your statement isn't wrong but unless you're using old low-level commands (e.g. `apt-cache`, `apt-config` etc) you won't notice any practical difference between `apt` and `apt-get`.

> The existing CI code uses apt-get, not apt.

From what I saw in the existing CI code, there are only `update` and `install` comands in:

- [`analyze.yml/analyze-docs-linux`](https://github.com/acidanthera/OpenCorePkg/blob/6f25ba4f2c60d097f0ae54e8ecb765c6faa5e2db/.github/workflows/analyze.yml#L95-L96)
- [`analyze.yml/analyze-coverity`](https://github.com/acidanthera/OpenCorePkg/blob/6f25ba4f2c60d097f0ae54e8ecb765c6faa5e2db/.github/workflows/analyze.yml#L113-L114)
- [`build.yml/build-linux-clangpdb-gcc5`](https://github.com/acidanthera/OpenCorePkg/blob/6f25ba4f2c60d097f0ae54e8ecb765c6faa5e2db/.github/workflows/build.yml#L70-L71)
- [`build.yml/build-linux-clangdwarf`](https://github.com/acidanthera/OpenCorePkg/blob/6f25ba4f2c60d097f0ae54e8ecb765c6faa5e2db/.github/workflows/build.yml#L100-L101)
- [`uncrustify.yml/analyze-uncrustify`](https://github.com/acidanthera/OpenCorePkg/blob/6f25ba4f2c60d097f0ae54e8ecb765c6faa5e2db/.github/workflows/uncrustify.yml#L22-L23)

Therefore there shouldn't be any issue in switching from `apt-get` to `apt` unless you plan to use some old low-level commands as mentioned before.

P.S. I'd like to share this article which may better explain the changes between `apt` and `apt-get`: https://itsfoss.com/apt-vs-apt-get-difference/

Dockerfiles/oc-dev-edk2/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:22.04
+
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && \
+    apt-get install -y lsb-release wget software-properties-common gnupg build-essential nasm uuid-dev libssl-dev iasl curl git zip && \
+    { [ "$(dpkg --print-architecture)" == "amd64" ] || { apt-get install -y gcc-x86-64-linux-gnu && export GCC5_BIN=x86_64-linux-gnu- ; } ; } && echo "export GCC5_BIN=$GCC5_BIN" > ~/.edk2_rc.sh && echo ". ~/.edk2_rc.sh" > /etc/profile.d/edk2-gcc5.sh && \
+    wget https://apt.llvm.org/llvm.sh && chmod +x llvm.sh && ./llvm.sh 13 && rm -f llvm.sh && \
+    apt-get purge --auto-remove -y wget software-properties-common gnupg && \
+    rm -rf /var/lib/apt/lists/*
+ENV PATH="$PATH:/usr/lib/llvm-13/bin"
```
dreamwhite, line 10: If `llvm` version `13` gets deprecated/not available through the downloaded script how will you handle that?
Is caching the working `llvm.sh` a way to fix it? Eventually a folder (e.g. `docker-stuff` or whatever) with all the necessary files can be mounted within the container as a `ro` volume (using the proper syntax)

mhaeuser: We only support the most current LLVM release, this concern is not realistic.

.github/workflows/build.yml
```diff
@@ -59,71 +59,66 @@ jobs:
...
-      - run: ./build_duet.tool
-      - run: ./build_oc.tool
+  build-linux-gcc5:
+    name: Linux GCC5
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: ./build_duet.tool
+        run: docker compose run -e TOOLCHAINS=GCC5 build-duet
```
vit9696, line 86: Are these running in separate containers? If so, build-oc will not see build-duet results, and thus builds will be incomplete.

Dockerfiles/oc-dev-docs/Dockerfile
mhaeuser: Why would we use a tool that is explicitly not built for the scenario we are in? apt does not even have some kind of stable output, unlike apt-get, which is sometimes parsed by scripts. Yes, we do not parse here, but there is neither an issue with apt-get, nor any advantage with apt here.

.github/workflows/build.yml
mhaeuser: They indeed run in separate containers, but they both mount the host folder, so it should work.

vit9696:
Looks good to me now.

PR #414, opened 2023-01-03
EnableGop: Provide standalone GOP driver for EFI-era Macs

mikebeaton:
So, this is working for me and is confirmed working on a couple of other graphics cards.

**EDIT:** Have tested on considerably more machines and GPUs, and modified the insertion technique and the Apple firmware unlock technique.


Staging/EnableGop/EnableGop.c
```diff
@@ -0,0 +1,233 @@
...
+STATIC EFI_GET_MEMORY_SPACE_MAP  mOriginalGetMemorySpaceMap = NULL;
+
+STATIC
+VOID
+LogStage (
+  CHAR8  Stage
+  )
+{
+  UINT8   One    = 1;
+  CHAR16  Name[] = L"efi-gop-stage0";
```
mhaeuser, line 30: I'd just use ``#define`` - if the optimizer fucks up, there will be a pointless copy on the stack.

Library/OcBootManagementLib/OcBootManagementLib.c
```diff
@@ -525,26 +536,49 @@ OcRunFirmwareApplication (
...
-                  (VOID **)&FirmwareUI
+                  (VOID **)UIProtocol
                   );
 
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  if (UIVars != NULL) {
+    *UIVars = (VOID *)((UINT8 *)*UIProtocol + sizeof (APPLE_FIRMWARE_USER_INTERFACE_PROTOCOL));
```
mhaeuser, line 559: Imo write a wrapper structure with the protocol as first member and then use `BASE_CR`

Library/OcBootManagementLib/OcBootManagementLib.c
```diff
@@ -459,16 +459,34 @@ OcRunBootPicker (
...
 
+STATIC
+EFI_STATUS
+SetPickerEntryReason (
+  IN APPLE_PICKER_ENTRY_REASON  PickerEntryReason
+  )
+{
+  EFI_STATUS  Status;
+
+  Status = gRT->SetVariable (
```
mhaeuser, line 470: Can just return immediately

Library/OcDeviceMiscLib/ForgeUefi.c
```diff
@@ -25,6 +25,10 @@
 #include <Library/OcDeviceMiscLib.h>
 #include <Library/UefiBootServicesTableLib.h>
 
+#define __UEFI_MULTIPHASE_H__
+#define __PI_MULTIPHASE_H__
+#include <Pi/PiDxeCis.h>
```
mhaeuser, line 30: Drop these three, replace `Uefi.h` with `PiDxe.h`

Library/OcDeviceMiscLib/ForgeUefi.c
```diff
@@ -93,14 +110,34 @@ OcForgeUefiSupport (
     return EFI_INVALID_PARAMETER;
   }
 
-  NewBS = AllocateZeroPool (sizeof (EFI_BOOT_SERVICES));
-  if (NewBS == NULL) {
-    DEBUG ((DEBUG_INFO, "OCDM: Failed to allocate BS copy\n"));
-    return EFI_OUT_OF_RESOURCES;
+  if (Trash) {
+    Signature = *(UINT64 *)(&gBS->CreateEventEx);
```
mhaeuser, line 114: Does old Apple EFI allocate gBS and gDS consecutively and this checks for it? Can't you use the header size or something?

Library/OcDeviceMiscLib/ForgeUefi.c
```diff
@@ -93,14 +110,34 @@ OcForgeUefiSupport (
...
+      ));
+    NewBS = gBS;
+  } else {
+    NewBS = AllocateZeroPool (sizeof (EFI_BOOT_SERVICES));
+    if (NewBS == NULL) {
+      DEBUG ((DEBUG_INFO, "OCDM: Failed to allocate BS copy\n"));
+      return EFI_OUT_OF_RESOURCES;
+    }
+
+    CopyMem (NewBS, gBS, gBS->Hdr.HeaderSize);
```
mhaeuser, line 138: What if `gBS->Hdr.HeaderSize != sizeof (EFI_BOOT_SERVICES)` (perfectly legal when spec versions of the host and our driver mismatch)?

Staging/EnableGop/EnableGop.c
```diff
@@ -0,0 +1,233 @@
...
+                    Guid,
+                    NULL,
+                    &DataSize,
+                    NULL
+                    );
+
+  //
+  // Do not modify existing value.
+  //
+  if (Status == EFI_BUFFER_TOO_SMALL) {
```
mhaeuser, line 126: What if there is a different error? Imo check for "not success" (i.e., `!= EFI_NOT_FOUND` I think) rather than an explicit "failure" (i.e., var does exist and can be read successfully).

Staging/EnableGop/EnableGop.c
```diff
@@ -0,0 +1,233 @@
...
+    return EFI_SUCCESS;
+  }
+
+  Status = gRT->SetVariable (
```
mhaeuser, line 130: Can return immediately

Include/Acidanthera/Library/OcBootManagementLib.h
```diff
@@ -19,6 +20,12 @@
 #include <Protocol/LoadedImage.h>
 #include <Protocol/AppleBeepGen.h>
 #include <Protocol/OcAudio.h>
+#include <Protocol/GraphicsOutput.h>
+#include <Protocol/AppleUserInterface.h>
+
+#define __UEFI_MULTIPHASE_H__
+#define __PI_MULTIPHASE_H__
+#include <Pi/PiDxeCis.h>
```
mhaeuser, line 28: Remove these three, replace `Uefi.h` with `PiDxe.h` (makes me wonder whether this should live in a different file)

Library/OcBootManagementLib/OcBootManagementLib.c
```diff
@@ -525,26 +536,49 @@ OcRunFirmwareApplication (
 }
 
 EFI_STATUS
-OcLaunchAppleBootPicker (
-  VOID
+OcGetAppleFirmwareUI (
+  IN APPLE_FIRMWARE_USER_INTERFACE_PROTOCOL  **UIProtocol,
+  IN APPLE_FIRMWARE_UI_VARS                  **UIVars OPTIONAL
```
mhaeuser, line 541: Aren't both only `OUT`?

Library/OcBootManagementLib/OcBootManagementLib.c
mikebeaton: Done in commit to follow.

Library/OcDeviceMiscLib/ForgeUefi.c
mikebeaton: The layout is old Apple EFI, but it is directly from standard edk EFI code, in fact.

What is being checked for here is that the memory about to be trashed is - fortuitously - exactly that value: something which is only used when the memory is being loaded, in fact (when we check for references to this value throughout edk code).

The point being, I _have_ to trash exactly that QWORD of memory. I would have just trashed it and prayed... but in the targeted firmware I can confirm that it is harmless, and proceed.

Does this sound okay, given that explanation?

Library/OcDeviceMiscLib/ForgeUefi.c
mikebeaton: I believe we've checked this above, in line 109.

Staging/EnableGop/EnableGop.c
mikebeaton: I'm not convinced I agree. My logic, fwiw, is that I might as well attempt the write on anything other than "the variable exists already and can be read", no?

Staging/EnableGop/EnableGop.c
mikebeaton: I this case, I sort of thought I could let the compiler deal with that and this layout would make it easier to insert other (e.g. DEBUG) code if wanted. But changed in upcoming.

Include/Acidanthera/Library/OcBootManagementLib.h
mikebeaton: Seems I can used Uefi.h _and_ PiDxe.h in the same file, if I need to, with no hacks - so have done that. Thank you.

Library/OcBootManagementLib/OcBootManagementLib.c
mikebeaton: Yes, doh. Though with BASE_CR change this is now deleted.

Staging/EnableGop/EnableGop.c
mikebeaton: Also - I'll have to retest the logic of detecting whether we are injected or running as Driver#### if I modify this...!

Include/Acidanthera/Library/OcBootManagementLib.h
mikebeaton: (Let me know if any apparent problems with that, though as it compiles and works, I guess not.)

Library/OcBootManagementLib/OcBootManagementLib.c
```diff
@@ -564,10 +564,10 @@ OcUnlockAppleFirmwareUI (
     //
     // Also acts as partial check that the memory is not something else entirely.
     //
-    if ((UINTN)(UIVars->mGopAlreadyConnected) != 1) {
-      DEBUG ((DEBUG_WARN, "OC: Cannot force reconnect Apple GOP %u\n", UIVars->mGopAlreadyConnected));
+    if (UIVars->mGopAlreadyConnected.FullFieldValue != 1) {
```
mikebeaton, line 567: @mhaeuser - Have you got time to confirm that this looks okay too? I thought this was what my previous check was doing, but re-reading it, realised it was not.

The change is not as clean a way to deal with what is actually a bool (padded to a QWORD), but I believe is a better check that we are in the memory we expect to be in.

Library/OcDeviceMiscLib/ForgeUefi.c
mhaeuser: Yes, thanks. Is this commented somewhere and I missed it?

Library/OcDeviceMiscLib/ForgeUefi.c
mhaeuser: Ohh, I missed `CreateEventEx` is actually populated below. Ok.

Staging/EnableGop/EnableGop.c
mhaeuser: Well, if that makes sense to you, sure, I have no context how “bad” it is to have a false positive over a false negative.

Library/OcDeviceMiscLib/ForgeUefi.c
mikebeaton: No it's not - I'll add a comment. (I get the idea the existing OC code style is "don't comment if it's obvious what it's doing", where 'obvious' tends to mean, "read it 10 times and scratch your head for a week" - so I try not to over-comment! ;) )

mhaeuser: No, the existing code style is “You have two weeks off-ish? Write a Mach-O dynamic linker, fix most of the bugs, integrate it with a separate prelinkedkernel mod logic into a working kext injection and fuzz it” :) When the deadline is not yesterday, I strongly prefer comments, I think I tried with the crypto lib

Library/OcBootManagementLib/OcBootManagementLib.c
mikebeaton: Thanks for offline discussion - modified and marking as closed.

Include/Acidanthera/Library/OcBootManagementLib.h
```diff
@@ -2075,6 +2078,40 @@ OcAddEntriesFromBootEntryProtocol (
   IN     BOOLEAN        CreateForHotKey
   );
 
+#pragma pack (push, 1)
+
+/**
+  Useful internal vars from firmware UI protocol (cMP 144.0.0.0.0 ROM).
+**/
+typedef PACKED struct {
+  BOOLEAN                         mGopAlreadyConnected;
```
vit9696, line 2087: This is really packed with pointers misaligned?

Include/Acidanthera/Library/OcBootManagementLib.h
```diff
@@ -2075,6 +2078,40 @@ OcAddEntriesFromBootEntryProtocol (
...
+#pragma pack (pop)
+
+/**
+  Force Apple Firmware UI to reconnect to current console GOP.
+
+  @retval EFI_SUCCESS   Picker was successfully unlocked.
+  @retval other         Compatible firmware UI protocol for unlock could not be found.
+**/
+EFI_STATUS
+OcUnlockAppleFirmwareUI (
```
vit9696, line 2111: Maybe rename to OcReconnectFirmwareUI? There is no unlocking here.

Library/OcBootManagementLib/OcBootManagementLib.c
```diff
@@ -544,27 +551,39 @@ OcLaunchAppleBootPicker (
...
+
     //
-    // Location of relevant byte variable within loaded driver.
+    // Also acts as partial check that the memory is not something else entirely.
     //
-    aGopAlreadyConnected = (VOID *)((UINT8 *)FirmwareUI + sizeof (APPLE_FIRMWARE_USER_INTERFACE_PROTOCOL));
-
-    if (*aGopAlreadyConnected != 1) {
-      DEBUG ((DEBUG_WARN, "OC: Cannot force reconnect Apple GOP %u\n", *aGopAlreadyConnected));
+    if (*((UINTN *)(&UIVars->mGopAlreadyConnected)) != 1) {
```
vit9696, line 567: This makes me even more convinced it is not a boolean. Could you provide some efi file?
Also, that would need ReadUnaligned...

Library/OcBootManagementLib/OcBootManagementLib.c
```diff
@@ -544,27 +551,39 @@ OcLaunchAppleBootPicker (
...
-      "OC: Launch Apple picker incompatible FirmwareUI protocol revision %u != %u\n",
+      "OC: Unlock Apple picker incompatible FirmwareUI protocol revision %u != %u\n",
       FirmwareUI->Revision,
       APPLE_FIRMWARE_USER_INTERFACE_PROTOCOL_REVISION
       ));
     Status = EFI_UNSUPPORTED;
   }
 
   if (!EFI_ERROR (Status)) {
+    UIVars = &BASE_CR (FirmwareUI, APPLE_FIRMWARE_UI_LAYOUT, UIProtocol)->UIVars;
```
vit9696, line 562: Does it work with 32-bit efi? In terms of at least not breaking.

Library/OcDeviceMiscLib/ForgeUefi.c
```diff
@@ -93,14 +115,34 @@ OcForgeUefiSupport (
     return EFI_INVALID_PARAMETER;
   }
 
-  NewBS = AllocateZeroPool (sizeof (EFI_BOOT_SERVICES));
-  if (NewBS == NULL) {
-    DEBUG ((DEBUG_INFO, "OCDM: Failed to allocate BS copy\n"));
-    return EFI_OUT_OF_RESOURCES;
+  if (Trash) {
+    Signature = *(UINT64 *)(&gBS->CreateEventEx);
+    if (Signature != DXE_SERVICES_SIGNATURE) {
```
vit9696, line 120: Does this ever happen?

Staging/EnableGop/EnableGop.c
```diff
@@ -0,0 +1,231 @@
...
+{
+  UINT8   One    = 1;
+  CHAR16  Name[] = L"efi-gop-stage0";
+
+  if (FeaturePcdGet (PcdEnableGopDirect)) {
+    Name[6] = L'd';
+  }
+
+  Name[L_STR_LEN (Name) - 1] = Stage;
+  OcSetSystemVariable (
```
vit9696, line 37: Should really be debug-only no to wear off NVRAM, to be honest.

Staging/EnableGop/EnableGop.c
```diff
@@ -0,0 +1,231 @@
...
+}
+
+STATIC
+EFI_STATUS
+WrapGetMemorySpaceMap (
+  VOID
+  )
+{
+  mOriginalGetMemorySpaceMap = gDS->GetMemorySpaceMap;
+  gDS->GetMemorySpaceMap     = WrappedGetMemorySpaceMap;
```
vit9696, line 195: No checksum here?

Library/OcDeviceMiscLib/ForgeUefi.c
mikebeaton: No. I thought it was safer to check that the memory is safe to trash. See comment at start of this function.

mikebeaton: i.e. it would only ever happen if injected/used on some other firmware, but I guess _that_ could definitely happen.

Staging/EnableGop/EnableGop.c
mikebeaton: This is called from a driver in which OcVariableInit is never called ... maybe it should be, or maybe it does not matter ... so this will actually be non-volatile always, I believe.

Is that okay?

Staging/EnableGop/EnableGop.c
mikebeaton: I saw you'd added all this check-summing, but I've never seen it matter - maybe is required only on different/newer firmware? If it is required somewhere, I agree I should add it.

(I wrote tons of trapping, of gBS, gRT and this, while developing this, and never checksummed - although ... oh ... I did get one crash when trapping early ... maybe that was it!)

mhaeuser: This is definitely *sometimes* checked. Probably not for gDS, because this dies when exiting boot services, but definitely for gRT. This is a clear case of "just do it". :)

vit9696: I think we did check for gDS checksum in our Linux patches when we added memory protection code to the efistub. Please fix.

Library/OcDeviceMiscLib/ForgeUefi.c
vit9696: Just I would rather abort and try to reallocate. Reallocting boot services will likely cause hard to debug hibernation or boot issues, so better not to put the mines for us here.

Staging/EnableGop/EnableGop.c
mikebeaton: PS Planning to remove when the driver is a bit more established as working - though I suppose I could make it debug code only and keep it.

vit9696: Maybe really make it DEBUG_CODE_BEGIN/DEBUG_CODE_END? This will be handy…

Include/Acidanthera/Library/OcBootManagementLib.h
mikebeaton: Doh - sleep deprived when add this - after actually using those table pointers for debug early (when it wasn't marked packed, ofc). Thank you!

Library/OcDeviceMiscLib/ForgeUefi.c
mikebeaton: Hang on though - if "reallocting boot services will likely cause hard to debug hibernation or boot issues", then my best bet is to abort the forge attempt and do nothing (won't forge UEFI, driver probably won't work, but no such hard to debug problems). Which is what the code does, no? (What have I missed?!)

vit9696: Yes, abort and do nothing. I only wanted to remove the reallocation part, it will not do us any good I fear.

vit9696: Disregard this comment. Realised that I was wrong with the suggestions above. The issue is that the driver in DXE mode cannot reallocate the BS table and thus needs to patch it.

Library/OcBootManagementLib/OcBootManagementLib.c
mikebeaton: Think this is resolved in offline chat - will fix the packed issue - ty!

Library/OcBootManagementLib/OcBootManagementLib.c
mikebeaton: I think this was also related to the packed error - so resolving

Include/Acidanthera/Library/OcBootManagementLib.h
mikebeaton: Done in upcoming commit, so resolving.

Include/Acidanthera/Library/OcBootManagementLib.h
mikebeaton: Hi - hmm, the thing about `OcReconnectAppleFirmwareUI` (now that I try renaming it, and looking at the code) is that it doesn't reconnect GOP at the time of the call, instead it 'unlocks' it, or opens it up again, so that it will reconnect, next time it's asked to. Hence previous name `OcReconnectAppleFirmwareUI`. Which also fits with my naming of the method in the firmware UI protocol interface, `ConnectGop`, which is the method which is affected by this.

vit9696: Well, ok, I guess it is fine to leave it as is then.

Include/Acidanthera/Library/OcBootManagementLib.h
```diff
@@ -2075,6 +2078,37 @@ OcAddEntriesFromBootEntryProtocol (
   IN     BOOLEAN        CreateForHotKey
   );
 
+/**
+  Useful internal vars from firmware UI protocol (cMP 144.0.0.0.0 ROM).
+**/
+typedef PACKED struct {
```
vit9696, line 2084: This is still packed. Is it resolved or not?

mikebeaton: Oh for crying out loud, I'm editing while knackered and now ill 🤢 - very sorry. No it is wrong, thank you for rechecking. (I removed one of the two ways of specifying packed, only. I'll remove this one too, and write some debug code to retest before pushing again.)

Library/OcDeviceMiscLib/ForgeUefi.c
```diff
@@ -67,21 +67,43 @@ OcCreateEventEx (
...
+    &gBS
     ));
 
+  if (!Forge) {
+    return EFI_SUCCESS;
+  }
+
   //
   // Already too new.
   //
```
dakanji, line 109: ForgeUEFI should guard against being passed already forged instances below.

That is, something similar to the `Spoof Proofing` added to OpenVariableRuntimeDxe here: https://github.com/acidanthera/OpenCorePkg/pull/405

mikebeaton: It already does

dakanji: You mean it will filter this spoof example out:
- SystemTable = UEFI 2.10
- BootServices = EFI 1.10
- RunTimeServices = EFI 1.10

mikebeaton: Well expand the code below in GitHub and look, I would say clearly, yes, it will filter that. I did not modify these checks.

dakanji: I only see a check on SystemTable against UEFI 2.3 (which would pass) and a check against BootServices size (which is not fool proof) but I suppose I am missing the obvious.

I know you did not change this element btw. Just thought you might want to improve the function given that you are making changes there. Not needed in any case given what you said though.

mikebeaton: @dakanji, The _immediately_ following lines here are:

```
  if (gST->Hdr.Revision >= EFI_2_30_SYSTEM_TABLE_REVISION) {
    return EFI_ALREADY_STARTED;
  }
```

which means, given the data you mentioned, OcForgeUefiSupport will abort (i.e. not forge anything) immediately, returning EFI_ALREADY_STARTED - which I still make the correct behaviour, no? Or are you saying you _want_ it to forge in the case you have given?

dakanji: I saw that but in the case I posted, gST at 2.10 would not abort here (2.30 and greater abort).

mikebeaton: Oh dear, you're right, of course, but then I'm still not quite sure what we'd want it to do, in this case.

Are you stating that, as far as you know, the above has example forged elements, e.g. CreateEventEx, within gBS, even though the version of gBS has not been updated?

mikebeaton: I saw your PM with more detail - the answer to my question is yes - thank you.

We/I do need to process what we want to happen here, though.

mikebeaton: Mind you, we are still going to not forge, for the firmware you've mentioned, based on the gBS header size, right? (Feel free to set me straight again!)

dakanji: I think testing all three elements and aborting if not all UEFI 2.x is the best course of action.

Btw, it doesn't really matter that in this case, BootServices seems to have had CreateEventEx added.
The mere presence of version inconsistency is the indication of potential issues.

This was the same reason for checking them all in the driver.

mikebeaton: I now fully understand your reason for [checking them all](https://github.com/acidanthera/OpenCorePkg/pull/405) - thank you. But I'm still not sure checking them all is right...

Maybe we should do something like:

 - Check that we do not already have the minimum UEFI version which would provide the to-be-spoofed item in gST, abort if we do
 - Check that we do not already have the minimum UEFI version which would provide the to-be-spoofed item in gBS (or gRT, when spoofing gRT), abort if we do
 - Check that we do not already have a header size large enough to contain the to-be-spoofed item, in the header of the table we are spoofing (i.e. gBS or gRT, but only the one we are spoofing), abort if we do
 - Othwerwise, spoof (or, as the case is in the emu nvram code, provide in the first place)
 
This would mean some changes to [your check](https://github.com/acidanthera/OpenCorePkg/pull/405) - namely removing the check on gBS version, since there we are spoofing gRT; and some changes to the checks here - namely changing the version check down to 2+, even if we eventually spoof to 2.3.

I'll do a PR of these changes, to get some feedback from yourself and Vit. (Though feel free to feedback here, first.)

dakanji: I don't think the listed changes covers anything that is not already covered or might need to be but will leave things to gurus.

> some changes to the checks here - namely changing the version check down to 2+
I similarly don't think this is as good, or as comprehensive (I can imagine situations that will slip through) as a triple check, but will defer.

mikebeaton: I think maybe you are thinking only of changes that help when you want to avoid spoofing, where I'm trying to also think of changes that help when you want to avoid NOT spoofing?

mikebeaton: Anyway, if I do a PR, we can see what you, me, and gurus think!

dakanji: > I think maybe you are thinking only of changes that help when you want to avoid spoofing

Yes. I am indeed thinking that if someone has been spoofing beforehand, then further spoofing should be avoided.
Gurus, such as yourself, may have of course have other considerations!

mikebeaton:
Have tested on considerably more machines and GPUs, and modified the insertion technique and the Apple firmware unlock technique.

Ready for re-review, I believe, tyvm.

Library/OcDeviceMiscLib/ForgeUefi.c
mikebeaton: As currently provided, this _will_ replace an earlier forge to 2.0 <= UEFI < 2.3. I believe this is desirable in some cases and harmless in others.

LeniReinfe:
unsuccessful.  Doesn't work on RX580 (5.1 (2010) MX25L3205D/MX25L3208D)

mikebeaton:
There are definitely success reports for RX580 here: https://forums.macrumors.com/threads/pre-opencore-gop-support-for-efi-era-imacs-and-mac-pros.2378942/

I would post on that thread, we do not offer support as such within the repo and bugtracker - they are for tracking and fixing bugs - but I and other users of the driver offer support on the above thread.

Please do describe what you tried, and what results you got. Just "doesn't work" is not a great bug report - nor even a great way to request help, over on the support forum!

PR #415, opened 2023-01-06
DeviceMiscLib/OpenVariableRuntime: Improve response to previous UEFI forging

mikebeaton:
In response to recent discussions with @dakanji [here](https://github.com/acidanthera/OpenCorePkg/pull/414) and [here](https://github.com/acidanthera/OpenCorePkg/pull/405), I have attempted to improve the responsiveness to potential prior (possibly incomplete) UEFI forging, either added to firmware or in other earlier code.

By the way, the interesting example of a prior, incomplete 'forge', of which @dakanji has a valid example, has the following:
 - gST forged to version 2.1
 - gBS version unforged, so version 1.1, but header size is as 2.1 and gBS interface is believed to contain a working CreateEventEx
 - gRT version unforged, version 1.1, and unchanged header size

Given the disagreements over the best approach to dealing with prior forges like this, I'd appreciate another pair of eyes if @vit9696 or @mhaeuser had a moment.

cc @joevt

### DIFF
--- a/Library/OcDeviceMiscLib/ForgeUefi.c
+++ b/Library/OcDeviceMiscLib/ForgeUefi.c
@@ -85,7 +85,7 @@ OcForgeUefiSupport (
   //
   // Already too new.
   //
-  if (gST->Hdr.Revision >= EFI_2_30_SYSTEM_TABLE_REVISION) {
+  if (gST->Hdr.Revision >= EFI_2_00_SYSTEM_TABLE_REVISION) {
     return EFI_ALREADY_STARTED;
   }
 
--- a/Platform/OpenVariableRuntimeDxe/VariableDxe.c
+++ b/Platform/OpenVariableRuntimeDxe/VariableDxe.c
@@ -648,10 +648,8 @@ VariableServiceInitialize (
   )
 {
   EFI_STATUS           Status;
-  UINTN                OffsetQVI;
-  UINTN                HeaderQVI;
-  EFI_EVENT            EndOfDxeEvent;
   EFI_EVENT            ReadyToBootEvent;
+  EFI_EVENT            EndOfDxeEvent;
   EFI_CREATE_EVENT_EX  OriginalCreateEventEx;
 
   SaveAcpiGlobalVariable (SystemTable);
@@ -685,19 +683,16 @@ VariableServiceInitialize (
   SystemTable->RuntimeServices->SetVariable         = VariableServiceSetVariable;
   //
   // Avoid setting UEFI 2.x interface member on EFI 1.x.
+  // Note 1: It is important to set this when possible. We have completely replaced
+  // SetVariable and GetVariable, and QueryVariableInfo must be consistent with them
+  // if it exists at all.
+  // Note 2: We attempt to allow for a previous inconsistently faked UEFI 2+ by
+  // checking that space in the interface is actually available.
   //
-  // First test all systable elements as some may have been spoofed and pass a limited element check
-  // Then check that QueryVariableInfo is specifically available before setting the interface member
-  //
-  if (  ((SystemTable->Hdr.Revision >> 16U) > 1)
-     && ((SystemTable->BootServices->Hdr.Revision >> 16U) > 1)
-     && ((SystemTable->RuntimeServices->Hdr.Revision >> 16U) > 1))
+  if (  (SystemTable->RuntimeServices->Hdr.Revision >= EFI_2_00_SYSTEM_TABLE_REVISION)
+     && (SystemTable->RuntimeServices->Hdr.HeaderSize > OFFSET_OF (EFI_RUNTIME_SERVICES, QueryVariableInfo)))
   {
-    OffsetQVI = OFFSET_OF (EFI_RUNTIME_SERVICES, QueryVariableInfo);
-    HeaderQVI = OffsetQVI + sizeof (SystemTable->RuntimeServices->QueryVariableInfo);
-    if (SystemTable->RuntimeServices->Hdr.HeaderSize >= HeaderQVI) {
-      SystemTable->RuntimeServices->QueryVariableInfo = VariableServiceQueryVariableInfo;
-    }
+    SystemTable->RuntimeServices->QueryVariableInfo = VariableServiceQueryVariableInfo;
   }
 
   //


Platform/OpenVariableRuntimeDxe/VariableDxe.c
```diff
@@ -685,19 +685,14 @@ VariableServiceInitialize (
...
+  // We attempt allow for a previous inconsistently faked UEFI 2+.
   //
-  // First test all systable elements as some may have been spoofed and pass a limited element check
-  // Then check that QueryVariableInfo is specifically available before setting the interface member
-  //
-  if (  ((SystemTable->Hdr.Revision >> 16U) > 1)
-     && ((SystemTable->BootServices->Hdr.Revision >> 16U) > 1)
-     && ((SystemTable->RuntimeServices->Hdr.Revision >> 16U) > 1))
+  if (  (SystemTable->RuntimeServices->Hdr.Revision >= EFI_2_00_SYSTEM_TABLE_REVISION)
+     && (SystemTable->RuntimeServices->Hdr.HeaderSize > OFFSET_OF (EFI_RUNTIME_SERVICES, QueryVariableInfo)))
```
mikebeaton, line 693: I believe https://github.com/acidanthera/OpenCorePkg/pull/405 was over-cautious. In fact, the important thing here is defintely to provide QueryVariableInfo if there is space to put it, not to avoid providing it if there is any sign of a previous forge.

Library/OcDeviceMiscLib/ForgeUefi.c
```diff
@@ -85,7 +85,7 @@ OcForgeUefiSupport (
   //
   // Already too new.
   //
-  if (gST->Hdr.Revision >= EFI_2_30_SYSTEM_TABLE_REVISION) {
+  if (gST->Hdr.Revision >= EFI_2_00_SYSTEM_TABLE_REVISION) {
```
mikebeaton, line 88: Since CreateEventEx is defined at 2.0 upwards, this check currently seems safer, to me (even though we fake to 2.30 below) - and would deal with the firmware which @dakanji found.

Platform/OpenVariableRuntimeDxe/VariableDxe.c
```diff
@@ -648,10 +648,8 @@ VariableServiceInitialize (
   )
 {
   EFI_STATUS           Status;
-  UINTN                OffsetQVI;
-  UINTN                HeaderQVI;
-  EFI_EVENT            EndOfDxeEvent;
   EFI_EVENT            ReadyToBootEvent;
+  EFI_EVENT            EndOfDxeEvent;
```
mikebeaton, line 652: (These now match the original order, in the file from which this code was derived, again.)

Platform/OpenVariableRuntimeDxe/VariableDxe.c
mikebeaton: I have avoided also looking at SystemTable->Hdr.Revision (e.g. perhaps trying to fake if _either_ version is 2 or above and there is enough space to put the pointer), because of the existence of code like this:

https://github.com/tianocore/edk/blob/master/Foundation/Efi/Include/EfiApi.h#L946-L960

Library/OcDeviceMiscLib/ForgeUefi.c
mikebeaton: I could also add, I don't think there was any real problem with replacing one working CreateEventEx with another one (as the previous code would have done, in the firmware @dakanji found), but there's also (at least AFAICT ATMO) no real need to.

mikebeaton: If anyone knows a reason why we definitely _should_ fake, if we find a version below 2.30 - which was perhaps implicit but uncommented in the code before? - then please do say so!

joevt: I think 2.3 was required by my Nvidia Titan X (Maxwell) GOP driver - or somebody's nvidia GPU. I don't remember which thread or conversation this was discussed.
 https://github.com/acidanthera/bugtracker/issues/1280

Oh wait, I found this:

> joevt Nov 4, 2020
> I did the same test `#5` but this time I used the LoadPCIRom command in the EFI Shell to load some GPU roms from the EFI partition for GPUs that I do not have installed (and also the rom for GTX 680 Mac which already has it's driver in control). They all add some handles and protocols (DriverBinding) except the unflashed Titan X (Maxwell) rom which returned an error:
> ```
> LoadPciRom: Start image `#1` error - Incompatible Version
> loadPciRom: Image titanx.rom loaded - Not Found
> ```
> The Shell is at handle 16D so the drivers are after that (starting at 16F).
> 
> I showed at Sunday at 2:07 AM how to use the EFI Shell to extract a rom. I'll extract the GTX 1070 tomorrow and see what that does. I should also see how the LoadPCIRom command behaves on the Z-170 motherboards I have.
> 
> Anyway, back to the error, it could be from this code:
> [edk2/DriverEntryPoint.c at 3806e1fd139775610d8f2e7541a916c3a91ad989 · tianocore/edk2](https://github.com/tianocore/edk2/blob/3806e1fd139775610d8f2e7541a916c3a91ad989/MdePkg/Library/UefiDriverEntryPoint/DriverEntryPoint.c#L102)
>
> The Revision of EFI for my MacPro3,1 is shown in the dmem output as 0001000A (which I guess means 1.10 according to the output of the ver command).
> My Z-170X Gaming 7 has revision 00020032 (UEFI v2.50).
> My Parallels Version 16.1.0 (48950) VM has 00020046 (UEFI v2.70).
> 
> I loaded the TitanX rom successfully into the Parallels VM.
> One interesting thing is the handle has a SupportedEfiSpecVersion protocol 0002000A (UEFI v2.10) which I guess is the minimum Revision required. GOP protocol doesn't get added to the GPU unless the GPU exists (which the added DriverBinding protocol will look out for).
> 
> I suppose one hack you could try is alter the memory pointed to by system table - change Revision from 0001000A to 0002000A. This might work if the driver doesn't try to use gBS->CreateEventEx. If it does, then you would need to add CreateEventEx function pointer to Boot Services (not sure how to do that - need to see how BS is allocated in EFI 1.1 and see if it can be increased in size). There may be other UEFI 2.10 differences that the driver relies on.

So if my Titan X only requires 2.10 then maybe it was for someone else's GPU that we selected 2.3.

edk2 header says `QueryVariableInfo` was part of UEFI 2.0 spec, so only the version number would need to be changed in that case. The UEFI spec is now up to 2.10 but I think we want to choose the minimum version that works for all GPUs that we can use GOP from.

After that post:

> joevt Nov 5, 2020
> I was looking at my unflashed GTX 1070 in Z170N-Gaming 5 motherboard. I noticed that it was using a CsmVideo driver "BIOS[INT10] Video Driver" to manage the GPU and GOP driver as a child of the GPU because CSM was enabled in BIOS. Disabling CSM causes the GOP driver of the GPU to be used instead. With the GOP driver, you get more resolutions, and the resolution is 16:9 instead of ugly 4:3 (on 4K display).
> 
> Anyway, this means a BIOS compatible GOP driver is another option (but maybe more difficult). Most every GPU has a BIOS driver starting way back in the olden days - so any GPU could be used for startup screen (imagine using really old PCI GPUs - or AGP GPU if you can find a PCIe to AGP bridge - a company named Albatron was said to have made one - but I don't know if it was ever sold).
> 
> Back to using UEFI GOP drivers, I downloaded the `GOPUpd_v1.9.6.5.rar` to look at the roms. Using the `loadpciroms * > loadpciroms_all.txt` command to load all the roms at once, all of them load successfully in my Parallels VM as expected. Of the new Nvidia ROMs, only the GV100 [TITAN V] and TU104 [GeForce RTX 2080 Rev. A] roms would load on my MacPro3,1. All the AMD roms and my old Nvidia roms loaded successfully. You can view the errors and the loaded handles in the attached files.
> 
> All the roms that didn't load gave the `Incompatible Version` error (as seen in the `loadpcirom_all.txt` file). the dh.txt file shows which drivers loaded (10 of them for the MacPro3,1 and 23 in the Parallels VM). The SupportedEfiSpecVersion protocol of UEFI 2.x is called DriverEFIVersion in EFI 1.1. If there's an Incompatible Version then that version is probably the one being compared to the Revision in the System Table.
> 
> I used the `mm` command to change the Revision from `1.1` to `2.1`. Now when I try to load my titanx.rom with `loadpcirom`, instead of giving an error, it just hangs. I suspect the driver is calling `gBS->CreateEventEx` as this was how the XhciDxe driver behaves when I was working on it to make it work for EFI 1.1. So the next step is to add a `CreateEventEx` function to the Boot Services table.
> 
> Here's the change I made to Xhci.c for XhciDxe
> ```
> #if(0)
>   // For UEFI 2.0 and the future use CreateEventEx
>   Status = gBS->CreateEventEx (
>                   EVT_NOTIFY_SIGNAL,
>                   TPL_NOTIFY,
>                   XhcExitBootService,
>                   Xhc,
>                   &gEfiEventExitBootServicesGuid,
>                   &Xhc->ExitBootServiceEvent
>                   );
> #else
>   // prior to UEFI 2.0 use CreateEvent
>   Status = gBS->CreateEvent (
>                   EVT_SIGNAL_EXIT_BOOT_SERVICES,
>                   TPL_NOTIFY,
>                   XhcExitBootService,
>                   Xhc,
>                   &Xhc->ExitBootServiceEvent
>                   );
> #endif
> ```
> Events other than gEfiEventExitBootServicesGuid will need to be created differently or not created at all.
> 

Versions in the `dh` output:
```
SupportedEfiSpecVersion(0x0002000A) ComponentName2 ComponentName DriverBinding ImageDevicePath(..9872-474D-AE62-A65F4651B971)) LoadedImage(PrlVideoDxe) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(..\roms\nv_gop_GF10x.efirom[0]) LoadedImage(FS0:\roms\nv_gop_GF10x.efirom[0]) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(..\roms\nv_gop_GF119.efirom[0]) LoadedImage(FS0:\roms\nv_gop_GF119.efirom[0]) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(..\roms\nv_gop_GK1xx.efirom[0]) LoadedImage(nvgop-gk1xx) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(..\roms\nv_gop_GM1xx.efirom[0]) LoadedImage(nvgop-gmxxx) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(..\roms\nv_gop_GM2xx.efirom[0]) LoadedImage(nvgop-gm2xx) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(..\roms\nv_gop_GP1xx.efirom[0]) LoadedImage(nvgop-gp1xx) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(..\roms\nv_gop_GT21x.efirom[0]) LoadedImage(nvgop-gt21x) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(..\roms\nv_gop_GV1xx.efirom[0]) LoadedImage(nvgop-gv1xx) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(..\roms\nv_gop_TU1xx.efirom[0]) LoadedImage(nvgop-tu1xx) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(..nv_gop_GK1xx_multi.efirom[0]) LoadedImage(FS0:\roms\nv_gop_GK1xx_multi.efirom[0]) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(..s\nv_gop_GK1xx_MXM.efirom[0]) LoadedImage(nvgop-gk1xx_mxm) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(..s\nv_gop_GM1xx_MXM.efirom[0]) LoadedImage(nvgop-gmxxx_mxm) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(FS0:\roms\gtx1070.rom[1]) LoadedImage(nvgop-gp1xx) 
SupportedEfiSpecVersion(0x0002000A) ComponentName2 DriverBinding ImageDevicePath(FS0:\roms\titanx.rom[1]) LoadedImage(nvgop-gm2xx) 
SupportedEfiSpecVersion(0x0002001E) ComponentName2 DriverBinding ImageDevicePath(..:\roms\amd_gop_mcu.efirom[0]) LoadedImage(FS0:\roms\amd_gop_mcu.efirom[0]) 
SupportedEfiSpecVersion(0x0002001E) ComponentName2 DriverBinding ImageDevicePath(..:\roms\amd_gop_mod.efirom[0]) LoadedImage(FS0:\roms\amd_gop_mod.efirom[0]) 
SupportedEfiSpecVersion(0x0002001E) ComponentName2 DriverBinding ImageDevicePath(..\roms\amd_gop_vega.efirom[0]) LoadedImage(FS0:\roms\amd_gop_vega.efirom[0]) 
SupportedEfiSpecVersion(0x0002001E) ComponentName2 DriverBinding ImageDevicePath(..amd_gop_1.57.0.0.0.efirom[0]) LoadedImage(FS0:\roms\amd_gop_1.57.0.0.0.efirom[0]) 
SupportedEfiSpecVersion(0x0002001E) ComponentName2 DriverBinding ImageDevicePath(FS0:\roms\amd_gop.efirom[0]) LoadedImage(FS0:\roms\amd_gop.efirom[0]) 
SupportedEfiSpecVersion(0x0002001E) ComponentName2 DriverBinding ImageDevicePath(FS0:\roms\W5700.rom[1]) LoadedImage(FS0:\roms\W5700.rom[1]) 
SupportedEfiSpecVersion(0x00020028) ComponentName2 ComponentName DriverBinding ImageDevicePath(..53CF-46A3-A6A9-73C34A6E5EE3)) LoadedImage(NvmExpressDxe) 
```
Results for `loadPciRom` of all the ROMs I tested:
```
loadPciRom: Image W5700.rom loaded - Success
loadPciRom: Image Ge8800GT.rom loaded - Success
loadPciRom: Image gtx680.rom loaded - Success
LoadPciRom: Start image #1 error - Incompatible Version
loadPciRom: Image titanx.rom loaded - Not Found
loadPciRom: Image amd_gop_1.57.0.0.0.efirom loaded - Success
loadPciRom: Image amd_gop_mcu.efirom loaded - Success
loadPciRom: Image amd_gop_mod.efirom loaded - Success
loadPciRom: Image amd_gop_vega.efirom loaded - Success
loadPciRom: Image amd_gop.efirom loaded - Success
LoadPciRom: Start image #0 error - Incompatible Version
loadPciRom: Image nv_gop_GF10x_MXM.efirom loaded - Not Found
LoadPciRom: Start image #0 error - Incompatible Version
loadPciRom: Image nv_gop_GF10x.efirom loaded - Not Found
LoadPciRom: Start image #0 error - Incompatible Version
loadPciRom: Image nv_gop_GF119.efirom loaded - Not Found
LoadPciRom: Start image #0 error - Incompatible Version
loadPciRom: Image nv_gop_GK1xx_multi.efirom loaded - Not Found
LoadPciRom: Start image #0 error - Incompatible Version
loadPciRom: Image nv_gop_GK1xx_MXM.efirom loaded - Not Found
LoadPciRom: Start image #0 error - Incompatible Version
loadPciRom: Image nv_gop_GK1xx.efirom loaded - Not Found
LoadPciRom: Start image #0 error - Incompatible Version
loadPciRom: Image nv_gop_GM1xx_MXM.efirom loaded - Not Found
LoadPciRom: Start image #0 error - Incompatible Version
loadPciRom: Image nv_gop_GM1xx.efirom loaded - Not Found
LoadPciRom: Start image #0 error - Incompatible Version
loadPciRom: Image nv_gop_GM2xx.efirom loaded - Not Found
LoadPciRom: Start image #0 error - Incompatible Version
loadPciRom: Image nv_gop_GP1xx.efirom loaded - Not Found
LoadPciRom: Start image #0 error - Incompatible Version
loadPciRom: Image nv_gop_GT21x.efirom loaded - Not Found
loadPciRom: Image nv_gop_GV1xx.efirom loaded - Success
loadPciRom: Image nv_gop_TU1xx.efirom loaded - Success
```
So it appears that all the Nvidia ROMs specify 2.1 and most of them check the table revision.
AMD ROMs specify 2.3 so maybe that's why that version was chosen even though AMD GPUs don't check the table revision.

mikebeaton: Tyvm - this is getting quite specific.

I make it:

 - The previous forge to 2.1 which @dakanji found in the wild sounds like it _is_ enough for all known GPUs (as we believe it is intended to be)
 - OCs previous (and still as suggested in this PR) forge to 2.3 is possibly an abundance of caution, but arguably also a better choice (as that much more future proof, and seen in the wild, even if not seen checked, in the wild)

Tbh, they way it looks to me now, the existing code OcForgeUefiSupport was fine - we _could_ decrement the check from the previous 2.30 to 2.10, to avoid overwriting the same work already done by another injected firmware (even though doing so ought to be completely harmless). Even then, since OC is configurable, and someone should know whether or not they have that other firmware, I still can't really see any good reason to change the OC existing code.

Many thanks for that @joevt.

mikebeaton: On the gRT->QueryVariableInfo which I am modifying - this is *not* intended as a spoof addition to a real firmware which already has GetVariable and SetVariable. Rather, in this case, we are putting in a completely new implementation of GetVariable and SetVariable, a brand new, memory-only, NVRAM driver, and QueryVariableInfo, _if it previously existed in the firmware_, must be replaced to match.

Again, it's not to be honest clear to me that we need to change the existing code from what was there before #405 . The changes here are only to deal with an imaginary (never seen) very clumsy prior spoofing of gRT, but why would anyone do that - why would anyone move the version up to 2+ and not provide (space for) the members which this implies. I guess you could try to argue some obscure case, but I've never seen it, and I'm not sure we should provide for it. I'm actually minded to revert #403, which imo misunderstood that purpose of the original code. cc @vit9696

I do have a final question about gRT->QueryVariableInfo , @joevt, because it is mentioned in your quotes above. Are you aware of any GPU (or other card we want to enable) which requires it? Even if so, this is a completely separate matter from the current gRT changes being discussed in this PR and in #405, as this code was never designed as a way to add a forged QueryVariableInfo onto an existing NVRAM implementation.

joevt: I have not checked if the GPU GOP drivers call QueryVariableInfo. I believe Windows UEFI does and I believe people do use ForgeEFI to run Windows UEFI on older Macs. Anyway, while faking UEFI 2.x, you should probably make these UEFI 2.x functions do something other than cause a crash. I suppose it might be interesting to add some logging. I did add some stack trace stuff to my fork of RefindPlus so it could maybe dump what code is calling the function.

dakanji: Wow, this has generated so much column inches!

It is a trivial matter at the end of the day (IMO). If it is felt strongly that things should be in one way, then this should just be done and energies directed elsewhere.

What is now more interesting to me, for educational purposes, is what exactly the link between `SetVariable` and `GetVariable` is, that mandates `QueryVariableInfo` **_MUST_** be changed on account of changes to those two as mentioned a few times.

Those two are typically the most changed/wrapped functions and such does not typically involve QVI.

Wery much curious as to why it is a must here for my edification as mentioned. This is more so given that UEFI-Mac does not even seem to use QVI to start with ... or at least returns clearly useless output if called.

dakanji: Based on my understanding of what has gone before, seems best to me that the previous change be reverted or the current proposal be adopted.

mikebeaton: Hi @joevt - You're credited in the OpenCore UEFI forging (of course!), but I did already note that the implementation of your work into OpenCore does a lot less than your original code. Basically, we *only* forge CreateEventEx and the header size in gBS, and then the UEFI version numbers in gST and gBS. As you know, gST, gBS and gRT each have their own separate UEFI version number, and these can be different - as to which version number some option rom or other driver will check (if any!), before attempting to use a given function, I guess there is no way to know for certain in advance. We don't fake anything in gRT at all, yet.

Even if someone used the current emulated NVRAM driver, it would not add QueryVariableInfo to an existing gRT which didn't it already have it (much less the other two new-in-UEFI 2 gRT functions), because it's not trying to. It's trying to replace an existing or not implemented (does nothing and returns not implemented) hardware implementation, with a replacement, RAM-based, implementation which is correct for the firmware version in which it finds itself. (Currently, and I think we're agreed on keeping this, it checks only the gRT version.)

When you say 'using ForgeEFI' in the above, you're talking about your own full version, not the OC flag, right. However, in terms of the OC implementation, I don't think forging QueryVariableInfo has been asked for, or even noted as missing (i.e. causing problems by being missing) in OpenCore.

dakanji: The coverage of the implementation of OpenCore's `ForgeUEFI` is the same as the original `FakeUEFI2` by @joevt . That is, add CreateEventEx and change the mentioned headers. QVI is not changed in either as was not needed for the aim, which was to load UEFI 2.x drivers that need CreateEventEx, such as XhciDxe and more narrowly and specifically, to run `ReloadOptionRoms`.

The main thing to note is that at the time, a query was put to @vit9696 as to whether the changes should be reverted after the aim was achieved but it was decided to leave it alone as it obviously added complexity and in any case, nothing was likely to need to check the EFI Version afterwards.  Implicit in that was an assumption that nothing was likely to have changed things beforehand.

Given that this is not quite always the case, my main aim here, regardless of actual implementation details and conclusions reached by gurus, is that any development involving such checks/changes, should now start to consider that checks/changes can happen both before and after the code segment is being run, and that this may be in wierd ways ... such as changing the contents but keeping the header unchanged.

joevt: > When you say 'using ForgeEFI' in the above, you're talking about your own full version, not the OC flag, right?
I don't think anyone is using my work directly. They use OpenCore or OCLP or RefindPlus - whatever has the functionality they require.

> any development involving such checks/changes, should now start to consider that checks/changes can happen both before and after the code segment is being run, and that this may be in wierd ways ... such as changing the contents but keeping the header unchanged.
I agree with this. People may use RefindPlus to launch OpenCore. They use RefindPlus to launch legacy Windows, or anything they want to run without OpenCore modifications. Or people may use the Driver#### mechanism to load stuff before the Apple Startup Manager or any boot loader begins. Any of those might make changes so sanity checks should exist.

mikebeaton: > Given that this is not quite always the case, my main aim here, regardless of actual implementation details and conclusions reached by gurus, is that any development involving such checks/changes, should now start to consider that checks/changes can happen both before and after the code segment is being run, and that this may be in wierd ways ... such as changing the contents but keeping the header unchanged.

Totally agreed, and thank your for bringing this to our attention!

> The coverage of the implementation of OpenCore's ForgeUEFI is the same as the original FakeUEFI2 by @joevt .

I'm talking about [this](https://github.com/joevt/joevtApps/blob/main/FakeUEFI2/FakeUEFI2.c) - which in fact covers considerably more UEFI 2 features. OC's ForgeUefiSupport was done (not long) before I started working on OC, but I've been assuming that it is a considerably cut-down version from that, after discussion as to which bits were actually needed for non-native GPU support.

joevt: > I'm talking about [this](https://github.com/joevt/joevtApps/blob/main/FakeUEFI2/FakeUEFI2.c) - which in fact covers considerably more UEFI 2 features. OC's ForgeUefiSupport was done (not long) before I started working on OC, but I've been assuming that it is a considerably cut-down version from that, after discussion as to which bits were actually needed for non-native GPU support.

Right. A lot of that is debug stuff and commented out. It was interesting enough to keep but isn't needed. There's also a lot of driver fluff that isn't necessary for this kind of driver because it's not a driver - it's just a patch. All the protocols don't need to exist and don't need to get installed by `EfiLibInstallAllDriverProtocols2`. It's just there to exist in the device handles list when you use the `dh` command in the UEFI Shell. This means `FakeUEFI2Unload` doesn't need to exist.

- I thought the pool allocation checks (`PoolType` and `LogPoolProc` were interesting enough to keep but they don't get called. They were originally put there to examine how the tables are allocated.
- It compares table sizes with compile time struct sizes.
- It creates a new Boot Services table if it's less than the expected size - the last function is CreateEventEx which it will add to the new Boot Service table.
- The code that does the same for Runtime Services table and the `QueryVariableInfo` and `MyQueryCapsuleCapabilities` functions is commented out.
- There's even code for Driver Services table and the `MySetMemorySpaceCapabilities` function but I don't know any reason to fix that so it's commented out.

Most of the functions to be added to modified tables are no-ops anyway. The only one that does anything is `MyCreateEventEx`.

dakanji: > I'm talking about this - which in fact covers considerably more UEFI 2 features

I kind of summarised there. The coverage is **_effectively_** the same, if not exactly so.

mikebeaton:
See comments at https://github.com/acidanthera/OpenCorePkg/pull/405#discussion_r1083452404 and https://github.com/acidanthera/OpenCorePkg/pull/414/files#r1083450225 .

PR #416, opened 2023-01-09
User: Corrects ASSERT behavior in UserPcd

savvamitrofanov:
The CpuBreakPoint in UserMisc normally shouldn't enter dead-loop condition, instead it should break the program using abort signal. Also sets DEBUG_PROPERTY_ASSERT_BREAKPOINT_ENABLED bit by default in PcdDebugPropertyMask. As a background was a problem with fuzzing when program process falls into dead-loop condition instead of just to exit with crash

### DIFF
--- a/User/Library/UserMisc.c
+++ b/User/Library/UserMisc.c
@@ -14,10 +14,7 @@ CpuBreakpoint (
   VOID
   )
 {
-  ASSERT (FALSE);
-
-  while (TRUE) {
-  }
+  abort ();
 }
 
 VOID
--- a/User/Library/UserPcd.c
+++ b/User/Library/UserPcd.c
@@ -4,10 +4,14 @@
 **/
 
 #include <UserPcd.h>
+#include <Library/DebugLib.h>
 
 #define _PCD_VALUE_PcdUefiLibMaxPrintBufferSize         320U
 #define _PCD_VALUE_PcdUgaConsumeSupport                 ((BOOLEAN)1U)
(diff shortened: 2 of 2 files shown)


User/Library/UserMisc.c
```diff
@@ -14,10 +14,7 @@ CpuBreakpoint (
   VOID
   )
 {
-  ASSERT (FALSE);
-
-  while (TRUE) {
-  }
+  abort ();
```
mikebeaton, line 17: What happens if we abort rather than deadloop in code injected into firmware, for instance? Or if this happened in OC in a machine using OpenDuet?

More generally, worrying that this change maybe affects wanted behaviour in situations like the above. (Guess @vit9696 , @mhaeuser will have better idea.)

savvamitrofanov: I think User library used only in userspace, it must not be injected into firmware, so I personally doesn't see any problem

mikebeaton: I am with you, this change is already in the userspace debug library only. Ofc! Fine, forget my comment then! Ty.

User/Include/UserPcd.h
```diff
@@ -35,6 +35,7 @@
 
 #include <Uefi.h>
 #include <Library/PcdLib.h>
+#include <Library/DebugLib.h>
```
vit9696, line 38: Please do not include this header for all files. OcMp3Lib build failure most likely happens due to this. Move this to UserPcd.c.

savvamitrofanov: Corrected

PR #417, opened 2023-01-11
Debug: Changed ImageBase field to load address for debug info search.

MikhailKrichanov:


mikebeaton:
Thank you for the quick fix - I can't test before this eve., but will do then.

Debug/Scripts/gdb_uefi.py
```diff
@@ -362,7 +363,7 @@ def parse_edii(self, edii, count):
                 entry = entry['NormalImage']
                 self.parse_image(entry['LoadedImageProtocolInstance'], syms)
             else:
-                print(f"Skipping unknown EFI_DEBUG_IMAGE_INFO(Type 0x{entry['ImageInfoType'].dereference():x})")
+                print("Skipping unknown EFI_DEBUG_IMAGE_INFO(Type " + str(entry['ImageInfoType'].dereference()) + ")")
```
dreamwhite, line 366: I think the print statement can be replaced by:

```python
print(f"Skipping unknown EFI_DEBUG_IMAGE_INFO(Type {str(entry['ImageInfoType'].dereference())})"
```

Debug/Scripts/lldb_uefi.py
```diff
@@ -373,7 +374,7 @@ def parse_edii(self, edii, count):
                 entry = entry.GetChildMemberWithName('NormalImage')
                 self.parse_image(entry.GetChildMemberWithName('LoadedImageProtocolInstance'), syms)
             else:
-                print(f'Skipping unknown EFI_DEBUG_IMAGE_INFO (Type 0x{image_type:x})')
+                print('Skipping unknown EFI_DEBUG_IMAGE_INFO (Type ' + str(image_type) + ')')
```
dreamwhite, line 377: Same here:

```python
print(f'Skipping unknown EFI_DEBUG_IMAGE_INFO (Type {str(image_type)})')
```

dreamwhite (changes requested):
Some small hints for Python3-like code

mikebeaton:
Correct me if I'm wrong, but even with the above fixes, I don't think I can now step out of the OpenCore image and into the OVMF image (e.g. if both were built using NOOPT), as I could before the new build system.

I'm not sure if everything built with normal EDK-II build commands (e.g. using the EDK-II provided `build.sh` script in the `OvmfPkg` directory) in theory should also use new build commands and new loaders, and so should load debug info correctly with these `efidebug.toold` changes, or not?

PR #418, opened 2023-01-17
Add ResizeUsePciRbIo to workaround broken PciIo on some UEFI firmwares

xCuri0:
I created this patch because I would get ```Capability I/O Error``` when using **ResizeGpuBars** or **ResizeAppleGpuBars**. It turns out my system had a buggy **PciIo** implementation that would not allow access to the PCIe extended configuration space (above 256 bytes). So I created this patch to add an option to use **PciRootBridgeIo** instead which works without any issues on my system. OpenCore has no issues resizing my RX 580's BAR from 8GB to 256MB and booting macOS after this patch.

Before
```
03:002 00:065 OC: Increasing GPU BARs to 8
03:034 00:031 OCDM: Setting RBAR to 8 inc 1 on 3/19
03:100 00:065 OCDM: Capability I/O error - Unsupported
03:132 00:031 OCDM: RBAR is unsupported by device - Device Error
03:197 00:065 OCDM: Setting RBAR to 8 inc 1 on 5/19
03:229 00:031 OCDM: Capability I/O error - Unsupported
03:261 00:031 OCDM: RBAR is unsupported by device - Device Error
```
After (using patch with **ResizeUsePciRbIo** enabled)
```
03:000 00:065 OC: Setting GPU BARs to 8
03:032 00:031 OCDM: RBAR using PciRootBridgeIo
03:064 00:032 OCDM: Setting RBAR to 8 inc 1 on 0/2/0
03:130 00:066 OCDM: Read from disabled device
03:162 00:031 OCDM: RBAR is unsupported by device - Invalid Parameter
03:228 00:066 OCDM: Setting RBAR to 8 inc 1 on 1/0/0
03:260 00:031 OCDM: RBAR control is D20, total 1 - Success
03:327 00:066 OCDM: Old BAR 0000000C 0000000C 0000000C 0000000E 0000E001 F7E00000 - Success
03:358 00:031 OCDM: RBAR 1/1 supports 0x3F00, sizing 8 inc 1 results setting from 13 to 8
03:390 00:032 OCDM: New BAR 0000000C 00000000 0000000C 0000000E 0000E001 F7E00000 - Success
03:456 00:065 OCDM: Reprogrammed BARs to original - Success
```

Newer UEFI firmwares supporting Resizable BAR officially likely don't need this workaround.  My motherboard is a Gigabyte B75M-D3H running BIOS F16 (December 2018) with [ReBarUEFI](https://github.com/xCuri0/ReBarUEFI/) module/patches I developed.

There is one issue which is Windows giving **Code 43** on the GPU when using **ResizeGpuBars** however Linux works without any issues even when using ```pci=nocrs``` to stop the AMDGPU driver from resizing BAR to 8GB. **ResizeUsePciRbIo** should be tested on other systems where **PciIo**  **ResizeGpuBars** works in Windows to see if using **PciRootBridgeIo** is causing that (IMO it's not, it's probably PCI Bridge or something related?).


Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -309,8 +377,241 @@ SetResizableBarOnDevice (
...
+                           EfiPciWidthUint32,
+                           PciAddrOffset (PciAddress, Offset),
+                           1,
+                           &Entries[Index].ResizableBarControl.Uint32
+                           );
+
+    ChangedBars = TRUE;
+  }
+
+  if (ChangedBars) {
```
mhaeuser, line 563: This is only ever set to TRUE above, so why have a var and a check?

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -309,8 +377,241 @@ SetResizableBarOnDevice (
...
+  }
+
+  ResizableBarNumber = MIN (ResizableBarControl.Bits.ResizableBarNumber, PCI_MAX_BAR);
+
+  Status = PciRootBridgeIo->Pci.Read (
+                                  PciRootBridgeIo,
+                                  EfiPciWidthUint8,
+                                  PciAddrOffset (PciAddress, ResizableBarOffset + sizeof (PCI_EXPRESS_EXTENDED_CAPABILITIES_HEADER)),
+                                  sizeof (PCI_EXPRESS_EXTENDED_CAPABILITIES_RESIZABLE_BAR_ENTRY) * ResizableBarNumber,
+                                  (VOID *)Entries
```
mhaeuser, line 446: Which compiler requires these kinds of cast?

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -388,3 +689,178 @@ ResizeGpuBars (
...
+ResizeGpuBars (
+  IN PCI_BAR_SIZE  Size,
+  IN BOOLEAN       Increase,
+  IN BOOLEAN       UseRbIo
+  )
+{
+  if (UseRbIo) {
+    DEBUG ((DEBUG_INFO, "OCDM: RBAR using PciRootBridgeIo\n"));
+    return ResizeGpuBarsRbIo (Size, Increase);
+  } else {
```
mhaeuser, line 862: Superfluous else (I think some compiler configurations may actually issue a warning)

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -388,3 +689,178 @@ ResizeGpuBars (
...
+    }
+
+    (*Descriptors)++;
+  }
+
+  if ((*Descriptors)->Desc == ACPI_END_TAG_DESCRIPTOR) {
+    *IsEnd = TRUE;
+  }
+
+  return EFI_SUCCESS;
```
mhaeuser, line 735: Why return a status when the function can only ever succeed? Maybe return IsEnd?

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -388,3 +689,178 @@ ResizeGpuBars (
...
+  )
+{
+  EFI_STATUS                         Status;
+  UINTN                              HandleCount;
+  EFI_HANDLE                         *HandleBuffer;
+  UINTN                              Index;
+  EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL    *PciRootBridgeIo;
+  EFI_ACPI_ADDRESS_SPACE_DESCRIPTOR  *Descriptors;
+  UINT8                              HdrType;
+  UINT8                              Bus, Dev, Func;
```
mhaeuser, line 751: Multiple variable declarations in a single line are prohibited

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -388,3 +689,178 @@ ResizeGpuBars (
...
+            PciRootBridgeIo->Pci.Read (PciRootBridgeIo, EfiPciWidthUint8, PciAddress + PCI_HEADER_TYPE_OFFSET, 1, &HdrType);
+            if (!Func && ((HdrType & HEADER_TYPE_MULTI_FUNCTION) == 0)) {
+              break;
+            }
+          }
+        }
+      }
+    }
+  }
+
```
mhaeuser, line 844: HandleBuffer is leaked

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -78,9 +79,76 @@ LocatePciCapability (
   return EFI_NOT_FOUND;
 }
 
+// Needed to access address larger than 256
+UINT64
+PciAddrOffset (
+  UINTN  PciAddress,
```
mhaeuser, line 85: PciAddress is a UINT64, not UINTN (this goes for all changes)

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -78,9 +79,76 @@ LocatePciCapability (
...
+  INTN   Offset
+  )
+{
+  UINTN  Reg  = (PciAddress & 0xffffffff00000000) >> 32;
+  UINTN  Bus  = (PciAddress & 0xff000000) >> 24;
+  UINTN  Dev  = (PciAddress & 0xff0000) >> 16;
+  UINTN  Func = (PciAddress & 0xff00) >> 8;
+
+  return EFI_PCI_ADDRESS (Bus, Dev, Func, ((INT64)Reg + Offset));
```
mhaeuser, line 94: Why cast to a signed int?

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -388,3 +689,178 @@ ResizeGpuBars (
...
+      *MinBus = (UINT16)(*Descriptors)->AddrRangeMin;
+      *MaxBus = (UINT16)(*Descriptors)->AddrRangeMax;
+      (*Descriptors)++;
+      return (EFI_SUCCESS);
+    }
+
+    (*Descriptors)++;
+  }
+
+  if ((*Descriptors)->Desc == ACPI_END_TAG_DESCRIPTOR) {
```
mhaeuser, line 731: This is always true and follows directly from the loop invariant

mhaeuser (changes requested):
Sorry, I only have general remarks, because PCIe is not my strong suit.

Library/OcDeviceMiscLib/SetResizableBar.c
xCuri0: This issue also exists in the PciIo version. I'm unable to test that on my system but it would need to be fixed too

Library/OcDeviceMiscLib/SetResizableBar.c
xCuri0: This also is there in the PciIo version fixing it anyways though

Library/OcDeviceMiscLib/SetResizableBar.c
xCuri0: Also was taken from PciIo version but will try to fix it

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -78,9 +79,76 @@ LocatePciCapability (
   return EFI_NOT_FOUND;
 }
 
+// Needed to access address larger than 256
+UINT64
+PciAddrOffset (
```
vit9696, line 84: Should be STATIC. Also reformat the comments in EDK II style.

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -309,8 +377,241 @@ SetResizableBarOnDevice (
...
+       && (Entries[Index].ResizableBarControl.Bits.BarSize > (UINT32)Size))
+    {
+      Bit = LowBitSet64 (Capabilities);
+    } else {
+      Bit = HighBitSet64 (NewCapabilities);
+    }
+
+    DEBUG ((
+      DEBUG_INFO,
+      "OCDM: RBAR %u/%u supports 0x%Lx, sizing %u inc %d results setting from %u to %d\n",
```
vit9696, line 530: Maybe change the comment to be able to distinguish PciIo version from RbIo version from the log?

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -388,3 +689,178 @@ ResizeGpuBars (
 
   return EFI_NOT_FOUND;
 }
+
+EFI_STATUS
+PciGetNextBusRange (
```
vit9696, line 694: Should be STATIC too.

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -388,3 +689,178 @@ ResizeGpuBars (
...
+
+  if ((*Descriptors)->Desc == ACPI_END_TAG_DESCRIPTOR) {
+    *IsEnd = TRUE;
+  }
+
+  return EFI_SUCCESS;
+}
+
+EFI_STATUS
+ResizeGpuBarsRbIo (
```
vit9696, line 739: And this should also be STATIC.

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -388,3 +689,178 @@ ResizeGpuBars (
...
+    if (EFI_ERROR (Status)) {
+      continue;
+    }
+
+    PciRootBridgeIo->Configuration (PciRootBridgeIo, (VOID **)&Descriptors);
+    if (EFI_ERROR (Status)) {
+      continue;
+    }
+
+    // Not sure if multiple root bridge systems even exist but this should support them
```
vit9696, line 787: They do, yeah, e.g. dual socket server stuff. Please change the comments to match the EDK II style though.

vit9696 (commented):
Looks mostly good to me. Added a few comments to @mhaeuser's.

Library/OcDeviceMiscLib/SetResizableBar.c
xCuri0: Already it gets printed that PciRootBridgeIo or PciIo is being used so I thought it wouldn't needed to add it here

Library/OcDeviceMiscLib/SetResizableBar.c
xCuri0: Should ResizeGpuBarsPciIo also be changed to STATIC then ?

xCuri0:
Applied the requested changes and confirmed it works (atleast for my hardware configuration). Do I need to squash the commits before they can be merged ?

Library/OcDeviceMiscLib/SetResizableBar.c
vit9696: Right, never mind.

vit9696:
We can squash the commits at merge time, thankfully GH has a button for this. Looks good on my side, @mhaeuser could you merge?

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -309,8 +381,234 @@ SetResizableBarOnDevice (
...
+  }
+
+  DEBUG_CODE_BEGIN ();
+
+  Status = PciRootBridgeIo->Pci.Read (
+                                  PciRootBridgeIo,
+                                  EfiPciWidthUint32,
+                                  PciAddrOffset (PciAddress, OFFSET_OF (PCI_TYPE00, Device.Bar)),
+                                  PCI_MAX_BAR,
+                                  (VOID *)NewBar
```
mhaeuser, line 569: Was there any trouble dropping the casts (if so, which warning)?

xCuri0: afaik I didn't get any warnings but can't be sure since I'm building it inside Docker and can't see the full log

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -309,8 +381,234 @@ SetResizableBarOnDevice (
...
+    ResizableBarControl.Uint32,
+    MIN (ResizableBarControl.Bits.ResizableBarNumber, PCI_MAX_BAR),
+    Status
+    ));
+
+  if (EFI_ERROR (Status)) {
+    return EFI_UNSUPPORTED;
+  }
+
+  ResizableBarNumber = MIN (ResizableBarControl.Bits.ResizableBarNumber, PCI_MAX_BAR);
```
mikebeaton, line 440: @mhaeuser Might it be worth adding a static assert here that `PCI_MAX_BAR > 0`, since the previous `ChangedBars` logic would have come into effect in that (implausible? broken?) case?

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -102,7 +174,7 @@ SetResizableBarOnDevice (
 
   ChangedBars = FALSE;
```
mikebeaton, line 175: We also have the `ChangedBars` logic here, in the original code.

Library/OcDeviceMiscLib/SetResizableBar.c
mikebeaton: Yes - imo

Library/OcDeviceMiscLib/SetResizableBar.c
mhaeuser: It’s not configurable, is it? I only add STATIC_ASSERTs for non-configurable macros when the property is not obvious to hold (say inequality with another unobvious number, whether it’s a power of two, or divisible by some value). This looks fairly obvious to me, but I might be missing something.

Library/OcDeviceMiscLib/SetResizableBar.c
mhaeuser: Sorry, what does this refer to?

mikebeaton: You asked @xCuri0 to remove this logic in his (modified) clone of this method. I think it should be removed here too?

Library/OcDeviceMiscLib/SetResizableBar.c
```diff
@@ -309,8 +381,234 @@ SetResizableBarOnDevice (
...
+    DEBUG ((DEBUG_INFO, "OCDM: RBAR caps cannot be read - %r\n", Status));
+    return EFI_UNSUPPORTED;
+  }
+
+  Status = PciRootBridgeIo->Pci.Read (
+                                  PciRootBridgeIo,
+                                  EfiPciWidthUint32,
+                                  PciAddrOffset (PciAddress, OFFSET_OF (PCI_TYPE00, Device.Bar)),
+                                  PCI_MAX_BAR,
+                                  (VOID *)OldBar
```
mikebeaton, line 459: Sorry - another one - for consistency if we're removing `(VOID *)` in one place where it's not needed, I guess we should remove it in all calls to `Pci.Read` or `Pci.Write` in this file where we are passing a pointer to a `VOID *` parameter.

mikebeaton: (I'm picking up on this because @xCuri0 said that he got the code which he was asked to change from the original code!)

Library/OcDeviceMiscLib/SetResizableBar.c
mikebeaton: (I'm picking up on this because @xCuri0 said that he got the code which he was asked to change from the original code!)

Library/OcDeviceMiscLib/SetResizableBar.c
mikebeaton: I thought it could sometimes be, as much as anything, a way to document an assumption made, in order for the code which follows to be correct.

But happy to accept that this is such an obvious or unavoidable 'assumption' of this coffee, that it is not worth documenting.

PS I did wonder whether the check that was removed might perhaps have been added in the first place to placate some previous static coffee analyser?

Library/OcDeviceMiscLib/SetResizableBar.c
mhaeuser: You can if you want to, but that's really a separate thing. About all the "it's in the original code" comments - just because things slip review once, doesn't mean they should slip twice. With regards to casts, my personal policy to have them if and only if there is a noteworthy effect. E.g., an integer value may be shortened, signed-to-unsigned conversion, etc. All pointers cast implicitly to VOID *, there's no concerns with alignment or strict aliasing, it's as boring as it gets.

mikebeaton: I agree if things slip once they shouldn't slip twice, maybe my suggestion was misunderstood, my suggestion is to fix them in the other places where they also apply, at least in this file, now that they have been spotted.

Library/OcDeviceMiscLib/SetResizableBar.c
mikebeaton: The `ChangedBars` check which was removed because it's not needed because the thing that we are not static asserting is never going to be false!

Library/OcDeviceMiscLib/SetResizableBar.c
xCuri0: @mikebeaton I could fix the PciIo code but I'm unable to test it due to my system having a broken PciIo implementation. For that reason I will probably just leave PciIo as is

Library/OcDeviceMiscLib/SetResizableBar.c
mikebeaton: It took me a couple of reads, but I believe I understand you distinction for configurable vs. non-configurable constants, and when you would choose to static assert on the latter, now, tyvm.

Library/OcDeviceMiscLib/SetResizableBar.c
mikebeaton: Despite the other discussion below, I would still feel the same about this: if it was worth changing in the modified clone of this method, then surely it's worth changing here, in the original method in the same file, no?

Library/OcDeviceMiscLib/SetResizableBar.c
mhaeuser: > The `ChangedBars` check which was removed because it's not needed because the thing that we are not static asserting is never going to be false!

Wh-What? :D The only check I suggested to "remove" was when a BOOLEAN was set to TRUE only on a single path. Instead of having a BOOLEAN, you might as well just move the conditional code there. I don't remember where or what it was. As I said, PCI is not my strength and I did not try to understand the code in any shape or form.

mhaeuser: > It took me a couple of reads, but I believe I understand you distinction for configurable vs. non-configurable constants, and when you would choose to static assert on the latter, now, tyvm.

Yes, just for the record, configurable constants (say PCDs or other constants that are not set in stone) are special, as they may be re-chosen with the risk of human error. So I'd ASSERT a lot more for those.

Library/OcDeviceMiscLib/SetResizableBar.c
mhaeuser: As I said for another change, feel free to, but I wouldn't require this for merging.

mikebeaton: PS If you were agreeing that it's theoretically worth doing that more than it's fair to ask @xCuri0 to pick up, for his PR to be accepted, then apologies for misunderstanding understanding that point - and - although I would still think they really are minor, and related, changes - of course be happy to do a small subsequent PR to make the additional changes (no new static asserts!) - that I am suggesting.

Library/OcDeviceMiscLib/SetResizableBar.c
mikebeaton: Okay. I'll close all these and often a separate PR applying the same minor fixes already applied in your new bits. Thank you!

xCuri0:
Requested changes all done, lmk if I missed something. I also compiled with ```-Wall -Wextra -Werror``` (GCC) and there were no errors

xCuri0:
Any updates on this ?

vit9696:
We have CI issues at the moment (GH Actions fail to start all over the org), and need to wait till GitHub support fixes it for us. We will merge this soon ^__^

mikebeaton:
Squashed and merged in https://github.com/acidanthera/OpenCorePkg/commit/d928b90aa509986b87c51e934cf3fd214770750f -- thank you

xCuri0:
I found another bug relating to this, `PciIo->Mem.Read` / `PciIo->Mem.Read` doesn't work on devices allocated above 4G. 
On my system AudioDxe doesn't work because of it though it does identify the GPU audio which is allocated below 4G. I don't have a monitor with speakers to test it though.

```
00:210 00:007 HDA: Connecting controller - PciRoot(0x0)/Pci(0x1,0x0)/Pci(0x0,0x1)
00:211 00:001 HDA: Controller version 1.0
00:212 00:001 HDA: Capabilities:
00:213 00:001 HDA:  | 64-bit: Yes  Serial Data Out Signals: 0
00:214 00:001 HDA:  | Bidir streams: 0  Input streams: 0  Output streams: 6
00:215 00:001 HDA: Controller is AMD Ellesmere HD Audio Controller
00:316 00:101 HDA: Controller protocols installed
00:327 00:011 HDA: Controller initialized
00:328 00:001 HDA: Connecting codec 0x0
00:334 00:006 HDA:  | Codec ID: 0x1002:0xAA01
00:340 00:006 HDA:  | Codec name: AMD (Unknown)
00:346 00:006 HDA:  | Codec contains 1 function groups, start @ 0x1, end @ 0x1
00:352 00:006 HDA:  | Function group @ 0x1 is of type 0x1
00:378 00:026 HDA:  | Function group @ 0x1 output amp capabilities: 0x0
00:389 00:011 HDA:  | Function group @ 0x1 GPIO capabilities: 0x0
00:395 00:006 HDA:  | Function group @ 0x1 contains 14 widgets, start @ 0x2, end @ 0xF
00:822 00:426 HDA:  |  Widget @ 0x2 (type 0x0)
00:823 00:001 HDA:  | Port widget @ 0x3 is an output (pin defaults 0x185600F0) (bitmask 1)
00:824 00:001 HDA:  |  Widget @ 0x4 (type 0x0)
00:825 00:001 HDA:  | Port widget @ 0x5 is an output (pin defaults 0x185600F0) (bitmask 2)
00:826 00:001 HDA:  |  Widget @ 0x6 (type 0x0)
00:827 00:001 HDA:  | Port widget @ 0x7 is an output (pin defaults 0x185600F0) (bitmask 4)
00:828 00:001 HDA:  |  Widget @ 0x8 (type 0x0)
00:829 00:001 HDA:  | Port widget @ 0x9 is an output (pin defaults 0x185600F0) (bitmask 8)
00:830 00:001 HDA:  |  Widget @ 0xA (type 0x0)
00:831 00:001 HDA:  | Port widget @ 0xB is an output (pin defaults 0x185600F0) (bitmask 16)
00:832 00:001 HDA:  |  Widget @ 0xC (type 0x0)
00:833 00:001 HDA:  | Port widget @ 0xD is an output (pin defaults 0x185600F0) (bitmask 32)
00:834 00:001 HDA: Codec protocols installed
00:835 00:001 HDA: Codec initialized
00:836 00:001 HDA: Connecting controller - PciRoot(0x0)/Pci(0x1B,0x0)
00:837 00:001 HDA: Controller disable no snoop
**00:838 00:001 HDA: Init PCI HW - Invalid Parameter**
00:889 00:050 OC: Connecting drivers done...
```

I traced it to this line https://github.com/acidanthera/OpenCorePkg/blob/master/Staging/AudioDxe/HdaController/HdaController.c#L300

My ```AMD Ellesmere HD Audio Controller``` is allocated at `0xF7EFC000` (below 4G) and `High Definition Audio Controller` at `0xFFFFFC000` (above 4G).


I could just make it possible for AudioDxe to use PciRootBridgeIo too but it does seem like `ReleaseUsbOwnership` is affected too. 


Is there a better way we can implement a workaround for this ?

xCuri0:
I think we could replace `PciIo->Mem.Read` / `PciIo->Mem.Read` with 2 wrapped functions that use PciRootBridgeIo if requested (say `UseRbIoForPciIo`) ? 

Or could even replace PciIo completely with our own that uses PciRootBridgeIo if requested ?

These are the 2 bugs I'm aware of in these Aptio 4 firmware. There's probably more which I could do additional testing for.

xCuri0:
I've decided to go with replacing the faulty Mem.Read/Write functions with custom ones. Will provide an update

Probably will leave the IO functions as is for now

vit9696:
Yeah, that makes good sense to me. Please provide a PR on that ^_^

xCuri0:
Turns out the issue is actually PciRootBridgeIo here, I rewrote PciIo->Mem.Read to use PciRootBridgeIo and it works exactly the same as before failing on the 64-bit device.

32-bit AMD GPU audio
```
00:652 00:004 OCPIO: Reading from PciRootBridgeIo F7E60003 0 1
00:658 00:005 OCPIO: Reading from PciRootBridgeIo F7E60002 0 1
00:663 00:005 HDA: Controller version 1.0
00:668 00:004 OCPIO: Reading from PciRootBridgeIo F7E60000 1 1
00:674 00:005 HDA: Capabilities:
00:677 00:003 HDA:  | 64-bit: Yes  Serial Data Out Signals: 0
00:683 00:005 HDA:  | Bidir streams: 0  Input streams: 0  Output streams: 6
00:690 00:006 HDA: Controller is AMD Ellesmere HD Audio Controller
00:696 00:006 OCPIO: Reading from PciRootBridgeIo F7E60008 2 1
```

64-bit Motherboard audio

```
02:234 00:004 HDA: Controller disable no snoop
02:239 00:004 OCPIO: Reading from PciRootBridgeIo FFFC10003 0 1
02:245 00:005 HDA: Init PCI HW - Invalid Parameter
02:250 00:005 OCPIO: Reading from PciRootBridgeIo FFFC10008 2 1
```

Will still try and investigate this further. Need to learn more about MMIO because I thought CopyMem using MMIO as source would work but it didn't.

xCuri0:
@vit9696 Finally managed to get it working (haven't implemented write yet)

```
03:171 00:004 HDA: Controller disable no snoop
03:176 00:004 OCPIO: Reading from custom MMIO FFFC10003 0 1
03:182 00:005 OCPIO: CpuMemoryServiceRead
03:186 00:004 OCPIO: Reading from custom MMIO FFFC10002 0 1
03:192 00:005 OCPIO: CpuMemoryServiceRead
03:196 00:004 HDA: Controller version 1.0
03:201 00:004 OCPIO: Reading from custom MMIO FFFC10000 1 1
03:206 00:005 OCPIO: CpuMemoryServiceRead
03:211 00:004 HDA: Capabilities:
03:215 00:003 HDA:  | 64-bit: Yes  Serial Data Out Signals: 0
03:221 00:005 HDA:  | Bidir streams: 0  Input streams: 4  Output streams: 4
03:227 00:006 HDA: Controller is Intel 7 Series HD Audio Controller
03:234 00:006 OCPIO: Reading from custom MMIO FFFC10008 2 1
03:239 00:005 OCPIO: CpuMemoryServiceRead
03:244 00:004 HDA: Controller reset - Invalid Parameter
```

It turns out both PciRootBridgeIo and CpuIo2 have the invalid MMIO limitation. Using IoLib MMIO got it working

So CpuIo2, PciRootBridgeIo and PciIo all need overrides which I'll submit in a PR.

PR #419, opened 2023-01-19
Debug: Fixed lldb sections load addresses.

MikhailKrichanov:


mikebeaton:
This version does not seem to be working for OpenCore symbols any more (not tested OVMF as I would normally start in OpenCore then jump into OVMF as required). I get lots of errors `error: no section found that matches the section name '.text'`.

Screenshot 2023-01-19 at 20 10 46

vit9696:
Abandoned as an incorrect fix.

PR #420, opened 2023-01-23
OcConsoleLib: Prevent failure to properly start console in graphics mode…

mikebeaton:
…when previous console control protocol is non-existent or limited

Having already raised and briefly discussed this issue by PM, this is a PR for it.

I encountered this when developing #414 . The issue is the known problem of APFS log lines appearing too early (pre native picker) in the boot sequence, occurring again with the #414 driver. In this case this is fixable within OC, see https://github.com/mikebeaton/OpenCorePkg/blob/enable-gop/Staging/EnableGop/EnableGop.c#L49-L87 for a description of the problem (EDIT: sample code which was linked in preceding is now removed).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,7 @@ OpenCore Changelog
 - Fixed loading macOS with legacy boot without Apple Secure Boot
 - Added Linux support to legacy boot BootInstall script
 - Updated builtin firmware versions for SMBIOS and the rest
+- Fixed incomplete console mode initialisation when started in graphics mode
 
 #### v0.8.8
 - Updated underlying EDK II package to edk2-stable202211
--- a/Library/OcConsoleLib/ConsoleControl.c
+++ b/Library/OcConsoleLib/ConsoleControl.c
@@ -72,9 +72,9 @@ OcConsoleControlSetMode (
 
 EFI_STATUS
 OcConsoleControlInstallProtocol (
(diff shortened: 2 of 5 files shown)


mikebeaton:
I believe there were two issues in the pre-existing code:

 - At https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcConsoleLib/TextOutputBuiltin.c#L976-L977 the failure to provide `&mConsoleMode` (instead of `NULL`) as the last parameter means that OpenCore's `OcConsoleRendererBuiltinGraphics` mode will always initialise in text mode.
 - In both the lines just linked above (if I apply the just-mentioned change) and in these lines https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcConsoleLib/TextOutputSystem.c#L346-L352 `mConsoleMode` only gets set to the intended value if a) there is a previous console control implementation and b) it allows setting graphics mode.
   - In the case of the driver in #414 I do encounter a previous console control implementation, but it does not allow setting graphics mode: it returns `EFI_UNSUPPORTED` at https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcConsoleLib/ConsoleControl.c#L65-L68 (which is ignored - perhaps rightly, if everything else here is fixed).
   - Note that the functions called by the lines mentioned above do not assume that there is always a previous console control implementation.

As discussed by PM, there are multiple issues around the order of changing console mode, including potential unwanted flickering on some of the many systems supported by OC, and probably other issues, therefore this change would need very careful testing.

PR #421, opened 2023-01-29
EnableGop: Use patched GenSec to produce compressed .ffs which legacy Mac firmware can read

mikebeaton:
Final (EDIT: Next, rather than final - there is now a [further commit](https://github.com/acidanthera/OpenCorePkg/pull/422)) commit in this branch (builds on previous console-control and enable-gop commits) upgrades EDK-II GenSec to support legacy TianoCompress. This allows .fdf files to define compressed PE32 .ffs sections which legacy Mac firmware can read. Otherwise, with unpatched tools, we can defined uncompressed .ffs which legacy Mac can also read, but at about four times the size.


mikebeaton:
Requires https://github.com/acidanthera/audk/pull/17 to build.

PR #424, opened 2023-02-14
OcBootManagementLib: Avoid verbose boot overwriting graphics mode

mikebeaton:
Addresses https://github.com/acidanthera/bugtracker/issues/2225

I believe this is preferable to both the previous and the new behaviour, in that there is no overlap of verbose lines with graphics mode in any circumstance.

I also believe it is preferable to continue to omit the clear screen removed in https://github.com/acidanthera/OpenCorePkg/commit/2abe0760a00377e3e94a7d04ae7abb9428ba8c6f (which is what is at issue here; NB this visible change is not being caused by https://github.com/acidanthera/OpenCorePkg/commit/a189bd53bc2ac1490dac7125a9ce11fd03571755), as in normal boot progress it is unwanted, and a noticeable slow down on older, slower systems.


mikebeaton:
Apologies - give me a minute - I'm rushing - this doesn't appear to be launching correctly, in modes I tested before uploading it.

mikebeaton:
Redone to use existing code to properly check for -v in nvram or OC provided args.

mikebeaton:
Combining this into #426 .

PR #427, opened 2023-02-22
OcLogAggregatorLib: Fix non-standard linkage pattern

mikebeaton:
I believe these changes should be made to `OcLogAggregatorLib`.

Without them, it is impossible to compile and link [this code](https://github.com/mikebeaton/OpenCorePkg/commit/276e1ea3f5cf5b8a20dd0597c3759b3ee49e1f78) in all toolchains.

Here is [the same code](https://github.com/mikebeaton/OpenCorePkg/commit/2fef249388d5cdfba728b6dde07cb97d17a85859) added directly on top of master, without the changes in this PR, showing the errors that arise.

NB I don't want to commit the code in the test commits (the two examples above) to master. It was part of attempting to fully fix https://github.com/acidanthera/OpenCorePkg/pull/426 , but in the end was not needed. But I don't think it should have been impossible to link it.

Specifically (describing the code which is changed in this PR, before these changes):
 - `OcLogAggregatorLib.inf` lists `OcLogAggregatorLib` as an instance of `DebugLib`, but it is not, it is its own library, implementing the methods described in `Include/Acidanthera/Library/OcLogAggregatorLib.h`, whereas instances of `DebugLib` implement `MdePkg/Include/Library/DebugLib.h`
 - `OcDebugLibNull` and `OcDebugLibSerial` _are_ instances of `DebugLib`, but in `OcDebugLibNull.c` and `OcDebugLibSerial.c` we find not just null or serial implementations of `DebugLib`, but also null implementations of `OcLogAggregatorLib`

Additionally:
 - `OcAfterBootCompatLib.inf` is missing an explicit dependency on `OcLogAggregatorLib`, even though it has one:
   - https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcAfterBootCompatLib/ServiceOverrides.c#L538
   - https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcAfterBootCompatLib/ServiceOverrides.c#L29

(I don't believe this second item is the major issue. Just fixing it, without the rest of this PR, results in [different linkage errors](https://github.com/mikebeaton/OpenCorePkg/tree/example-bad2).)

In trying to understand why I couldn't compile and link the code shown in the test commits above, I spotted these apparent issues and made the changes in this PR, which immediately fixed the linkage problems, and I don't think there is another way to fix them.

---

cc @Goldfish64 (?) - It is likely these changes affect use of `OcDebugLibSerial`, which I do not use. If I've understood correctly, this would be used as a drop-in replacement for the main `DebugLib|` definition in `OpenCorePkg.dsc` (instead of the usual `OcDebugLibProtocol`) in some debugging situations. In this case, I believe - though perhaps someone who uses `OcDebugLibSerial` could test, if the motivation for these changes looks good - that the equivalent to just swapping in `OcDebugLibSerial`, before these changes, would be to also swap in `OcLogAggregatorLibNull` as default `OcLogAggregatorLib|` at the same time, after these changes.


mikebeaton:
I've now tried building OpenCore using `OcDebugLibSerial` and the proposed new `OcLogAggregatorLibNull` instead of `OcDebugLibProtocol`. It [works fine](https://github.com/mikebeaton/OpenCorePkg/commit/b0562a8a8a6f640354055d7bb2693bb28880eac2) but showed up additional missing .inf dependencies, which I have [added to this PR](https://github.com/acidanthera/OpenCorePkg/commit/eccb26637e08a57407f7f0a4cf0f38bd9809f187).

I've changed the requested reviewers to @PMheart and @vit9696 (apologies for unnecessary ping @Goldfish64 - see comments below for why!).

I do not fully understand the relationship between `OcLogAggregatorLib` and `OcDebugLibXXX` - in particular, I am not clear whether the null implementation of `OcLogAggregatorLib` should always be used with `OcDebugLibSerial` or not. (But I am pretty sure that the null implementation of `OcLogAggregatorLib` should never have been provided by versions of `OcDebugLib`, once these had been made separate libraries - i.e. the main thing this PR suggests changing.)

---

Perhaps we should also update some of the copyright names and dates in the files updated in [this commit](https://github.com/acidanthera/OpenCorePkg/commit/4c995bb1ba8f73d4eca511589e73c860ac43395e), which split up OcDebugLogLib into OcLogAggregatorLib and OcDebugLibProtocol, since these now range from slightly out of date:

 - https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcDebugLibNull/OcDebugLibNull.c#L4
 - https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcDebugLibNull/OcDebugLibNull.inf#L5
 - https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcLogAggregatorLib/OcAppleLog.c#L2

to seriously out of date:

 - https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcLogAggregatorLib/OcLog.c#L2
 - https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcLogAggregatorLib/OcLogAggregatorLib.inf#L5
 - https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcLogAggregatorLib/OcLogInternal.h#L2

, given that they were updated with this fairly significant change in the meantime, which I found somewhat confusing.

vit9696:
Originally we wanted to extend the DebugLib, but then we abandoned this idea. These functions in the null lib are just leftovers. Thanks for fixing this.

cc @savvamitrofanov, we might need the changes on the amaranth side, and definitely want the OpenCorePkg submodule to be updated in audk.

savvamitrofanov:
@mikebeaton Thanks, Mike!

> Originally we wanted to extend the DebugLib, but then we abandoned this idea. These functions in the null lib are just leftovers. Thanks for fixing this.
> 
> cc @savvamitrofanov, we might need the changes on the amaranth side, and definitely want the OpenCorePkg submodule to be updated in audk.

Sure, I'll do the changes

PR #429, opened 2023-03-01
Utilities: Adds userspace tool for EnhancedFatDxe testing and fuzzing

savvamitrofanov:
- Adds test tool for EnhancedFatDxe driver
- Adds UserUnicodeCollation library
- Corrects memory allocation checks in TestNtfsDxe


Utilities/TestFatDxe/TestFatDxe.c
```diff
@@ -0,0 +1,504 @@
...
+  // Closes root directory
+  //
+  if (RootDirInstance != NULL) {
+    FatClose (RootDirInstance);
+  }
+
+  //
+  // Abandon volume structure
+  //
+  if (Volume != NULL) {
```
vit9696, line 116: This does not look right to me. Why does not Fat itself do it and we need to manually free the internals like DiskIo or BlockIo, which we did not allocate?

savvamitrofanov: Well, we definitely allocate both `BlockIo` and `DiskIo` [here](https://github.com/acidanthera/OpenCorePkg/blob/d55bf707e9bfccaa11713a48824533ce4ef6db86/Utilities/TestFatDxe/TestFatDxe.c#L235-L264) and then call `FatAllocateVolume`, [which prepares the Volume structure.](https://github.com/acidanthera/audk/blob/18722f129ff1f06347daa0504906903ffa785bcc/FatPkg/EnhancedFatDxe/Init.c#L50-L54) This is the same scheme we used in TestNtfsDxe ([1](https://github.com/acidanthera/OpenCorePkg/blob/savvas-fatpkg-fuzz-20230301/Utilities/TestNtfsDxe/TestNtfsDxe.c#L128-L152), [2](https://github.com/acidanthera/OpenCorePkg/blob/savvas-fatpkg-fuzz-20230301/Utilities/TestNtfsDxe/TestNtfsDxe.c#L50-L60)) and TestExt4Dxe ([1](https://github.com/acidanthera/OpenCorePkg/blob/savvas-fatpkg-fuzz-20230301/Utilities/TestExt4Dxe/TestExt4Dxe.c#L279-L306), [2](https://github.com/acidanthera/OpenCorePkg/blob/savvas-fatpkg-fuzz-20230301/Utilities/TestExt4Dxe/TestExt4Dxe.c#L146-L156)). 

The EnhancedFatDxe itself allocates these interfaces inside DriverBindingStart using [`OpenProtocol`](https://github.com/acidanthera/audk/blob/master/FatPkg/EnhancedFatDxe/Fat.c#L384-L422) call, which support we don't implement in User library. And deallocates, at least DiskIo, in DriverBindingStop using [`CloseProtocol`](https://github.com/acidanthera/audk/blob/master/FatPkg/EnhancedFatDxe/Fat.c#L516-L532). Maybe if we implement such logic in the User library, then we can make protocol usage look more native than self-allocate such structures. But it's not so easy.

PR #430, opened 2023-03-01
Build: Provide EDK-II BaseTools in releases

mikebeaton:
This might be useful - as discussed EfiRom is useful for EnableGop vBiosInsert.sh script - it's not particularly easy to find pre-built [EDK-II BaseTools](https://edk2-docs.gitbook.io/edk-ii-basetools-user-guides/tianocompress) (perhaps especially for macOS), and it is easy for us to provide them.

Adds about 1Mb to the release zip size.

Screenshot 2023-03-01 at 23 07 16

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,7 @@ OpenCore Changelog
 - Added version number to EnableGop UI section, so tool builders can track it
 - Added `ProvideCurrentCpuInfo` support for macOS 13.3 DP
 - Added AMD support, GOP offset auto-detection and macOS 10.11+ support to EnableGop vBIOS insertion script
+- Included precompiled EDK-II `EfiRom` and `GenFfs` in `Utilities/BaseTools` with OpenCore releases
 
 #### v0.8.9
 - Improved debug logging when applying ACPI patches
--- a/build_oc.tool
+++ b/build_oc.tool
@@ -275,6 +275,16 @@ package() {
     cp "${selfdir}/Staging/EnableGop/${file}" "${dstdir}/Utilities/EnableGop"/ || exit 1
   done
 
+  # Provide EDK-II BaseTools.
+  mkdir "${dstdir}/Utilities/BaseTools" || exit 1
+  if [ "$(unamer)" = "Windows" ]; then
+    cp "${selfdir}/UDK/BaseTools/Bin/Win32/EfiRom.exe" "${dstdir}/Utilities/BaseTools" || exit 1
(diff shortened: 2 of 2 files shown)


mikebeaton:
Moved to draft as I am just waiting to confirm this is sane in all builds - I believe should be, but want to be sure.

vit9696:
I do not mind. Just in practice, do we really need all these tools? Thought most are very internal.

mikebeaton:
Hmmm fair, some are useful, some less so. The Gen* ones can also be useful for people trying to hack together ROM mods. It might be useful to provide just EfiRom and those ones?

vit9696:
- I think GenFw is legacy and is no longer used, though maybe we just stripped it down, need to recheck.
- ImageTool may be useful, but it is not here for some reason.
- GenFfs is handy for .ffs creation for firmware modification.

Any others I don't know.

mikebeaton:
ImageTool is built somewhere else. But it does not appear to be behaving sanely on my local build (neither the Posix wrapper - script error - not the 32 and 64 bit built binaries - assert).

So skipping that, I could either put just GenFfs and EfiRom in a /Utilities/BaseTools dir, or go back to the basic, simpler idea and just copy EfiRom into /Utilities/EnableGop.

vit9696:
Please keep GenFfs and EfiRom in BaseTools directory. We will add ImageTool there later I think.

mikebeaton:
Have just added a recent copy of `EfiRom` to my MacRumors help/support [post](https://forums.macrumors.com/threads/pre-opencore-gop-support-for-efi-era-imacs-and-mac-pros.2378942/) about the EnableGop driver, which is itself linked to in the `README.md` in `/Utilities/EnableGop` in the OpenCore releases. Leaving this for now.

mikebeaton:
Sorry - I keep missing your posts! Apologies.

Shall I revive this, with the two files mentioned?

vit9696:
I think we can ignore Windows, non-macOS builds are not too important.

vit9696:
Need to land this before Monday to make it into the March release.

mikebeaton:
Okay. I think the current version is okay - it is putting the exact same version of the tools into each Linux build, which I guess means the tools themselves are not built by the same toolchain as OC/Duet.

mikebeaton:
Resolved the Windows issue, so FWIW our build actions are a reasonable place to grab recent `EfiRom.exe` and `GenFfs.exe` now, too.

Pretty sure this is all fine (modulo the proviso above, that the tools in a given release are the ones made during the build, but not necessarily made with the same compilers and options as the build).

PR #431, opened 2023-03-09
Application: Adds ListPartitions

savvamitrofanov:
Adds ListPartitions application used for testing firmware run in QEMU


Application/ListPartitions/ListPartitions.c
```diff
@@ -0,0 +1,91 @@
...
+
+    if (IsZeroGuid (&PartitionEntry->UniquePartitionGUID)) {
+      DEBUG ((DEBUG_WARN, "Partition #%u GPT entry has all-zero GUID\n", (UINT32)(Index + 1)));
+      continue;
+    }
+
+    DEBUG ((DEBUG_WARN, "Partition #%u details:\n", (UINT32)(Index + 1)));
+    DEBUG ((DEBUG_WARN, "  PartitionName      : %-36s\n", PartitionEntry->PartitionName));
+    DEBUG ((DEBUG_WARN, "  PartitionTypeGUID  : %g\n", &PartitionEntry->PartitionTypeGUID));
+    DEBUG ((DEBUG_WARN, "  UniquePartitionGUID: %g\n", &PartitionEntry->UniquePartitionGUID));
```
mikebeaton, line 85: Why all DEBUG_WARN? (Not necessarily stating it is wrong, just wondering if there is a reason.)

mikebeaton: I was meaning to suggest using DEBUG_INFO for the non-warning messages - unless ofc there's a good reason we want this to show up in contexts where WARN will be visible and INFO won't be.

PR #434, opened 2023-03-12
Merge, min size and debug_raise().

MikhailKrichanov:
In this attempt  MdeBuildOptions.dsc.inc does not need to be included in every .dsc file.

### DIFF
--- a/build_duet.tool
+++ b/build_duet.tool
@@ -165,5 +165,5 @@ else
   export SELFPKG
   export NO_ARCHIVES
 
-  src=$(curl -Lfs https://raw.githubusercontent.com/acidanthera/ocbuild/master/efibuild.sh) && eval "$src" || exit 1
+  src=$(curl -Lfs https://raw.githubusercontent.com/acidanthera/ocbuild/merge-min_size-assert/efibuild.sh) && eval "$src" || exit 1
 fi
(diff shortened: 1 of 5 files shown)


build_duet.tool
```diff
@@ -165,5 +165,5 @@ else
   export SELFPKG
   export NO_ARCHIVES
 
-  src=$(curl -Lfs https://raw.githubusercontent.com/acidanthera/ocbuild/master/efibuild.sh) && eval "$src" || exit 1
+  src=$(curl -Lfs https://raw.githubusercontent.com/acidanthera/ocbuild/merge-min_size-assert/efibuild.sh) && eval "$src" || exit 1
```
mikebeaton, line 168: Change from `master` in this line I believe was an error last time it occurred. Is this intended?

build_oc.tool
```diff
@@ -335,7 +335,7 @@ export SELFPKG
 export NO_ARCHIVES
 export DISCARD_PACKAGES
 
-src=$(curl -Lfs https://raw.githubusercontent.com/acidanthera/ocbuild/master/efibuild.sh) && eval "$src" || exit 1
+src=$(curl -Lfs https://raw.githubusercontent.com/acidanthera/ocbuild/merge-min_size-assert/efibuild.sh) && eval "$src" || exit 1
```
mikebeaton, line 338: Ditto

build_duet.tool
MikhailKrichanov: It is only needed for CI, these changes are not supposed to be included in the merge.

MikhailKrichanov:
Merged manually.

PR #436, opened 2023-03-20
OcMemoryLib: Optionally return page table level and control bits from OcGetPhysicalAddress

mikebeaton:
Optionally return page table level and mapping flags from OcGetPhysicalAddress. Useful when examining/debugging the page table entries for a memory location.


mikebeaton:
Closed, used in #437.

PR #441, opened 2023-04-05
code optimization SSDT-PLUG-ALT

Lorys89:
the cpu in the affected chipsets is always under the path _SB_.PRXX so the returns are valid for all the affected chipsets


Lorys89:
updated compatibility with new W790 chipset and Xeon cpu

PR #442, opened 2023-04-06
Docs: Comment DVL0 device in SSDT-SBUS-MCHC

stevezhengshiqi:
According to https://www.insanelymac.com/forum/topic/233292-patching-sbus-in-dsdt/?do=findComment&comment=1561152, which is the earliest discussion of DVL0 device in SSDT-SBUS I can found,

> This "diagsvault" is the diagnostic vault where messages are stored. It's located at address 87 (0x57) on the SMBus controller. More about the specific SMBus controllers in your hack, can be found in the Intel datasheet (use Google).

By looking into several motherboard datasheets and SMBus specifications, I see Intel makes changes starting from 7 series.
Looking into [Intel® 7 Series / C216 Chipset Family Platform Controller Hub (PCH)](https://www.intel.com/content/dam/www/public/us/en/documents/datasheets/7-series-chipset-pch-datasheet.pdf#page807) pg. 807,

> SPD Write Disable — R/WO.
0 = SPD write enabled.
1 = SPD write disabled. Writes to SMBus addresses 50h - 57h are disabled.

which means 0x57 address write ability is disabled. People in [How to patch BIOS to enable SPD Write capability on Intel 8 Series and higher Chipset Family PCH](http://www.softnology.biz/tips_spdwritecap.html) also discovered that `writes to SMBus addresses 50h - 57h are disabled by default via SMBus Host Controller registers`. Based on the long history of the SSDT-SBUS, it might not be compatible with changing SMBus specifications. Also, there's limited documents that show where 0x57 address comes from. Hence, I believe user should use caution when injecting this DVL0 device which might cause unwanted communication to reserved region.

### DIFF
--- a/Docs/AcpiSamples/Source/SSDT-SBUS-MCHC.dsl
+++ b/Docs/AcpiSamples/Source/SSDT-SBUS-MCHC.dsl
@@ -29,6 +29,19 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "MCHCSBUS", 0x00000000)
     {
         Name (_CID, "smbus")  // _CID: Compatible ID
         Name (_ADR, Zero)  // _ADR: Address
+
+        /*
+         * Uncomment replacing 0x57 with your own value which might be found
+         * in SMBus section of Intel datasheet for your motherboard.
+         *
+         * The "diagsvault" is the diagnostic vault where messages are stored.
+         * It's located at address 87 (0x57) on the SMBus controller.
+         * While "diagsvault" may refer to diags, a hardware diagnosis program via EFI for Macs
+         * and communicates with the SMBus controller, the effect is really unknown for hacks.
+         * Uncomment this with caution.
+         */
+
+        /**
         Device (DVL0)
         {
             Name (_ADR, 0x57)  // _ADR: Address
@@ -50,6 +63,8 @@ DefinitionBlock ("", "SSDT", 2, "ACDT", "MCHCSBUS", 0x00000000)
                 })
             }
         }
+        **/
+
         Method (_STA, 0, NotSerialized)  // _STA: Status
         {
             If (_OSI ("Darwin"))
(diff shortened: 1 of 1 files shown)


vit9696:
Makes sense to me. Does `AppleSMBusControllerICH` still attach in this case?

1Revenger1:
This also makes sense to me. The address range 0x50-0x57 is normally used for SPD data from RAM, so this DVL0 device sounds specific to Apple (more specifically, the MacPro5,1/4,1 which are the only devices I could find with this).

I looked on my X1 Extreme and there isn't a device at 0x57, and I would assume this is the case for most devices. I used `i2cdetect` in Linux to list addresses with devices.

stevezhengshiqi:
@vit9696 I can confirm that `AppleSMBusControllerICH` is no longer attached to my laptop. However, I see this as expected behavior as real Macs (at least MacBookPros) do not attach that also. Below is a `MacBookPro16,2`'s PCI device tree.
Screenshot 2023-04-08 at 16 36 40

khronokernel:
`AppleSMBusControllerICH` doesn't load on (newer) mobile units, on desktop Macs such as iMac19,1 (and confirmed 20,1 as well) we do see `AppleSMBusControllerICH` is loaded:

Screenshot 2023-04-09 at 7 29 20 PM

Will note that MacBooks and Mac minis haven't had `AppleSMBusControllerICH` attach since MacBookAir5,x and MacBookPro10,x (Ivy Bridge). Perhaps something related to mobile chipsets with newer generations?

1Revenger1:
It could be that the controller service may not attach if there are no SMBus devices to service/communicate with. The SMBus controller still exists and behaves similarly from Core 2 Duo systems to the newest systems. This same controller is what SMBus Synaptics/Elan trackpads attach to.

I do know that the older MacBooks had an SMBus device used for microphone buttons, which was removed after the 10,x devices. Probably related to this rather than chipset differences.

kaoskinkae64:
I am not an expert on the subject of hacking but I have been learning and understanding it for many years. As I am reading the Device (DVL0) present in the generic before beta 0.9.3 It seems that this is only necessary for one type of equipment like I have read MacPro5.1 or 4.1 and the new or almost new smbios19.1 or 20.1 No...

iamyounix:
As i know, new sbus address is logic, `1` and `0`  which mean `yes` and `no`. `0x57` is for old device.

iamyounix:
> @vit9696 I can confirm that `AppleSMBusControllerICH` is no longer attached to my laptop. However, I see this as expected behavior as real Macs (at least MacBookPros) do not attach that also. Below is a `MacBookPro16,2`'s PCI device tree. Screenshot 2023-04-08 at 16 36 40

Read here: https://github.com/iamyounix/msimagb460_tomahawk#sbus-check

Using iMac20,1 SMBIOS without `sbus` is ok actually. Not need. Not display in ioreg doesn't mean `AppleSMBusController` and `AppleSMBUSPCI` is not loaded. Dont rely on IOreg too much.

PR #443, opened 2023-04-06
OcConsoleLib: Fix BuiltinGraphics init error present since b07843fe1d47454747ae4eda9ea0189aa9fb8c03

mikebeaton:
I have only just spotted that BuiltinGraphics was not entering graphics mode properly after b07843fe1d47454747ae4eda9ea0189aa9fb8c03. This addresses the reason why.

### DIFF
--- a/Library/OcConsoleLib/TextOutputBuiltin.c
+++ b/Library/OcConsoleLib/TextOutputBuiltin.c
@@ -357,6 +357,9 @@ RenderScroll (
                      );
 }
 
+//
+// Resync text renderer. May fail, e.g. if resolution is too small to use.
+//
 STATIC
 EFI_STATUS
 RenderResync (
@@ -997,8 +1000,13 @@ OcUseBuiltinTextOutput (
   Status       = AsciiTextResetEx (&mAsciiTextOutputProtocol, TRUE, TRUE);
 
   if (!EFI_ERROR (Status)) {
-    OcConsoleControlInstallProtocol (&mConsoleControlProtocol, NULL, NULL);
+    //
+    // We are intentionally setting the mode using the pre-existing protocol,
+    // if present, then not using the pre-existing protocol again, even if this
+    // means that subsequent mode changes can't 'really' change the mode.
+    //
     OcConsoleControlSetMode (Mode);
+    OcConsoleControlInstallProtocol (&mConsoleControlProtocol, NULL, NULL);
 
     gST->ConOut    = &mAsciiTextOutputProtocol;
     gST->Hdr.CRC32 = 0;
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,7 @@ OpenCore Changelog
 - Fixed possible hang with `GopBurstMode` enabled on DEBUG builds
 - Enabled `GopBurstMode` even with natively supported cards, in EnableGop firmware driver
 - Fixed inability to patch force-injected kexts
+- Fixed `BuiltinGraphics` failure to display graphics in DEBUG builds on some systems since 0.8.9
 
 #### v0.9.1
 - Fixed long comment printing for ACPI patches, thx @corpnewt
--- a/Library/OcConsoleLib/TextOutputSystem.c
+++ b/Library/OcConsoleLib/TextOutputSystem.c
@@ -343,13 +343,22 @@ OcUseSystemTextOutput (
     ReplaceTabWithSpace
     ));
 
+  //
+  // For all except generic, set requested mode using pre-existing protocol, if
+  // present, before installing replacement.
+  // Note: We need the installed renderer to report the correct (intended) mode,
+  // even if there was no original protocol present.
+  //
   if (Renderer == OcConsoleRendererSystemGraphics) {
+    mConsoleMode = EfiConsoleControlScreenGraphics;
+    OcConsoleControlSetMode (mConsoleMode);
     OcConsoleControlInstallProtocol (&mConsoleControlProtocol, NULL, NULL);
-    OcConsoleControlSetMode (EfiConsoleControlScreenGraphics);
   } else if (Renderer == OcConsoleRendererSystemText) {
+    mConsoleMode = EfiConsoleControlScreenText;
+    OcConsoleControlSetMode (mConsoleMode);
     OcConsoleControlInstallProtocol (&mConsoleControlProtocol, NULL, NULL);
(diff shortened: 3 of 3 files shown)


Library/OcConsoleLib/TextOutputBuiltin.c
```diff
@@ -937,38 +946,6 @@ EFI_CONSOLE_CONTROL_PROTOCOL
...
-  }
-
-  gBS->InstallMultipleProtocolInterfaces (
-         &gST->ConsoleOutHandle,
-         &gEfiConsoleControlProtocolGuid,
-         &mConsoleControlProtocol,
-         NULL
-         );
-}
-
```
mikebeaton, line 971: I believe this removed, unused code is possibly a latent genetic memory of having faced the same issue which this PR addresses before.

Library/OcConsoleLib/TextOutputBuiltin.c
```diff
@@ -993,11 +970,11 @@ OcUseBuiltinTextOutput (
...
   DEBUG ((DEBUG_INFO, "OCC: Using builtin text renderer scale %u mode %u\n", mFontScale, Mode));
 
-  mConsoleMode = Mode;
+  mConsoleMode = EfiConsoleControlScreenText;
   Status       = AsciiTextResetEx (&mAsciiTextOutputProtocol, TRUE, TRUE);
 
   if (!EFI_ERROR (Status)) {
-    OcConsoleControlInstallProtocol (&mConsoleControlProtocol, NULL, NULL);
+    OcConsoleControlInstallProtocol (&mConsoleControlProtocol, &mOriginalConsoleControlProtocol, &mConsoleMode);
     OcConsoleControlSetMode (Mode);
```
mikebeaton, line 978: Prior to b07843fe1d47454747ae4eda9ea0189aa9fb8c03 , OcConsoleControlSetMode came first, which means we were really relying on the native implementation to (truly) set graphics mode, all along.

mikebeaton, line 978: Prior to b07843fe1d47454747ae4eda9ea0189aa9fb8c03 , OcConsoleControlSetMode came first, which means we were really relying on the native implementation to (truly) set graphics mode, all along.

mikebeaton:
I'm just trying to recreate logic of the previous fix, which is obviously relevant to assessing this one. I should have taken better notes or added more comments, sorry. But the issue is something like this: with the previous version of the code (i.e prior to the _previous_ fix), then with system renderer but where the existing console control protocol was missing or incomplete, then graphics mode did not get cleanly entered (it was either marked as entered when it wasn't properly entered or not marked as entered when it was, I think the latter). The net result of this was that later on, something which should have cleared the screen automatically wasn't doing so and there was an additional clear screen somewhere else in OpenCore which wasn't even needed once this issue was fixed! (I could, given a fair amount of time, recreate the exact details but that is roughly it.)

Library/OcConsoleLib/TextOutputBuiltin.c
```diff
@@ -909,10 +910,18 @@ ConsoleControlSetMode (
...
 {
   if (mConsoleMode != Mode) {
-    mConsoleMode = Mode;
-    if (mConsoleMode == EfiConsoleControlScreenText) {
+    //
+    // Our builtin renderer cannot genuinely change mode, so use original when present.
+    //
+    if (mOriginalConsoleControlProtocol.SetMode != NULL) {
+      mOriginalConsoleControlProtocol.SetMode (This, Mode);
+    }
```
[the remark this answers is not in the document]
vit9696: Ugh, why. It is by purpose not to call SetMode when using the builtin renderer. This change will clause flashes on many devices.

I agree with the issue, but disagree with the change. `Builtin` renderer should never call original `SetMode` on mode changes after entering the prescribed mode in the setting (Builtin → `Graphics` or Builtin → `Text`).

vit9696 (changes requested):

Library/OcConsoleLib/TextOutputBuiltin.c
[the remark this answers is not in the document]
mikebeaton: Okay, I'd better more carefully recreate the logic of the initial fix, then redo - with the initial set of the requested mode, only, done by the system renderer, then avoiding incomplete init in whatever way cleanly fits with that

mikebeaton:
The revised version now reverts https://github.com/acidanthera/OpenCorePkg/commit/b07843fe1d47454747ae4eda9ea0189aa9fb8c03 then adds another commit on top of that.

Just looking at the second commit, the Builtin code is now back to how it was before the original change. The System code is also back to how it was, in terms of setting the desired mode using the pre-existing console control protocol (when it exists), but is modified in a different way to achieve the purpose of the original change, so that the mode it is (supposed to be... acting as if it is...) in gets saved, even when there is no original console control protocol.

mikebeaton:
@vit9696 - Hi! I believe this is correct now. Builtin initialisation is back to how it was originally (thereby fixing the bug of failure to switch into graphics mode on some systems). I have also changed System text renderer back to how it was before https://github.com/acidanthera/OpenCorePkg/commit/b07843fe1d47454747ae4eda9ea0189aa9fb8c03 (i.e. also, as intended, using the pre-existing protocol to set the mode). So, instead of using the swapped order, I have switched to a different way of making sure mConsoleMode is set as requested, even if no previous cc protocol.

mikebeaton:
cc @dhinakg , @vit9696 -

The issue with simply reverting https://github.com/acidanthera/OpenCorePkg/commit/b07843fe1d47454747ae4eda9ea0189aa9fb8c03 is that mConsoleMode does not get set to graphics in SystemGraphics mode, with this code: https://github.com/acidanthera/OpenCorePkg/blob/revert-console-change/Library/OcConsoleLib/TextOutputSystem.c#L347-L348

This matters, e.g., because OpenCanopy switches back to its previous mode (obtained by GetMode, which will return console in this case), after password and after menu.

So, now, if it was always the intended behaviour that SystemGraphics would enter text mode, in the case where there was no previous console control protocol, then https://github.com/acidanthera/OpenCorePkg/commit/b07843fe1d47454747ae4eda9ea0189aa9fb8c03 can simply be reverted (although the change on one parameter from OUT to IN OUT, plus associated comment, is correct anyway, and should be kept).

If this wasn't always the intended behaviour, then the current state of this PR fixes it.

PR #444, opened 2023-04-07
OcConsoleLib: Additional info in GOP dump; additional guards on GopBurstMode

mikebeaton:
Should fix @khronokernel's issue; also fixes the issue observed on [PeterHolbrook](https://forums.macrumors.com/threads/opencore-on-the-mac-pro.2207814/post-32075548)'s graphics card (thx for assistance).

Since it adds a couple of new GOP utilities, I thought I would PR rather then push it directly. The new `GopUtils.c:OcGopModeBytesPerPixel` method here overlaps in functionality with the existing `OcBlitLib.c:BlitLibConfigurePixelFormat` but I thought it was best not to try to merge them, since getting a reliable bytes per pixel value is generally useful, wheres the existing BLIT configure method sets up some other things rather specific to its application (shl and shr values).

### DIFF
--- a/Library/OcConsoleLib/GopUtils.c
+++ b/Library/OcConsoleLib/GopUtils.c
@@ -1,6 +1,5 @@
 /** @file
-  Use PAT to enable write-combining caching (burst mode) on GOP memory,
-  when it is suppported but firmware has not set it up.
+  GOP buffer and pixel size utility methods, and GOP burst mode caching code.
 
   Copyright (C) 2023, Mike Beaton. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-3-Clause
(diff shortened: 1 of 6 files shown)


Library/OcConsoleLib/GopUtils.c
```diff
@@ -18,13 +17,114 @@
...
+      return EFI_SUCCESS;
+
+    case PixelBitMask:
+      break;
+
+    case PixelBltOnly:
+      return EFI_UNSUPPORTED;
+
+    default:
+      ASSERT (Mode->Info->PixelFormat < PixelFormatMax);
```
vit9696, line 67: What if we have some trash GOP implementation? Do we really want these asserts?

PR #446, opened 2023-04-11
Add Optional contentVisibility Qualification

dakanji:
It is possible to run multiple OpenCore instances but the `contentVisibility` feature currently does not allow targeting specific instances. This PR seeks to allow this via optional qualification of the contentVisibility instruction. The qualification string is of the form `@OPENCORE_PARENT_FOLDER_NAME@` and there can be more than one.

If qualification is not present, the contentVisibility instruction will work as it currently does. That is, apply to any instance.

An indication that an additional `Enable` instruction may be preferable was considered but while certainly easier, it appears to be far less flexible than qualifying the existing instructions. 
1. A new `Enable` instruction would be limited to disabling or enabling instances but qualifying allows application to both of the current instructions and can be extended to future instructions.
2. A new `Enable` instruction would not accommodate adding a new instance without changing expected default behaviour 

Needs an eye cast over some items such as memory allocation and leak avoidance in particular ... cc @PMheart

### DIFF
--- a/Library/OcBootManagementLib/BootEntryManagement.c
+++ b/Library/OcBootManagementLib/BootEntryManagement.c
@@ -200,7 +200,8 @@ STATIC
 INTERNAL_ENTRY_VISIBILITY
 ReadEntryVisibility (
   IN EFI_DEVICE_PATH_PROTOCOL  *DevicePath,
-  IN BOOLEAN                   IsFolder
+  IN BOOLEAN                   IsFolder,
+  IN CONST CHAR8               *InstanceName
   )
 {
   EFI_STATUS                       Status;
@@ -211,8 +212,12 @@ ReadEntryVisibility (
   UINTN                            VisibilityPathSize;
   CHAR16                           *VisibilityPath;
   CHAR16                           *VisibilityTerminator;
+  BOOLEAN                          VisibilityFlag;
+  BOOLEAN                          CheckQualifiers;
+  BOOLEAN                          QualifiedVisibility;
   BOOLEAN                          Result;
   CHAR8                            *Visibility;
+  CHAR8                            QualifierString[OC_MAX_INSTANCE_IDENTIFIER_SIZE];
 
   Status = gBS->LocateDevicePath (
                   &gEfiSimpleFileSystemProtocolGuid,
@@ -275,12 +280,21 @@ ReadEntryVisibility (
     }
   }
 
-  CopyMem (VisibilityTerminator, L".contentVisibility", sizeof (L".contentVisibility"));
+  CopyMem (
+    VisibilityTerminator,
+    L".contentVisibility",
+    sizeof (L".contentVisibility")
+    );
 
   //
   // Note, this guarantees nul-termination.
   //
-  Visibility = OcReadFile (FileSystem, VisibilityPath, NULL, 64);
+  Visibility = OcReadFile (
+                 FileSystem,
+                 VisibilityPath,
+                 NULL,
+                 OC_MAX_VISIBILITY_INSTRUCTION_SIZE
+                 );
 
   FreePool (VisibilityPath);
 
@@ -288,19 +302,106 @@ ReadEntryVisibility (
     return BootEntryNormal;
   }
 
+  //
+  // Handle Visibility qualification presence/absence.
+  //
+  if (AsciiStrStr (Visibility, "~") == NULL) {
+    QualifiedVisibility = FALSE;
+    CheckQualifiers     = FALSE;
+  } else {
+    CheckQualifiers = TRUE;
+
+    AsciiSPrint (
+      QualifierString,
+      sizeof (QualifierString),
+      "~%a~",
+      InstanceName
+      );
+
+    if (AsciiStrStr (Visibility, QualifierString) != NULL) {
+      QualifiedVisibility = TRUE;
+    } else {
+      QualifiedVisibility = FALSE;
+    }
+  }
+
+  VisibilityFlag = FALSE;
+
+  //
+  // Handle 'Disabled' instruction.
+  //
   if (AsciiStrnCmp (Visibility, "Disabled", L_STR_LEN ("Disabled")) == 0) {
-    FreePool (Visibility);
-    return BootEntryDisabled;
+    //
+    // Check absolute content if content qualification is absent.
+    //
+    if (!CheckQualifiers) {
+      VisibilityFlag = TRUE;
+      DEBUG ((
+        DEBUG_INFO,
+        "OCB: Found .contentVisibility with instruction:- 'Disabled'\n"
+        ));
+    }
+
+    //
+    // Check for qualified contents if absolute contents were not
+    // previously found and content qualification is present.
+    //
+    if (!VisibilityFlag && CheckQualifiers && QualifiedVisibility) {
+      VisibilityFlag = TRUE;
+      DEBUG ((
+        DEBUG_INFO,
+        "OCB: Found .contentVisibility with instruction:- 'Disabled ... %a'\n",
+        InstanceName
+        ));
+    }
+
+    if (VisibilityFlag) {
+      FreePool (Visibility);
+
+      return BootEntryDisabled;
+    }
   }
 
+  //
+  // Handle 'Auxiliary' instruction.
+  //
   if (AsciiStrnCmp (Visibility, "Auxiliary", L_STR_LEN ("Auxiliary")) == 0) {
-    FreePool (Visibility);
-    return BootEntryAuxiliary;
+    //
+    // Check absolute content if content qualification is absent.
+    //
+    if (!CheckQualifiers) {
+      VisibilityFlag = TRUE;
+      DEBUG ((
+        DEBUG_INFO,
+        "OCB: Found .contentVisibility with instruction:- 'Auxiliary'\n"
+        ));
+    }
+
+    //
+    // Check for qualified contents if absolute contents were not
+    // previously found and content qualification is present.
+    //
+    if (!VisibilityFlag && CheckQualifiers && QualifiedVisibility) {
+      VisibilityFlag = TRUE;
+      DEBUG ((
+        DEBUG_INFO,
+        "OCB: Found .contentVisibility with instruction:- 'Auxiliary ... %a'\n",
+        InstanceName
+        ));
+    }
+
+    if (VisibilityFlag) {
+      FreePool (Visibility);
+
+      return BootEntryAuxiliary;
+    }
   }
 
-  DEBUG ((DEBUG_INFO, "OCB: Discovered unsupported .contentVisibility\n"));
-
+  //
+  // Visibility instruction is not applicable or is not valid.
+  //
   FreePool (Visibility);
+
   return BootEntryNormal;
 }
 
@@ -463,8 +564,13 @@ AddBootEntryOnFileSystem (
   //
   // Do not add recoveries when not requested (e.g. can be HFS+ recovery).
   //
-  if (BootContext->PickerContext->HideAuxiliary && (EntryType == OC_BOOT_APPLE_RECOVERY)) {
-    DEBUG ((DEBUG_INFO, "OCB: Discarding recovery entry due to auxiliary\n"));
+  if (  BootContext->PickerContext->HideAuxiliary
+     && (EntryType == OC_BOOT_APPLE_RECOVERY))
+  {
+    DEBUG ((
+      DEBUG_INFO,
+      "OCB: Discarding recovery entry due to 'Auxiliary' setting\n"
+      ));
     if (IsReallocated) {
       FreePool (DevicePath);
     }
@@ -475,8 +581,13 @@ AddBootEntryOnFileSystem (
   //
   // Do not add Time Machine when not requested.
   //
-  if (BootContext->PickerContext->HideAuxiliary && (EntryType == OC_BOOT_APPLE_TIME_MACHINE)) {
-    DEBUG ((DEBUG_INFO, "OCB: Discarding time machine entry due to auxiliary\n"));
+  if (  BootContext->PickerContext->HideAuxiliary
+     && (EntryType == OC_BOOT_APPLE_TIME_MACHINE))
+  {
+    DEBUG ((
+      DEBUG_INFO,
+      "OCB: Discarding time machine entry due to 'Auxiliary' setting\n"
+      ));
     if (IsReallocated) {
       FreePool (DevicePath);
     }
@@ -487,9 +598,13 @@ AddBootEntryOnFileSystem (
   //
   // Skip disabled entries, like OpenCore bootloader.
   //
-  Visibility = ReadEntryVisibility (DevicePath, IsFolder);
+  Visibility = ReadEntryVisibility (
+                 DevicePath,
+                 IsFolder,
+                 BootContext->PickerContext->InstanceIdentifier
+                 );
   if (Visibility == BootEntryDisabled) {
-    DEBUG ((DEBUG_INFO, "OCB: Discarding disabled entry by visibility\n"));
+    DEBUG ((DEBUG_INFO, "OCB: Entry disabled via .contentVisibility\n"));
     if (IsReallocated) {
       FreePool (DevicePath);
     }
@@ -501,7 +616,7 @@ AddBootEntryOnFileSystem (
   // Skip custom auxiliary entries.
   //
   if ((Visibility == BootEntryAuxiliary) && BootContext->PickerContext->HideAuxiliary) {
-    DEBUG ((DEBUG_INFO, "OCB: Discarding auxiliary entry by visibility\n"));
+    DEBUG ((DEBUG_INFO, "OCB: Entry hidden via .contentVisibility\n"));
     if (IsReallocated) {
       FreePool (DevicePath);
     }
--- a/Library/OcBootManagementLib/BootEntryManagement.c
+++ b/Library/OcBootManagementLib/BootEntryManagement.c
@@ -328,18 +328,34 @@ ReadEntryVisibility (
   VisibilityFlag = FALSE;
 
   //
-  // Handle 'Disabled' instruction.
+  // Handle instruction: 'Ignore'.
+  // Handle instruction: 'Disabled' (Deprecated).
+  // Former instruction deprecated for legacy compatibility.
   //
-  if (AsciiStrnCmp (Visibility, "Disabled", L_STR_LEN ("Disabled")) == 0) {
+  if (  (AsciiStrnCmp (Visibility, "Ignore", L_STR_LEN ("Ignore")) == 0)
+     || ((AsciiStrnCmp (Visibility, "Disabled", L_STR_LEN ("Disabled")) == 0)))
+  {
     //
     // Check absolute content if content qualification is absent.
     //
     if (!CheckQualifiers) {
       VisibilityFlag = TRUE;
-      DEBUG ((
-        DEBUG_INFO,
-        "OCB: Found .contentVisibility with instruction:- 'Disabled'\n"
-        ));
+
+      DEBUG_CODE_BEGIN ();
+
+      if (AsciiStrnCmp (Visibility, "Ignore", L_STR_LEN ("Ignore")) == 0) {
+        DEBUG ((
+          DEBUG_INFO,
+          "OCB: Found visibility instruction:- Ignore on all instances\n"
+          ));
+      } else {
+        DEBUG ((
+          DEBUG_INFO,
+          "OCB: Found deprecated visibility instruction:- Disabled on all instances\n"
+          ));
+      }
+
+      DEBUG_CODE_END ();
     }
 
     //
@@ -348,11 +364,28 @@ ReadEntryVisibility (
     //
     if (!VisibilityFlag && CheckQualifiers && QualifiedVisibility) {
       VisibilityFlag = TRUE;
-      DEBUG ((
-        DEBUG_INFO,
-        "OCB: Found .contentVisibility with instruction:- 'Disabled ... %a'\n",
-        InstanceName
-        ));
+
+      DEBUG_CODE_BEGIN ();
+
+      if (AsciiStrnCmp (Visibility, "Ignore", L_STR_LEN ("Ignore")) == 0) {
+        DEBUG ((
+          DEBUG_INFO,
+          "OCB: Found visibility instruction:- Ignore on '%a' instances\n",
+          InstanceName
+          ));
+      } else {
+        DEBUG ((
+          DEBUG_INFO,
+          "OCB: Found deprecated visibility instruction:- Disabled on '%a' instances\n",
+          InstanceName
+          ));
+        DEBUG ((
+          DEBUG_WARN,
+          "OCB: Deprecated 'Disabled' instruction should not be qualified! Replace with 'Ignore'\n"
+          ));
+      }
+
+      DEBUG_CODE_END ();
     }
 
     if (VisibilityFlag) {
@@ -363,18 +396,34 @@ ReadEntryVisibility (
   }
 
   //
-  // Handle 'Auxiliary' instruction.
+  // Handle instruction: 'Cloak'.
+  // Handle instruction: 'Auxiliary' (Deprecated).
+  // Former instruction deprecated for legacy compatibility.
   //
-  if (AsciiStrnCmp (Visibility, "Auxiliary", L_STR_LEN ("Auxiliary")) == 0) {
+  if (  (AsciiStrnCmp (Visibility, "Cloak", L_STR_LEN ("Cloak")) == 0)
+     || ((AsciiStrnCmp (Visibility, "Auxiliary", L_STR_LEN ("Auxiliary")) == 0)))
+  {
     //
     // Check absolute content if content qualification is absent.
     //
     if (!CheckQualifiers) {
       VisibilityFlag = TRUE;
-      DEBUG ((
-        DEBUG_INFO,
-        "OCB: Found .contentVisibility with instruction:- 'Auxiliary'\n"
-        ));
+
+      DEBUG_CODE_BEGIN ();
+
+      if (AsciiStrnCmp (Visibility, "Cloak", L_STR_LEN ("Cloak")) == 0) {
+        DEBUG ((
+          DEBUG_INFO,
+          "OCB: Found visibility instruction:- Cloak on all instances\n"
+          ));
+      } else {
+        DEBUG ((
+          DEBUG_INFO,
+          "OCB: Found deprecated visibility instruction:- Auxiliary on all instances\n"
+          ));
+      }
+
+      DEBUG_CODE_END ();
     }
 
     //
@@ -383,11 +432,28 @@ ReadEntryVisibility (
     //
     if (!VisibilityFlag && CheckQualifiers && QualifiedVisibility) {
       VisibilityFlag = TRUE;
-      DEBUG ((
-        DEBUG_INFO,
-        "OCB: Found .contentVisibility with instruction:- 'Auxiliary ... %a'\n",
-        InstanceName
-        ));
+
+      DEBUG_CODE_BEGIN ();
+
+      if (AsciiStrnCmp (Visibility, "Cloak", L_STR_LEN ("Cloak")) == 0) {
+        DEBUG ((
+          DEBUG_INFO,
+          "OCB: Found visibility instruction:- Cloak on '%a' instances\n",
+          InstanceName
+          ));
+      } else {
+        DEBUG ((
+          DEBUG_INFO,
+          "OCB: Found deprecated visibility instruction:- Auxiliary on '%a' instances\n",
+          InstanceName
+          ));
+        DEBUG ((
+          DEBUG_WARN,
+          "OCB: Deprecated 'Auxiliary' instruction should not be qualified! Replace with 'Cloak'\n"
+          ));
+      }
+
+      DEBUG_CODE_END ();
     }
 
     if (VisibilityFlag) {
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@ OpenCore Changelog
 ==================
 #### v0.9.4
 - Fixed kext blocker `Exclude` strategy for prelinked on 32-bit versions of macOS
+- Added `InstanceIdentifier` setting and option to limit `.contentVisibility` to specific OpenCore instances
 
 #### v0.9.3
 - Added `--force-codec` option to AudioDxe, thx @xCuri0
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -3148,21 +3148,21 @@ The algorithm to determine boot options behaves as follows:
   \item For file device paths, check for presence on the file system directly.
   % Just kill all \EFI\APPLE\ paths.
   \item Exclude entries if there is a \texttt{.contentVisibility} file (near a loader or within a boot directory)
-  with contents starting with \texttt{Disabled}. \\
+  with contents starting with \texttt{Ignore}. \\
   See \texttt{Qualified .contentVisibility files} below for more.
   \item Mark device handle as \textit{used} in the list of partition handles if any.
   % Each partition handle will basically have a list of boot option entries for quick lookup later.
   \item Register the resulting entries as primary options and determine their types. \\
   The option will become auxiliary for some types (e.g. Apple HFS recovery)
-  or if its \texttt{.contentVisibility} file contents start with \texttt{Auxiliary}. \\
+  or if its \texttt{.contentVisibility} file contents start with \texttt{Cloak}. \\
   See \texttt{Qualified .contentVisibility files} below for more.
   \item Qualified \texttt{.contentVisibility} files.
     \begin{itemize}
     \item \texttt{.contentVisibility} files may optionally target only specific instances
-    of OpenCore when such files contain \texttt{(Disabled|Auxiliary) [\{Instance-Set\}]}. \\
+    of OpenCore when such files contain \texttt{(Ignore|Cloak) [\{Instance-Set\}]}. \\
     Where \texttt{Instance-Set} is a set of one or more OpenCore \texttt{InstanceIdentifiers},
     specified in \texttt{config.plist}, each bounded by a tilde (\texttt{~}). \\
-    For example, if \texttt{Disabled ~ABC~XYZ~} is entered, the target item will be ignored on instances with
+    For example, if \texttt{Ignore ~ABC~XYZ~} is entered, the target item will be ignored on instances with
     \texttt{ABC} or \texttt{XYZ} defined as \texttt{InstanceIdentifier} in their \texttt{config.plist} files.
     \item When an \texttt{Instance-Set} is present in a \texttt{.contentVisibility} file, the
     visibility instruction is only applied if the \texttt{InstanceIdentifier} value of the
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -3125,6 +3125,8 @@ installed on the media, while alternate options represent recovery options for t
 \end{itemize} \medskip
 \end{itemize}
 
+\subsubsection{Boot Algorithm}\label{bootalgorithm}
+
 The algorithm to determine boot options behaves as follows:
 
 \begin{enumerate}
@@ -3145,18 +3147,33 @@ The algorithm to determine boot options behaves as follows:
   \item For disk device paths (not specifying a bootloader), execute ``bless'' (may return > 1 entry).
   \item For file device paths, check for presence on the file system directly.
   % Just kill all \EFI\APPLE\ paths.
-  \item Exclude entries if there is a \texttt{.contentVisibility} file near the bootloader or
-  inside the boot directory with \texttt{Disabled} contents (ASCII).
+  \item Exclude entries if there is a \texttt{.contentVisibility} file (near a loader or within a boot directory)
+  with contents starting with \texttt{Disabled}. \\
+  See \texttt{Qualified .contentVisibility files} below for more.
   \item Mark device handle as \textit{used} in the list of partition handles if any.
-  % Each partition handle will basically have a list of boot option entries for later quick lookup.
+  % Each partition handle will basically have a list of boot option entries for quick lookup later.
   \item Register the resulting entries as primary options and determine their types. \\
   The option will become auxiliary for some types (e.g. Apple HFS recovery)
-  or if its \texttt{.contentVisibility} file contains \texttt{Auxiliary}.
+  or if its \texttt{.contentVisibility} file contents start with \texttt{Auxiliary}. \\
+  See \texttt{Qualified .contentVisibility files} below for more.
+  \item Qualified \texttt{.contentVisibility} files.
+    \begin{itemize}
+    \item \texttt{.contentVisibility} files may optionally target only specific instances
+    of OpenCore when such files contain \texttt{(Disabled|Auxiliary) [\{Instance-Set\}]}. \\
+    Where \texttt{Instance-Set} is a set of one or more OpenCore \texttt{InstanceIdentifiers},
+    specified in \texttt{config.plist}, each bounded by a tilde (\texttt{~}). \\
+    For example, if \texttt{Disabled ~ABC~XYZ~} is entered, the target item will be ignored on instances with
+    \texttt{ABC} or \texttt{XYZ} defined as \texttt{InstanceIdentifier} in their \texttt{config.plist} files.
+    \item When an \texttt{Instance-Set} is present in a \texttt{.contentVisibility} file, the
+    visibility instruction is only applied if the \texttt{InstanceIdentifier} value of the
+    current OpenCore instance is included in the set. When \texttt{Instance-Set} is absent,
+    visibility instructions are applied to all OpenCore instances.
+    \end{itemize}
   \end{itemize}
 \item For each partition handle:
   \begin{itemize}
   \item If the partition handle is marked as \textit{unused}, execute ``bless'' primary option list retrieval. \\
-    In case a \texttt{BlessOverride} list is set, both standard and custom ``bless'' paths will be found.
+  In case a \texttt{BlessOverride} list is set, both standard and custom ``bless'' paths will be found.
   \item On the OpenCore boot partition, exclude OpenCore bootstrap files using header checks.
   \item Register the resulting entries as primary options and determine their types if found. \\
   The option will become auxiliary for some types (e.g. Apple HFS recovery).
@@ -3172,7 +3189,7 @@ The algorithm to determine boot options behaves as follows:
 The display order of the boot options in the OpenCore picker and the boot process
 are determined separately from the scanning algorithm.
 
-The display order as follows:
+The display order is as follows:
 
 \begin{itemize}
 \tightlist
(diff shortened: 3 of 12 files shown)


mikebeaton:
This appears invalid. `.contentVisibility` currently targets the .efi file which it is located next to.

I have spotted that the existing documentation says: "Exclude entries if there is a .contentVisibility file near the bootloader or inside the boot directory with `Disabled` contents (ASCII)." This may be a source of some confusion, as 'the bootloader' here is not OpenCore, but e.g. `boot.efi` or `bootmgfw.efi`. (Btw 'or inside the boot directory' refers to folder-based boot items such as Recovery DMG - but again it is per-boot entry.)

dakanji:
> "Exclude entries if there is a .contentVisibility file near the bootloader or inside the boot directory with Disabled contents (ASCII)."

This now becomes what is written (as with the `Auxiliary` instruction) and stays that way unless **"qualified"** to only happen when booted via specified OpenCore instances. If there is no need for limiting to specific instances, then the default **"unqualified"** instruction is used and things will work as it has always done, which it for ContentVis to always work as per the instruction for all Opencore instance.  In other words, you only **"qualify"** instructions where you specifically need to. Hence the **"Optional"** discription.

> contentVisibility currently targets the .efi file which it is located next to.

Yes, that is the point of ContentVis.  It applies the instruction to the target efi file it is located next/close to as said. For those of us that run multiple OpenCore instances, there is a need for the instruction to be applied to a target efi file in one instance but not in another.

So if I have four OpenCore instances, **"OC_A""**, **"OC_B"**, **"OC_C"** and **"OC_D"** and three efi files, **"EFI_1"**, **"EFI_2"** and **"EFI_3"**.

Now, I want the following:
1. **EFI_1** to be disabled in **OC_A** and **OC_D**
2. **EFI_2** to be disabled in **OC_B**
3. **EFI_1** and **EFI_2** to show up in **OC_C**
4. **EFI_3** to be disabled in all four

This currently cannot be done and I either have to disable them all and leave them all.
With this PR, it becomes possible by doing the following:
1. **EFI_1** : ContentVis file will contain `Disabled@OC_A@OC_D@`
2. **EFI_2** : ContentVis file will contain `Disabled@OC_B@`
3. **EFI_3** : ContentVis file will contain `Disabled`

Result:
* When booted via **OC_A:** 
  - **EFI_1** will be **disabled**
  - **EFI_2** will be enabled
  - **EFI_3** will be **disabled**
* When booted via **OC_B:** 
  - **EFI_1** will be enabled
  - **EFI_2** will be **disabled**
  - **EFI_3** will be **disabled**
* When booted via **OC_C:** 
  - **EFI_1** will be enabled
  - **EFI_2** will be enabled
  - **EFI_3** will be **disabled**
* When booted via **OC_D:** 
  - **EFI_1** will be **disabled**
  - **EFI_2** will be enabled
  - **EFI_3** will be **disabled**

So, as it was apparently not as clear as I thought in the description, this PR is not about disabling OpenCore instances (which is not what ContentVis is about to start with as was stated), but to control OpenCore instances where the standard ContentVis instructions would be active. A subtle but significant difference.

Library/OcBootManagementLib/BootEntryManagement.c
```diff
@@ -288,19 +294,137 @@ ReadEntryVisibility (
...
+        OCFolderName = OcCopyDevicePathFullName (AbsPath, NULL);
+        FreePool (AbsPath);
+      }
+    }
+
+    if (OCFolderName != NULL) {
+      CheckQualifiers = UnicodeGetParentDirectory (OCFolderName);
+    }
+
+    if (CheckQualifiers == TRUE) {
```
PMheart, line 329: For all booleans, we simplify use `if (CheckQualifiers)` and `if (!CheckQualifiers)`

Library/OcBootManagementLib/BootEntryManagement.c
```diff
@@ -288,19 +294,137 @@ ReadEntryVisibility (
...
+      QualifierString = AllocatePool (sizeof (CHAR16) * (StrLen (OCFolderName) + (L_STR_LEN ("@") * 2) + 1));
```
PMheart, line 330: sizeof a variable makes more sense.

dakanji:
Tagged as draft as not enough testing done

mikebeaton:
> So, as it was apparently not as clear as I thought in the description, this PR is not about disabling OpenCore instances (which is not what ContentVis is about to start with as was stated), but to control OpenCore instances where the standard ContentVis instructions would be active. A subtle but significant difference.

Thank you for the clarification. Not that subtle a distinction, but subtle enough that I didn't understand what you were trying to do without the additional info.

Okay, I now understand what you are trying to do - thanks again - but could you now provide some motivation as to why you are trying to do this? That is, why do you need this proposed feature? Exactly what will go wrong (presumably?) for you without it?

dakanji:
The motivation was set out here:

> For those of us that run multiple OpenCore instances, there is often a need for ContentVis instructions to be applied to a target efi file when booted into one OpenCore instance but not when booted into another OpenCore instance.

The primary reason behind having multiple instances is to be able to run things in one instance that cannot be run in another. I would have thought the desirability of being able to only show what can be run in a specific instance would be obvious.

Regardless, is it not the case once multiple OpenCore instances are permitted, that it should also be possible to separately configure items such as ContentVis and not only have a one-size-fits-all setup?

Library/OcBootManagementLib/BootEntryManagement.c
dakanji: Thanks ... updated

Library/OcBootManagementLib/BootEntryManagement.c
```diff
@@ -288,19 +294,137 @@ ReadEntryVisibility (
...
+      VisibilityFlag = TRUE;
+    }
+
+    //
+    // Check for qualified contents if 'absolute' contents were
+    // not previously found and content qualification is present
+    //
+    if (  (CheckQualifiers == TRUE)
+       && (VisibilityFlag == FALSE)
+       && (StrStr (QualifierString, (const CHAR16 *)Visibility) != NULL))
```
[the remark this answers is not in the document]
dakanji: Thanks ... updated

dakanji:
@PMheart ... Unrelated item I came across while researching is that this seems to be a bug (testing the same condition twice):

https://github.com/acidanthera/OpenCorePkg/blob/c42343b6bf3207ca0cb0145b46345da76627a3bb/Application/Bootstrap/Bootstrap.c#L64-L65

Blame says you were the last to edit it although @vit9696 apparently originally added it.

mikebeaton:
There was some internal discussion a little while ago about whether certain OC features should perhaps be keyed to macOS version, in order to make booting different versions of macOS from the same OpenCore instance easier.

But one upshot of that discussion was, that we are _not_ currently aware of any combinations of, say, a pair of macOS versions which could not both be booted successfully, from just one correctly constructed config.

So I'll repeat again that it will very much help to have a clear description of something which can't currently be done with one OpenCore instance, and which needs two or more.

Also, I believe it would really help to raise this discussion in a new Issue - e.g. 'OpenCore can't currently boot both abc and def from the same instance', with a link to this PR as one proposed solution.

dakanji:
What can be done with one instance or not is a side issue as far as this PR is concerned and  I am sure you and the others in your discussions are perfectly capable of independently concluding on that item.

If the idea is that multiple instances should not be used or are not required, as is coming across, then the capability should simply be removed. As long as the capability is there however, then it should also be possible to separately fully configure them as a matter of principle. Especially if there is no negative impact to doing so.

So, while I understand you want to discuss the need for multiple instances in a separate issue etc, I am sorry but I am not keen on entering that discussion at this time; bearing in mind that the person that added the capability to start with should already be in your discussions and can explain why it was added in the first place.

mikebeaton:
> There has been a persistent misunderstanding of what this PR is about from the first intervention

Really? Perhaps you could clarify what you believe persisted.

This PR came out of the blue, you did not raise an issue first - even clarifying why you believe this is needed, much less discussing whether your own suggested solution is the best/preferred/cleanest approach.

Personally, I believe _both_ discussions are needed before this PR (or any revised version of it) should be accepted. But since you're not willing to even start on the first discussion, I'll leave it to others to comment.

dakanji:
I am sure a judgment can easily be made as to whether adding these few lines of code is the best/preferred/cleanest approach to making the contentVisibility function work for multiple OpenCore instances (the scope of the PR) without needing discussions on out of scope issues such as the need for multiple instances. 

I do agree with you that it is best there is an objective review and comments and that is what I had hoped for to start with.

mikebeaton:
@dakanji - I can think of somewhat cleaner ways to achieve what you're trying to achieve, which the powers that be agree with me are somewhat cleaner, but there is also agreement that the case for doing this in the first place has not yet been made, I'm afraid!

It's worth bearing in mind that individual instances of OC always need hiding from all other instances of OC, so ofc there is no case based just on that usage for making the hiding be installation-specific - in fact more a case against. (Not just my opinion.)

So I'd still suggest that it would be in your best interests - if you want this feature - to make a case along the lines of "I cannot do this (something specific; not just a general principle), which it is reasonable to want to do in OC, therefore I need this feature".

dakanji:
Thanks for the update which was as expected.
In that case anyway, I will just close and move on.

dakanji:
Okay. Apparently not a hard '**No**' as _yet_.
Will continue building this in the interim while the powers ruminate.

mhaeuser:
@dakanji I’m majorly confused by your vague language, so I will try to be clear. Features for the sake of “if this is allowed, it *obviously* makes sense to allow…” will not be accepted. Every feature needs a clear and unambiguous motivation, as scope follows use cases, not vice-versa. If you are not ready to motivate the change and instead deflect to “ask whoever…”, justifying merging is tough.

To still give context on the support, it was added for OC debugging, not to run multiple production OCs in parallel. Regression and feature testing simply proved to be annoying when the OC folder had to be swapped each time a new test is run. This use case is not sufficient to motivate your change as you implied.

Still, discussions are ongoing, but they are only so fruitful when there is a lot of guesswork and puzzling forum posts together to get an idea of the motivation.

dakanji:
@mhaeuser ... Thanks for shedding some light on a few things.

> Features for the sake of “if this is allowed, it obviously makes sense to allow…” will not be accepted.

I need to clarify that I am not proposing any new feature but simply trying to make one existing feature, ContentVis, work better with another existing feature, Multiple Instances.

> scope follows use cases, not vice-versa.

Makes perfect sense but can _sometimes_ be short sighted and limiting. As an aside, the principle of making things work as best as they can is important regardless of whether one person or another can articulate why at that point in time. That is, where the scope does in fact go ahead of the current use case. This is more so the case when this can be trivially done without a negative impact. However, this is a digression on my part and whatever works best for the team is perfectly fine.

One thing the team has consistently demonstrated is that while it may sometimes be initially be blinded by use case detail requirements, it generally does get there in the end and this is great. Adding the capability to use multiple features to aid development and testing is a classic example of something that was initially dismissed out of hand when first suggested before it turns out that it was ultimately realised this was indeed the case. 

> If you are not ready to motivate the change and instead deflect to “ask whoever…”

Never deflected and just wanted to stay within my scope as set out. You see, the team is by essence seen as a single unit from outside and it seemed crazy for it to turn around to ask me to justify an existing feature it added itself.

From my perspective, what was going on was that the team was saying:
```
WE, the OpenCore Team, added the capability to use multiple instances to OpenCore. 
YOU are however not allowed to make it work better unless YOU can justify this feature
that WE added to OpenCore.
```
In light of your clarifications however, it is actually not as irrational as it came across.

As for motivating the change, the issue was simply because I reached a conclusion, based on the discussion that took place, that the chances of success were non existent. You see, whatever I come up with is always likely to be considered to be somewhat of an edge case.

That something might be seen as somewhat of an edge case is not by itself a reason not to advance it for consideration as long as it is not going against the grain and has an advocate. So, it is normal that adding some obscure detail related to running OS X Tiger would be perfectly fine but some other item would require going through a session with an inquisitor without ending up tussled up nicely for the stake.

So while I could not quite articulate it until you clarified things, I had concluded regardless that there was some sort of default position against end users leveraging multiple instances. This was apparent because the queries were never about the details and mechanics of the proposal I made but on why multiple features were being used in the first place. There was no value judgment made on the position by the way, just an understanding that it existed and an understanding of the implications of its existence. This, coupled with the apparent oddity of the team's query, pointed at a _"classic going through the motions before being kicked into the long grass at best"_ exercise that I did not have the energy to even begin think of participating in. This seemed confirmed by the feedback that on discussion up on Mount Olympus, the thumbs were trending down. 

Given however that I may have been wrong about _"going through the motions on route to the long grass"_ following your clarifications, following are some basic use cases.

# TL;DR
Multiple instances allow a huge amount of flexibility. Some users do develop and want to be able to easily recover when things go south but that as you said is not a biggie. I can run everything from Ventura to Tiger via OpenCore on my unit. It was mentioned in an earlier post that it is possible to boot them all from one config file. I take that at face value but unless forced to, I would rather avoid such a config file and would very much prefer to continue to use tailored config files for ranges of boot targets as such configs are far easier to maintain.

Additionally, some use older units and are constrained to older versions of tools. One that has appeared quite a few times is a requirement to run Final Cut Pro on High Sierra without GPU acceleration but wanting this acceleration when running other Mac OS instances. The potential need for one option or another to be active in one instance but not in another is infinite. I am sure some other stuff escapes my mind having rambled so much.

However, all that can easily be considered edge cases as noted earlier. As also noted, stuff being considered somewhat of an edge case is typically not the primary determinant between the Grand inquisitor offering a benediction and his signet to be kissed or him nodding to the burly chaps lurking in the shadows. The main thing is whether it fits or it doesn't fit.

Sorry, can't resist the puzzling style

vit9696:
Hey, before I look at this, let me clarify about multiple instances:

- It is not officially supported still.
- We might be close to accepting multiple instance feature and we need more reasons why it is useful.
- After we are settled that multiple instances is fine we need more reasons why multiple instances need different boot options.

Current list I was able to build from your post is:

1. Having a backup bootloader in case upgrading the main one results in issues. I would say it could sit on a USB flash, but a user may want to permanently have this on one of the main drives.
2. Having different setups to support different macOS versions. The particular case is hardware acceleration options.
3. Having different setups to reduce test load for the end user of new releases. One can use older working version to boot older OS and newer/experimental version for newer OS or e.g. tests.
4. Having different setups to support different operating system configurations. For example, one can have borked ACPI which he has no spare time to fix right now, for instance, but can boot the OS without it.

The way these listed features apply to a feature request of locking boot entries to different instances is as follows:

1. Feels like the list of the options should be the same, only the backup OC instance should be hidden. For this the existing functionality is ok.
2. This might be the case where extending OC configuration can be more difficult than supporting multiple boot options. But we need short examples with option names for this.
3. This might also be the case. Is it relevant?
4. Feels like all unsupported OS options can be disabled with the existing functionality and be then "BIOS"-booted. However, I think I can remember a situation where a second OC is justified — Windows SLIC table.

Before we proceed, please answer the following questions (ideally shortly):

- Did I miss any case?
- Did I miss a reason why multiple OC instances with different sets of boot options are needed to cover this case?
- What are the particular configuration differences between different macOS versions for hardware accelerations on MP? A short list of keys/values would do.

mhaeuser:
@dakanji Thanks for the elaborate answer. The use case is still not clear to me, but I’ll still try respond.

Regarding the dismissal of proposals: Yes, that happens, but very rarely. Often contributors mistake the rejection of a specific implementation as the rejection of the general idea. For example, years ago, merely hiding recovery partitions was rejected, but not scanning them to begin with (which makes a difference for performance, memory consumption, etc.) was eventually supported. The original PR for letting OC run from any path was rejected for its terrible code quality, but our comments actually explicitly stated we’re open to have a sane version contributed. But yes, in rare instances, minds can change. :)

On the dwelling for use cases: Have you ever had someone less proficient than you in some area ask you how to do XY and you wondered why XY would be needed to be done, only to realise there is a better (easier, faster, …) solution to the actual problem to solve? This is why exact use case definitions are important. For example, there was the theory that the use case was booting a broader range of macOS versions on the same machine, and in this case it might be worthwhile to look at how to make one instance of OC work better with that rather than forcefully using multiple. We’re looking for solutions to a problem, not for solutions in search of a problem.

As for this proposal, it will definitely not be accepted as-is. The main reason is that a path does not semantically identify an OC instance. Say, you have your main instance at `OC`. Now you want to update it with a backup, so you move `OC` to `OC_BKP` and paste a new `OC` folder. You also have some alternative instance `OC_A`, which you did not touch, but you do have different visibility rules set up for both of them for some reason. Now, unless you specifically handled `OC_BKP`, it now behaves differently just from renaming the folder, while it’s still the exact same instance as before. Worse yet, it will behave differently from `OC`, which is just a version-refreshed variant, for no obvious reason (you have to deduct the behaviour from the exact way you set up the files or from the log). You may argue `OC_BKP` specifically could just be considered in the visibility files, but you can construct arbitrarily complex testing/debugging scenarios where it becomes absurd.

One idea floating around is to solve this with a value in the config. This may be called an ID, a class, or whatever really, but it would control the visibility behaviour rather than the source path. Thus, duplicating OC instances to different paths preserves their semantics.

But first, we need to make sure this is actually the best solution to a real-world problem. :)

dakanji:
@vit9696 ... I need to wrap my head around your post a bit later as I didn't fully understand at first glance. Perhaps it will be clearer if someone else chimes in. In the meantime, I will push a "Final/Working" version of this presently and then it's que sera, sera for that!

mikebeaton:
@dakanji - Hi. I think you have same alternative OC configs suggested on your MR thread for VMM, non-VMM, and LEGACY? (I may not have these exactly right - and there may be some others?) If they're not already listed in Vit's post, could you clarify when these different OC versions come in useful? (Don't assume it's a trick or loaded question - it's not - and assume that if I'm asking, it's because I don't fully know the answer, and that it would be helpful for getting to the bottom of this to have the answer explained (reasonably patiently, but reasonably briefly). Thanks!)

dakanji:
@mikebeaton ... MyBootMgr will be a target but note that what it creates varies based on several params. It would be potentially misleading to consider it the sum of what is envisaged but here goes off the top of my head...

1. Whether the GPU can be accelerated or not and whether running Nvidia or AMD
2. Whether I am running it or someone else 
3. The Mac OS versions the user specifies.
4. Whether they want to incorporate the OCLP or not
5. Whether running DosDude patched instances or not

In any case, if understanding what it creates will help, they are (Assuming the user is running Mojave/Older to Ventura):
1. OC_AAA: Boots Catalina to Monterey (Spoofs MacPro7,1 for GPU Acceleration depending on GPU and SecureBootModel)
2. OC_VMM: For Installing/Updating Monterey and Older (Spoofs VMM via CPUID)
3. OC_LEG: For Booting BigSur and Older (Spoofs iMacPro1,1 for GPU Acceleration and no SecureBootModel)
4. OC_EXT: For holding OCLP Configured Ventura (No idea what that does in detail)
5. OC_DEV: For development. (Only created when run on my unit)
6. OC_WRK: Known working. (Only created when run on my unit)

There is a Tiger Specific instance (for me only but currently in use).
 
There is no standard "OC" instance at this time.

Hope that is clear but as mentioned, there are other setups that I don't know how they are set up but they seem to have used MyBootMgr as a base. Totally different folder names obviously.

mikebeaton:
It helps. Thank you.

dakanji:
Finalised to do what is says on the tin:
1. User can target what to show or hide per OpenCore instance
2. User can specify this for any efi file and not limited to one OS such as Mac OS
3. If a new OpenCore instance is added, qualified ContentVis files will not apply
4. Qualification can be applied to 'Auxiliary' (now) or new instructions added in future
5. Qualification is optional and plain instructions can still be used

dakanji:
@vit9696 ... Was finally able to properly digest your post in light of background stuff I think I now better understand.

> Did I miss any case?

Seems fairly comprehensive apart from handling updates, where it is better (or the only available option in some cases) to use VMM spoofing (without "Board ID Only" Spoofing AKA `Hybridisation`) with SecureBootModel **_inactive_** in some cases, typically Catalina/Older ... remember some of us are on MacPro3,1, while fiddling with FirmwareFeatures (along with Hybridisation) and SecureBootModel **_active_** is possible, some would argue better, for BigSur/Newer. 

Also, features should not be limited to MacPro and I have no idea what the users of other units, Macs or UEFI-PC or Other, might want to variably have on or off.

> Did I miss a reason why multiple OC instances with different sets of boot options are needed to cover this case?

Again mostly covered from what I can think of. There are however always going to be things that are user specific. For instance, one might want to adjust the APFS `MinDate` (and the other related config item) to match what is being booted. So an High Sierra targeted instance may want to ensure old APFS drivers can be loaded. I can't spell out every possible use case every possible user may have though.
 
> What are the particular configuration differences between different macOS versions for hardware accelerations on MP? A short list of keys/values would do.

I believe this involves simply changing the Board ID and leaving everything else unchanged. Typically iMacPro1,1 or MacPro7,1 depending on the target. Bear in mind however that there are different approaches to OpenCore configuration.

One involves `Minimisation` where very few things are changed. This is popular with MacPro users. The main signature is that the `Automatic` setting is off. I use this myself but it would be a mistake to assume that all MacPro users use this though. I believe the OCLP calls this `Minimal Spoofing`

At the other end of the spectrum is the `Automatic` one and IIRC, the OCLP has something that falls in between. In other words, there is a wide range of settings to consider. I personally only know the `minimalised` approach; where acceleration is achieved by the so called `Hybridisation`. Not sure how the more heavily amended configs actually work.

 > Feels like all unsupported OS options can be disabled with the existing functionality and be then "BIOS"-booted.

I don't think "BIOS" booting is an option on real Macs though.
Unless you mean via the native picker but that is not always available.

vit9696:
Now that we are mostly settled with the GOP output bugs, perhaps it is about the time to return to this issue.

I suggest to introduce `Misc` → `Boot` → `InstanceIdentifier` string property, which should be passed to the picker and then used for boot visibility selection instead of the device path. Other design decisions should probably stay.

@dakanji, @mhaeuser, @mikebeaton, any objections?

mikebeaton:
> I suggest to introduce Misc → Boot → InstanceIdentifier string property, which should be passed to the picker and then used for boot visibility selection instead of the device path.

Lgtm.

> Other design decisions should probably stay.

Was not entirely sure of reference, but I assumed you meant "other design decisions as suggested in the PR should stay"?

Just quickly re-reviewing the contents of the PR, I wondered if the format of the files could be updated slightly, from `Disabled@OC_A@OC_D@` to `Disabled:OC_A,OC_D`? I think it clearer to read and more standard (I don't mean any official standard, but just more like how things are often done, and people are used to).

vit9696:
Agree with the suggested format change. Feels more intuitive.

@dakanji, could you update the PR please?

dakanji:
The requested changes are non-trivial (for me at least) and with an ingrained aversion to last minute non-trivial changes, it seems this is best left for the version after the imminent one. Not to mention that I will simply not have the time for some days in any case.

Please see items outlined below in the interim:

**ITEM 01:** I am not clear on what `Misc → Boot → InstanceIdentifier` would actually contain. Is it a random string as in something a user is free to generate without constraint that could then be used to id that instance later in the way the folder name is being used as things stand? Interesting if so but does that not open up possibilities for collisions? In other words, what exactly does this bring to the table or how does this make things better that just grabbing the folder name (which is guaranteed to be unique within a setup)?

**ITEM 02:** The decision to incorporate `@` was fairly thought out before adoption but firstly, it is important to state that the actual character could just as well have been `%` or `~` or `#` or any other special character as the role is as a boundary demarcating the target string. This makes it possible to determine whether the ContentVis file is qualified or not by the presence/absence of the character and proceed accordingly right off the bat. Then, given that an instance name could be anything and that a particular naming format is not imposed, it makes it possible to search for a match with a 100% guarantee that there is no collision ... more so since the folder name is already unique. So `@led@` or `~led~` (or some other if preferred ... Seems   `~` might be better than  `@`) can be _trivially_ and 100% accurately matched on the ContentVis content without ever colliding with something else.

**ITEM 03:** For the avoidance of doubt, I am struggling to see the point of the suggested changes but if I am off base and can't see the aim due to my limitations, I believe it is possible for one of the pantheon to push a change into the PR to implement the vision. I presume this would be quite trivial and would be fine with me. Otherwise, will try to tackle this later after understanding what I am supposed to be doing and how to implement.

mikebeaton:
You correctly described how Misc → Boot → InstanceIdentifier should work.

In my experience, if Vit suggests changes, they invariably make things better.

I believe directory name is not unique either, you are (I think) assuming that OC can only be found in subdirectories of one ESP per system, but that is not the case.

Despite the necessarily arcane nature of many of the options in OpenCore, you will be surprised to learn (I jest, more or less) that even there, the standard tradeoff is made between ease of programming and ease of use. My suggested change to the format emphasizes a reasonable gain in the latter at the expense of a small and not unreasonable cost in the former (at least, that's how I would see it).

I would imagine Vit would be happy to agree that yes, let's save this for the next release, and yes, it would be very much appreciated if you could find the time to do a first pass at the suggestions, in due course.

vit9696:
Yeah, surely we are not merging this before 0.9.2 is out. For 0.9.3 it should be doable in a calm pace and testable.

dakanji:
> You correctly described how Misc → Boot → InstanceIdentifier should work.

Good to know. Best to be clear  than assume.

> In my experience, if Vit suggests changes, they invariably make things better.

True but in my experience, he is not too mighty to change or amend his suggestions based on other input

> I believe directory name is not unique either, you are (I think) assuming that OC can only be found in subdirectories of one ESP per system, but that is not the case.

I think it is effectively unique but as said, not a hill to die for.

>  My suggested change to the format emphasizes a reasonable gain in the latter at the expense of a small and not unreasonable cost in the former (at least, that's how I would see it).

I did get that but wanted to raise why that approach was selected to that it could be considered. As said, a small cost for cost for a prodigy does not necessarily transfer to others. Right now I am still thinking the bounded strings have a certain merit that is not minor and that it is not particularly difficult for a user to grasp and implement.

mikebeaton:
@dakanji There is code in the LogModule stuff which I was just modifying (recent commits) to split a string at a comma separator, then check whether something matches any of the items. I wrote it for OpenLinuxBoot, then @PMheart re-used it for the LogModules code - easiest for yourself, and good fit with existing codebase, I think, would be to just look at the LogModules code and do that part the same way. HTH! ;)

dakanji:
Having an existing function does potentially make it easier as opposed to creating a whole infrastructure for handling comma delimited strings.

Would still need to determine whether qualified or not, I suppose `:` could serve, then split out the comma delimited portion, then process this. Sounds like an unnecessary PITA but NVM, will look later.

Thanks

mikebeaton:
> Having an existing function does potentially make it easier as opposed to creating a whole infrastructure for handling comma delimited strings.
> 
> Would still need to determine whether qualified or not, I suppose `:` could serve, then split out the comma delimited portion, then process this. Sounds like an unnecessary PITA but NVM, will look later.
> 
> Thanks

I just had to gets to grips with and then update the entire console renderer code, because my recent commits were (rightly) considered not quite right for the rest of the project. It happens... ;-)

dakanji:
Putting on hold until an implementation in progress and liked by @vit9696 is wrapped up.

dakanji:
Completed and tested as working.
- Implements `Misc → Boot → InstanceIdentifier` based on work done in [#473](https://github.com/acidanthera/OpenCorePkg/pull/473).
- Does not implement `Disabled:OC_A,OC_D` but `Disabled~OC_A~OC_D~`. I.E., drops `@`, which is not so readable and too similar to email addresses, for `~`.

On the latter, I _think_ the bounded strings have certain advantages and that the advantage of intuitiveness of the comma delimited instances (which is understood why is said) may be overstated since a user still needs to go somewhere to find out what format is to be used and will not come up with it themselves. I get the reasoning though.

Once decided that is not as big an item, the bounded items have the following advantages of straightforward and easy to maintain code and being resilient to misconfiguration.

These all work:
`Disabled~OC_A~OC_D~`
`Disabled ~OC_A~OC_D~`
`Disabled~OC_A~~OC_D~`
`Disabled~OC_A~   ~OC_D~`
`Disabled   ~OC_A~   ~OC_D~`
`Disabled   ~OC_A~ Some Random Gibberish  ~OC_D~`

Not tested but certain this would as well:
```
`Disabled
Some Random Gibberish
~OC_A~ 
More Random Gibberish
~OC_D~
```

dakanji:
Added a commit to handle and ensure legacy compatibility
- **Commit 01**: Handles things as described to date with a weakness in that if an older version of OpenCore is used, such as running an older known working instance, the visibility instruction is applied without discrimination and without considering that they are qualified. 
- **Commit 02**: Deprecates the current instructions for new ones as follows: `Disabled → Ignore` and `Auxiliary → Cloak`. ContentVis files with the current instructions are handled as before but older versions of OpenCore will see the new instructions as invalid, skip them and not apply the instructions; which is a desired outcome.
  - Not documented is that qualifying the current instructions will work fine with this arrangement. That is, `Disabled ~OC_A~OC_D~` actually works like `Ignored ~OC_A~OC_D~` with this commit in a build but is not documented and discouraged since it have the issue outlined.
  - See [issuecomment-1596256092](https://github.com/acidanthera/OpenCorePkg/pull/473#issuecomment-1596256092) (and following) for more information

Commit 01 can stand alone if deprecation is not wanted but I think it is important to handle legacy compatibility.

mikebeaton:
@dakanji -
 - I do not see any motivation for the legacy compatibility feature. It is not that I do not "think it is important to handle legacy compatibility", as a general principle - I do. It is that I do not see the motivation for it, here.
   - In [this related commit](https://github.com/acidanthera/OpenCorePkg/tree/7b7d6ea7569cd7f18d44fe163e85587184e65519), neither the old nor the new way of hiding is understood by the other
   - Surely it is reasonable to expect any user to update all their OpenCore instances in the case where they care about mutual visibility, and that very feature has just changed?
   - I cannot see why you would prefer that old instances always ignore new-style .contentVisibility (therefore always showing such items), over the 'default' behaviour which more or less comes for free (i.e. unless we work to prevent it - in both your **01** and my implementation) of obeying these as if it they were unqualified (therefore always hiding such items)? Tbh, if you _really_ want to make a new-style .contentVisibility which is ignored by old instances, just make it longer than 64 bytes, padded with junk, you can already do this!

Equally, if you are happy to accept `Disabled:OCA,ABC` (with only a single line processed), even though I know you prefer `Disabled@OCA@OCB@` (with perhaps multiple lines processed), then perhaps we can move ahead and just merge 'my' version of the feature you've asked for, and you can start using it?

dakanji:
A bit under pressure with some deadlines, so just grabbing a moment and might not be able to follow through but:

> In [this related commit](https://github.com/acidanthera/OpenCorePkg/tree/7b7d6ea7569cd7f18d44fe163e85587184e65519), neither the old nor the new way of hiding is understood by the other

The difference here is that an older version does not simply not understand things as they stand (and do nothing) but that they misunderstand them and then they act in an undesirable way. What I have done is make sure they in fact do not understand things and do nothing ... just as was the situation with the referenced related commit.

> Surely it is reasonable to expect any user to update all their OpenCore instances in the case where they care about mutual visibility, and that very feature has just changed?

It is not completely unreasonable but at the same time it could be better. There are many reasons why one might boot into an older setup and as things stand, if this is used, such could behave wrongly. 

> I cannot see why you would prefer that old instances always ignore new-style .contentVisibility (therefore always showing such items), over the 'default' behaviour which more or less comes for free (i.e. unless we work to prevent it - in both your 01 and my implementation) of obeying these as if it they were unqualified (therefore always hiding such items)?

Always hiding anything that is not specifically intended to be hidden is so bad that I suprised that it is being advocated. So if I qualify every Mac OS instance I have so that they only show up on OpenCore instances that can boot them. Then for whatever reason, I need to load an older known working instance, do you mean to say it is better for this to hide every single instance or show them all? I think the former is a flawed approach.

> Tbh, if you really want to make a new-style .contentVisibility which is ignored by old instances, just make it longer than 64 bytes, padded with junk, you can already do this!

I would rather not have to add junk to stuff when something can be made to work without such ... trivially as already done. Not such what the objection is.

It might be better to let someone not so attached take a dispassionate look at things and make a call (I believe you did ask for such from @vit9696 ... not sure if done).

I was expecting him to take another at this as well but if you are in a rush to merge, then of course you can proceed as you please.

mikebeaton:
>  So if I qualify every Mac OS instance I have so that they only show up on OpenCore instances that can boot them. Then for whatever reason, I need to load an older known working instance, do you mean to say it is better for this to hide every single instance or show them all?

Thank you for the specific example. I still think the current approach _may_ be overcomplicated, but I now understand why you are suggesting this.

mikebeaton:
How about "OCA,OCB:Disabled" for the new, qualified .contentVisibility in https://github.com/acidanthera/OpenCorePkg/pull/473. This will result in old instances of OpenCore still showing the entry, which is a good idea for the (good) reason you give.

mikebeaton:
@dakanji  - Hi - okay, I have [implemented that](https://github.com/acidanthera/OpenCorePkg/pull/473#issuecomment-1599382686).

You'll note that each thing you've asked for - with clear examples as to why it would be a good idea - including this whole feature itself! - I have implemented. And you'll also note that it was not just myself asking for specific examples (to motivate and clarify what was needed), but even @mhaeuser and @vit9696, in earlier discussions in this very thread.

While I hope it's not treading on your toes too much to (re)implement the feature, you have remained steadfast in not preferring a comma separated list (even though @vit9696 immediately agreed it was preferable) and indeed in stating that you'd find it a pita to implement. So I hope that - given the above - you might also see that there was an attempt to be helpful, here.

I think it would save @vit9696 some valuable time, and so am _hoping_ I can prompt you to agree that you have what you need, in https://github.com/acidanthera/OpenCorePkg/pull/473 ?

dakanji:
I suppose the new format will work in terms of the functionality but have to say it is perhaps the most unintuitive format that could be dreamt of and the only way that it can be adopted, is if intuitiveness, the apparent reason for going for a comma list is completely throw out of the window.

 We will live at the end of the day however you decide to proceed since that new format should work as said. Just think it is the least intuitive of all.

vit9696:
Answering here, because there are more messages in this PR, and I kind of hope to keep it in one place.

I have not yet checked the code of either PR, but design-wise I would rather stick with the following:

1. Keep `Disabled:ID1,ID2` format as agreed initially.
2. Ignore all compatibility issues with older OpenCore. It is not the first time it happens, e.g. we had that with bootstrap, and the design goal is to use most intuitive solution for current OpenCore and avoid any legacy in the codebase.

mikebeaton:
@dakanji Can also make his own response - but I do feel he has actually has got a point, after eventually spelling out his argument (and much as I seem to get burnt for trying to support what he says/asks for! 🤣): for someone actively working with OpenCore, it is not that unusual to want to roll back to an earlier version (e.g. when doing a bisect, to find the commit where an issue arose). And because of what is (arguably?) a minor bug in the earlier version (should it really ever have accepted "Disabledsakjdas"?), any user would have to manually alter their `.contentVisibility` files every time they wanted to do this (otherwise, everything marked for conditional hiding in newer versions would become unconditionally hidden in older versions, since https://github.com/acidanthera/OpenCorePkg/commit/7b7d6ea7569cd7f18d44fe163e85587184e65519).

dakanji:
Now that I am just a bit tired and no longer totally knackered, here is my clearly fair and balanced take:
- Disabled:ABC,XYZ aka `The Inglorious Bastard`
  - PROS
    - Uses logical format
    - Uses comma delimted format (loved up on Mount Olympus)
  - CONS
    - Breaks legacy compatibilty
    - Potential whitespace issues
      - @mikebeaton touched on some earlier
      - Poor whitespace shandling can be fixed with yet more complex code
- ABC,XYZ:Disabled aka `The Ugly Duckling`
  - PROS
    - Uses comma delimted format (loved up on Mount Olympus)
    - Maintains legacy compatibilty without changing anything else than the format
  - CONS
    - Butt ugly format
    - Potential whitespace issues
      - @mikebeaton touched on some earlier
      - Poor whitespace handling can be fixed with yet more code
- Ignore:ABC,XYZ aka `The Dodgy Compromise`
  - PROS
    - Uses logical format
    - Uses comma delimted format (loved up on Mount Olympus)
    - Keeps backward compatibility
      - Does not actually add a legacy debt ... Only extra condition in if statement takes care of backward compatibility
  - CONS
    - Crappy unintuitive format
    - Potential whitespace issues
      - @mikebeaton touched on some earlier
      - Poor whitespace handling can be fixed with yet more code
- Ignore~ABC~XYZ~ aka `The Lovely Tilde`
  - PROS
    - Easy to follow and understand
      - Extra debug stuff should not count as not critical
    - Uses a logical format
    - No whitespace issues
    - Keeps legacy compatibilty
      - Does not actually add a legacy debt ... Only extra condition in if statement takes care of backward compatibility
      - Can go in as `Disabled~ABC~XYZ~` but this breaks legacy compatibilty as the `Inglorious Bastard` does
  - CONS
    - Does not use a comma delimted format (knickers get into a twist up on Mount Olympus)


My ranking of all these wonderful options is:
1. The Lovely Tilde
2. The Ugly Duckling
3. The Dodgy Compromise
4. The Inglorious Bastard

So, for me, if not the Lovely Tilde, then the Ugly Duckling otherwise the Dodgy Compromise with the Inglorious Bastard coming last.

mhaeuser:
@dakanji I was rewriting my first response here to be nicer several times before I posted it. And won't I share my exact thoughts this time either. But let me be crystal-clear that your tone throughout this entire ticket is not appreciated at all.

As for the technical matters. Nothing is intuitive or lovely about a tilde. I've hardly ever seen anything using a tilde as a delimiter, while Comma-Separated Values are a unversally-supported file format, let alone being used in natural language. If it was not for reading the documentation or a general intuition for how CS folks think, I would have no clue what "the lovely tilde" is trying to express whatsoever.

Then, the order, or what you call "logical format". It really is the other way round. "Disabled" or "Ignore" is an attribute of the data, i.e. the object, to be processed. Why is the attribute of the object the significant thing here to appear first?
* Speaking in mathematical/CS terms, the most suited data structure to represent the idea with is a map(ping). It is clear and concise to have a map ``OC ID -> VisibilityState`` (the syntax ``[OC ID] -> VisibilityState`` is syntactic sugar in this case, as the mapping is well-defined). It can easily be addressed in O(1) via the readily-available ID. A map ``VisibilityState -> [OC ID]`` requires you to go through all visibility states, completey defeating the purpose of a map, and then to traverse an ID list in O(n). Absolutely fugly. While we obviously won't parse this into a map data structure for OC, you can use the ID key still to not even parse the state for declarations that do not include the current instance as a key in a much nicer fashion.
* Speaking in natural language terms, I already hinted at it, the attribute "visibility state" is not the significant component. Using the OC IDs as a key, the expression "ABC,XYZ:Disabled" corresponds to "ABC, XYZ: Disable this entry", a clear and logical command, addressing the subject first and the object last. "Ignore\~ABC\~XYZ" on the other hand corresponds to "This entry must be ignored by: ABC, XYZ", which inverts the logical sentence structure and yields passive voice (recommended against by virtually any language guideline).

While this sounds like nitpicking and bikeshedding, I just want to lay out the details - to me, it is plain obvious that the subject (OC) is the important component compared to the object (the entry). It almost feels like you disagree only to not agree with @mikebeaton.

I hope you appreciate that my points are based on logical deduction and universal conventions, not on personal preference.

mikebeaton:
A career in politics awaits you.

My favourites are:

 - `ABC,XYZ:Disabled`
   - Very marginally favourite, because it gets you _everything_ you want, functionally, and you suggested the feature in the first place; with - to me - quite little downside 'aesthetically', or in terms of intuitiveness.
 - `Disabled:ABC,XYZ`
   - Logical, clean.
 - `Ignore:ABC,XYZ`
   - Would rule it out, adding new words like this, just for backwards compatibility, seems a mug's game.
 - `Ignore ~ABC~XYZ~`
   - I personally find the tildes ... what was your phrase...? Anyway, never minding that, same issue as previous.

I/we _could_ add extra whitespace processing to any of the first two (three) - need to hunt around, we may already have it (in stuff written for OpenLinuxBoot). But I don't think it's worth opening up code which definitely needs fuzzing to be secure, for such a minimal increase in usability.

Vit's favourite, and firm request, is:

 - `Disabled:ABC,XYZ`
   - Preferred format; no need to introduce a less logical format or to change keywords (either of which _are_ legacy debt) for 'forwards' (or backwards) compatibility, since the project does not worry about this anyway. I believe no worries about space processing, for ~the same reasons I said.

I suspect Vit isn't likely to be convinced to change his view as to what is preferred here, so at some point in the next days, I'll roll my own commit back to the preferred format.

dakanji:
@mhaeuser ... Sorry, I didn't read all what you wrote but I think I get the drift. I do think Mike did get that I was speaking tongue in cheek.
 
@mikebeaton ... I do prefer the tilde of course but yes, the `duckling` does work as the next best thing. (from my position of course)

mikebeaton:
I think it's probably better to avoid tongue in cheek, in these discussions; it can easily be taken wrongly - even more so (but not only) with speakers of multiple different native languages on a project. (Yes, I do it a bit - and probably too much - too.)

dakanji:
> it is probably better to read all of what a senior member of the project writes when commenting on your PR, really, too.

I suppose that didn't come across well but meant I was only able to skim through at the time and saw that he was unhappy about the tone.

I have since read and noted the technical feedback part. I still don't like the format (as a preference thing) but it is not a deal breaker.

> I/we could add extra whitespace processing

This would be useful (although not absolutely critical)

dakanji:
On reflection, seems best to close this one down, take the stand up routine off the mount and await the implementation.

dakanji:
@mhaeuser ... just as an aside out of interest.

First, thanks for taking the time to set those items out. Quite interesting stuff.
Now, the tilde option is out so only thing left is just an academic interest in the intersting stuff you wrote. (if you have the time to indulge me as I appreciate it is totally OT. You will also not be afflicted with wierd writing and absolutely no tongue in cheek stuff). 

Now, with this:
> Speaking in natural language terms, I already hinted at it, the attribute "visibility state" is not the significant component. Using the OC IDs as a key, the expression "ABC,XYZ:Disabled" corresponds to "ABC, XYZ: Disable this entry", a clear and logical command, addressing the subject first and the object last. 

I see this differently but does "ABC,XYZ:Disabled" have equal or higher "ranking" in your view as "Disabled: ABC,XYZ"? To me, from a natural language perspective, that the latter does but just curious about other perspective. To me, "Disabled: ABC,XYZ" stands for "Instruction: Detail" which I find more logical. (probably should have been "Disable: XYZ" not "Disabled: XYZ")

> "Ignore~ABC~XYZ" on the other hand corresponds to "This entry must be ignored by: ABC, XYZ", which inverts the logical sentence structure and yields passive voice (recommended against by virtually any language guideline).

Firstly, being clear there is no attempt to push the tilde thing but this is actually the same "Instruction: Detail" format and just dispenses with the colon. The actual format was actually "Ignore ~ABC~XYZ~".  That is, it was not different to "Disabled: ABC,XYZ". It was was able to work without that space but the documented format had a space.

Just an academic curiousity as understanding that others may see things differently without just being obtuse, as I have at times wrongly thought @mikebeaton was being  or as you wrongly think I have been towards him.

mhaeuser:
> To me, "Disabled: ABC,XYZ" stands for "Instruction: Detail" which I find more logical.

This sounds like a fallacy, honestly. What follows is not "Detail" but "Subject". Even in spoken language, it is best practice to address the subject first, i.e. "Mike, please..." over "Please ..., Mike", especially when talking to a group of people. Prefixing the name easily allows people to filter what is and what is not relevant to them _before_ they get an instruction (or whatever else), much like OC can use the prefixed IDs to filter lines relevant to it without looking at the rest of the line. "Disable:OC_A" sounds like you instruct something to disable the OC install with the ID OC_A, which is not at all what is happening.

> Firstly, being clear there is no attempt to push the tilde thing but this is actually the same "Instruction: Detail" format and just dispenses with the colon. The actual format was actually "Ignore ~ABC~XYZ~". That is, it was not different to "Disabled: ABC,XYZ". It was just able to work without that space and appeared in the thread at times that way but the documented format had a space.

Sorry, I don't really understand this part.

dakanji:
I suppose 'detail` was just made up.

So I take it you would always go for ABC,XYZ:Disabled over Disabled: ABC,XYZ? I would always prefer the latter myself. 

On the unclear bit, I meant to say it was equivalent:
Disabled: ABC,XYZ
Disabled ~ABC~XYZ~

That is, basically the same way you pass parameters in many cases ... `ParamName ParamData`

mikebeaton:
@dakanji - I was given the final decision, so you have the version that gives you everything you (technically) wanted.

dakanji:
Noted. Thanks.

PR #448, opened 2023-04-15
OcAppleKernelLib: Remove Ventura check for AppleCpuPmCfgLock

jozews321:
As it says in the comment that the check can be removed if an older version can be injected correctly.

If you have an Ivy Bridge CPU or older that uses the AppleIntelCPUPowermanagent and AppleIntelCPUPowerManagementClient kexts you can inject these ones [from the OCLP project](https://github.com/dortania/OpenCore-Legacy-Patcher/tree/main/payloads/Kexts/Misc) and they will work in Ventura but only if you have the CFG lock previously disabled or else there will be a kernel panic 

The PR removes the Ventura check to allow the patch to work in the injected kexts so its no longer necessary to unlock the CFG lock. Already tested with a build and the patch is working correctly

### DIFF
--- a/Library/OcAppleKernelLib/CommonPatches.c
+++ b/Library/OcAppleKernelLib/CommonPatches.c
@@ -60,6 +60,12 @@ PatchAppleCpuPmCfgLock (
   UINT8  *WalkerEnd;
   UINT8  *WalkerTmp;
 
+  //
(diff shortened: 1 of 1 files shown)


PMheart:
Hello! Thanks for the confirmation and testing this. However, I would leave the notes, stating that AICPUPM has been removed on macOS 13, while the older version can be injected and patched, etc. Could you please handle that?

PR #449, opened 2023-04-16
OcAppleKernelLib: Lilu integration for SysKC/AuxKC injection/block

NyanCatTW1:
(Description mostly copied from https://github.com/acidanthera/OpenCorePkg/pull/385)
(See also: https://github.com/acidanthera/Lilu/pull/86)
### Motivation
In Ventura, Apple has decided to remove kext binaries from the root volume and require KDK installation to generate a new PageableKC/SysKC via kmutil.

The Auxiliary KC remains creatable without the need for a KDK installation. However, this is not viable for use cases where kexts in the SysKC must be removed or supplanted.

This issue is particularly problematic for projects like OpenCore Legacy Patcher, where automation is the key, and Apple has locked the KDK download behind Apple ID authentication.

Such a feature will solve the above problem. In addition, it opens the door to "vanilla" injection of kexts with linkage requirements for other kexts in the SysKC (e.g. IOGraphicsFamily for a hypothetical fully-featured framebuffer driver from MacHyperVSupport).

### Description

This draft achieves the feature by passing the prelinked symbols, the kexts to inject, and the kexts to block to Lilu, which does the injecting/blocking. Specifically, it allocates a buffer on the EFI runtime memory and passes the pointer to Lilu with a volatile + read-only EFI variable.

### TODOs on the OpenCore side
- Improve code style and fix memory leaks, if any
- Make the data passing between OpenCore and Lilu more stable
- Put the Lilu integration logic behind an OpenCore option
- ~~Fix building on Windows VS2019~~


Library/OcAppleKernelLib/PrelinkedContext.c
```diff
@@ -899,6 +909,129 @@ PrelinkedInjectComplete (
...
+    return Status;
+  }
+
+  return OcSetSystemVariable (
+           OC_LILU_BLOCK_INFO_ADDR_VARIABLE_NAME,
+           OPEN_CORE_NVRAM_ATTR,
+           8,
+           (void *)&Context->LiluBlockInfos,
+           &Guid
+           );
```
mhaeuser, line 1032: All this may leave a broken, partial state where N vars succeeded to be set, but at least one didn't. Using one variable, as I suggested above, should resolve that though.

Library/OcAppleKernelLib/PrelinkedContext.c
```diff
@@ -899,6 +909,129 @@ PrelinkedInjectComplete (
...
+    LengthOfPrelinkedSymbols,
+    NumSymbolsInPrelinked
+    ));
+
+  // Expose the physical address of the LILU_PRELINKED_SYMBOLS struct via an EFI variable
+  Status = OcSetSystemVariable (
+             OC_LILU_PRELINKED_SYMBOLS_ADDR_VARIABLE_NAME,
+             OPEN_CORE_NVRAM_ATTR,
+             8,
+             (void *)&LiluPrelinkedSymbolsAddr64,
```
mhaeuser, line 1005: Redundant casts, also `void` is not allowed.

Library/OcMainLib/OpenCoreKernel.c
```diff
@@ -754,6 +773,9 @@ OcKernelInjectKexts (
       );
 
     Status = PrelinkedInjectComplete (Context);
+    if (Status == EFI_SUCCESS) {
```
mhaeuser, line 776: !EFI_ERROR

Library/OcAppleKernelLib/PrelinkedContext.c
```diff
@@ -1256,6 +1389,106 @@ PrelinkedInjectKext (
...
+                              );
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  Context->LiluKextCount++;
+
+  // To make caller of the function happy
+  if (BundleVersion != NULL) {
+    BundleVersion = "1.0.0";
```
mhaeuser, line 1486: Overwriting a local variable will make neither the caller nor me happy. :)

Library/OcAppleKernelLib/PrelinkedContext.c
```diff
@@ -1256,6 +1389,106 @@ PrelinkedInjectKext (
...
+
+  Buffer = (VOID *)LiluInjectionInfoAddr;
+  if (!Buffer) {
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  InjectionInfo              = (LILU_INJECTION_INFO *)Buffer;
+  InjectionInfo->Version     = 0;
+  InjectionInfo->EntryLength = EntryLength;
+  if (AsciiStrCpyS (InjectionInfo->BundlePath, sizeof (InjectionInfo->BundlePath), BundlePath) != RETURN_SUCCESS) {
```
mhaeuser, line 1431: Function calls in if-directives make their lines quite long and personally I don't like side effects in them.

mhaeuser (changes requested):
Just some high-level things I noticed, not a functional review.

Library/OcAppleKernelLib/PrelinkedContext.c
```diff
@@ -899,6 +909,129 @@ PrelinkedInjectComplete (
...
+        Symbols++;
+      }
+    }
+
+    Link = GetNextNode (&Context->PrelinkedKexts, Link);
+  }
+
+  // Allocate buffer and setup header
+  LiluPrelinkedSymbolsAddr = (UINTN)AllocateRuntimePool (LengthOfPrelinkedSymbols);
+  Buffer                   = (VOID *)LiluPrelinkedSymbolsAddr;
```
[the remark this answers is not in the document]
NyanCatTW1: The address passed to Lilu were incorrect when I didn't do that. I should probably find a cleaner fix though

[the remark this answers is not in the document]
mhaeuser: Unless there is a bad toolchain bug (which is virtually impossible with our validation), this is impossible, as the size and alignment constraints of `UINTN` and `VOID *` are the same.

Library/OcAppleKernelLib/PrelinkedContext.c
```diff
@@ -899,6 +909,129 @@ PrelinkedInjectComplete (
...
+
+        CurEntry = (LILU_PRELINKED_SYMBOLS_ENTRY *)(((UINT8 *)CurEntry) + CurEntry->EntryLength);
+        Symbols++;
+      }
+    }
+
+    Link = GetNextNode (&Context->PrelinkedKexts, Link);
+  }
+
+  LiluPrelinkedSymbolsAddr64 = (UINT64)LiluPrelinkedSymbolsAddr;
```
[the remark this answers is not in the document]
NyanCatTW1: I created that code path to fix 32-bit building. There should be a better way to do this though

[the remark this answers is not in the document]
mhaeuser: You cannot cast from `VOID *` to `UINT64` immediately due to the size difference, but you can cast to `UINTN`, which will implicitly convert to `UINT64`.

Library/OcAppleKernelLib/PrelinkedContext.c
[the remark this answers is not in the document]
NyanCatTW1: You mean:
```c
  Buffer                   = AllocateRuntimePool (LengthOfPrelinkedSymbols);
  LiluPrelinkedSymbolsAddr = (UINTN)Buffer;
```
?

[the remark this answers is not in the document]
mhaeuser: Yes

Include/Acidanthera/Guid/OcVariable.h
```diff
@@ -94,6 +94,21 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
...
+
+//
+// Variable used for passing the amount of kexts to inject to Lilu
+//
+#define OC_LILU_KEXT_COUNT_VARIABLE_NAME  L"lilu-kext-count"
+
+//
+// Variable used for passing the list of kexts to block/exclude to Lilu
+//
+#define OC_LILU_BLOCK_INFO_ADDR_VARIABLE_NAME  L"lilu-block-info-addr"
```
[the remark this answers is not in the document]
NyanCatTW1: The issue with that is that it makes determining the pool allocation size more difficult, and I am not sure if the benefits of having it all in a single struct would compensate that

Include/Acidanthera/Library/OcAppleKernelLib.h
```diff
@@ -167,6 +167,148 @@ typedef enum KERNEL_CACHE_TYPE_ {
...
+  //
+  UINT64    SymbolValue;
+  //
+  // Length of this symbol's name
+  //
+  UINT32    SymbolNameLength;
+  //
+  // This symbols's name
+  //
+  CHAR8     SymbolName[1];
```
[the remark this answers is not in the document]
NyanCatTW1: Using that caused the Windows build to fail. Same with [0].
It complains that:
```
 d:\a\opencorepkg\opencorepkg\udk\OpenCorePkg\Include\Acidanthera\Library/OcAppleKernelLib.h(219): error C2233: 'Entries': arrays of objects containing zero-size arrays are illegal
```

Library/OcAppleKernelLib/PrelinkedContext.c
```diff
@@ -1328,3 +1561,36 @@ PrelinkedContextBlock (
...
+  IN     BOOLEAN            Exclude,
+  IN     UINT8              KCKind
+  )
+{
+  LILU_BLOCK_INFO_ENTRY  *CurEntry;
+
+  ASSERT (Context != NULL);
+  ASSERT (Identifier != NULL);
+
+  if (Context->LiluBlockInfos->Header.Size + sizeof (LILU_BLOCK_INFO_ENTRY) > LILU_BLOCK_INFO_SIZE_LIMIT_VERSION_0) {
```
mhaeuser, line 1578: While an overflow here is highly unlikely, moving `sizeof (LILU_BLOCK_INFO_ENTRY)` to the right side would mitigate the (low) risk.

Include/Acidanthera/Guid/OcVariable.h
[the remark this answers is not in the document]
mhaeuser: What? The other two variables get two integers, so it's a constant-size addition. I think you understood it as "make one flat buffer"? That is not what I meant, I literally just mean, expose those three data as part of the same variable (with a struct definition).

Include/Acidanthera/Library/OcAppleKernelLib.h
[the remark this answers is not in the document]
mhaeuser: Zero-size is UB. What error do you get with a flexible array?

Include/Acidanthera/Guid/OcVariable.h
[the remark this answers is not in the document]
NyanCatTW1: Oh. I get what you mean now

Include/Acidanthera/Library/OcAppleKernelLib.h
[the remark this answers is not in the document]
NyanCatTW1: Same as that. See https://github.com/NootInc/OpenCorePkg/actions/runs/4711158291/jobs/8355356419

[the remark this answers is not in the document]
mhaeuser: For sane usages, that is not possible, we use them in many places: https://github.com/acidanthera/OpenCorePkg/blob/f4ca7b10dcab1b9b470719d445ac0067efd4b3d4/Include/Acidanthera/IndustryStandard/PeCoffImage.h#L192

Your issue is you have those nested. I.e., `LILU_PRELINKED_SYMBOLS_ENTRY` already has a FAM, so you cannot make an array out of that. Trying to declare them as such by forcing FAMs to be 1-sized is dangerous and can cause hidden bugs that look valid during code review. Imo, just drop the nested array member and address manually.

Library/OcAppleKernelLib/PrelinkedContext.c
ChefKissInc:
```c
  Status = AsciiStrCpyS (InjectionInfo->BundlePath, sizeof (InjectionInfo->BundlePath), BundlePath);
  if (EFI_ERROR(Status)) {
    FreePool (Buffer);
    return Status;
  }
```
Is this how it should be? @mhaeuser

mhaeuser: Yes

Library/OcAppleKernelLib/PrelinkedContext.c
ChefKissInc: > Overwriting a local variable will make neither the caller nor me happy. :)

This is doing nothing atm right @mhaeuser? Should I omit it or overwrite the contents in some way?

Library/OcAppleKernelLib/PrelinkedContext.c
ChefKissInc: > While an overflow here is highly unlikely, moving `sizeof (LILU_BLOCK_INFO_ENTRY)` to the right side would mitigate the (low) risk.

Can you elaborate? I don't get what you mean

Library/OcAppleKernelLib/PrelinkedContext.c
mhaeuser: You can omit it, doesn't really matter: https://github.com/acidanthera/OpenCorePkg/blob/0097c13e686186d876b6208a1f2851454230764f/Library/OcMainLib/OpenCoreKernel.c#L589

Library/OcAppleKernelLib/PrelinkedContext.c
mhaeuser: ``Context->LiluBlockInfos->Header.Size + sizeof (LILU_BLOCK_INFO_ENTRY) > LILU_BLOCK_INFO_SIZE_LIMIT_VERSION_0``
->
``Context->LiluBlockInfos->Header.Size > LILU_BLOCK_INFO_SIZE_LIMIT_VERSION_0 - sizeof (LILU_BLOCK_INFO_ENTRY)``

ChefKissInc: Done

Library/OcAppleKernelLib/PrelinkedContext.c
ChefKissInc: Done

Library/OcAppleKernelLib/PrelinkedContext.c
ChefKissInc: Done

Library/OcMainLib/OpenCoreKernel.c
ChefKissInc: Done

Library/OcAppleKernelLib/PrelinkedContext.c
ChefKissInc: Done

Include/Acidanthera/Library/OcAppleKernelLib.h
```diff
@@ -167,6 +167,148 @@ typedef enum KERNEL_CACHE_TYPE_ {
...
+
+typedef struct {
+  //
+  // Length of this entry
+  //
+  UINT32    EntryLength;
+  //
+  // Value of this symbol (or stab offset)
+  //
+  UINT64    SymbolValue;
```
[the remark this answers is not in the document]
NyanCatTW1: Considering that the structs are currently padded, this should be fine, I guess

Include/Acidanthera/Library/OcAppleKernelLib.h
```diff
@@ -167,6 +167,148 @@ typedef enum KERNEL_CACHE_TYPE_ {
 #define KERNEL_VERSION_BIG_SUR_MAX        (KERNEL_VERSION_MONTEREY_MIN - 1)
 #define KERNEL_VERSION_MONTEREY_MAX       (KERNEL_VERSION_VENTURA_MIN - 1)
 
+//
+// Lilu integration structs
+//
+#pragma pack(1)
```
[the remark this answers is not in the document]
NyanCatTW1: Well...I think I'll just keep it

Include/Acidanthera/Library/OcAppleKernelLib.h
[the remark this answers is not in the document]
NyanCatTW1: Will resolve this after implementing the change on the Lilu side

vit9696:
Hello, any reason we are closing this?

NyanCatTW1:
1. Visual and I disbanded (See Discord #coding for details) However, the repos were deleted without my awareness
2. The OCLP folks and I decided to make it a separate kext, instead of combining it with Lilu. This is because we are planning to move the injected kexts to the filesystem, instead of storing them on the EFI. This also means there are no longer needs to integrate with OpenCore. Henceforth, both PRs are no longer necessary, but I'd certainly enjoy reusing codes made here

PR #451, opened 2023-04-16
Update documentation for DisableIoMapperMapping kernel quirk

CaseySJ:
Further investigation with Ventura 13.3 has shown that the new `DisableIoMapperMapping` kernel quirk is needed on systems where the following conditions are present (all must be true):

1. One or more `Reserved Memory Region`s exist in System DMAR
2. More than 16GB memory installed
3. iGPU enabled

This problem wasn't reported on Z690 and Z790 because iGPU on these processors is not compatible with macOS. On Comet Lake and older Intel Core processors, this quirk is needed when all of the above conditions are true.

This pull request only updates the documentation file `Configuration.tex`.


perez987:
@CaseySJ 
Why remove the note about systems where DMAR has one or more Reserved Memory regions? Is this not also a must to need DisableIoMapperMapping?

CaseySJ:
> @CaseySJ Why remove the note about systems where DMAR has one or more Reserved Memory regions? Is this not also a "must" to need DisableIoMapperMapping?

Let me double-check myself. I was relying on a user's reply. Will update this in the next day or so.

CaseySJ:
@perez987, thank you for encouraging me to test this personally. After running my own tests, the original commit is correct and the subsequent commit is being reverted.

`Reserved Memory Regions` need to be removed. I'll update the first post in this thread.

perez987:
@CaseySJ 
You are very welcome.
The work you have done on this issue is impressive. Thoroughly studied and solution provided. Congratulations.

Docs/Configuration.tex
```diff
@@ -2665,10 +2665,10 @@ \subsection{Quirks Properties}\label{kernelpropsquirks}
 
   \emph{Note 1}: This option resolves compatibility issues with Wi-Fi, Ethernet and
   Thunderbolt devices when AppleVTD is enabled on systems where the native DMAR table
-  contains one or more Reserved Memory Regions.
+  contains one or more Reserved Memory Regions and iGPU is enabled and more than
+  16GB memory is installed.
```
PMheart, line 2669: `16 GB`; there are several instances in the documentation where there is one space between the numeric value and unit.

CaseySJ: Change has been made.

justinkb:
I needed this on Z390 with iGPU *disabled*

PR #453, opened 2023-05-01
Fix console

mikebeaton:
Apologies for earlier fat finger.

Here it is as a PR.

Commit to go to back to how everything was is now on master, as agreed. But note this causes I believe one different reversion outside of EnableGop (namely initial verbose boot text writes over graphics picker; see subsequent comment for more details), and multiple reversions inside EnableGop (which is temporarily disabled from bundling in OC as agreed).

Then, still in this PR itself:

 - First commit is meant to be non-controversial changes to BuiltinRenderer
 - Second commit implements InitialMode option
 - Third and fourth commits attempt to address all desiderata (see below)
 - Fifth commit (seemed like a good idea) supports ConsoleMode text resolution option in Builtin renderer
 - Sixth just syncs changelog

My understanding (very briefly, see also docs in the relevant commit) of requirement for InitialMode:
 - Set the initial console mode (text, graphics), if you can, but anyway report it as being set, if not
 - For the renderers, initialise them just as they were previously (before all changes), but taking the result of the above as what they think the current mode is, rather than reading the actual current mode (when these differ)

(I believe this code achieves this, but...)

**Desiderata for these changes:**

- APFS log lines must not appear before Mac logo/OC menu (or over UI theme colour) on native Macs using EnableGop, but will do so unless it is possible to start builtin renderer in graphics mode (i.e. reporting itself as graphics mode)
- There should be no flash to console bg (typically black, and in general cannot always match theme UI colour) between native picker and Canopy, but will be unless builtin renderer in EnableGop disables screen clearing in graphics mode, and new builtin renderer in OC disables clearing when staying in graphics mode
- If we were in graphics mode, or uncontrolled text mode, we want a console bg cls before macOS verbose output starts. Note that the ideal timing for this is precisely when macOS itself requests to switch to console mode
- We ideally _don't_ want a cls when verbose output starts and we have the (controlled) builtin picker text on screen
- Mode switches should be cheap or free (Vit's comment)
- Apple picker used as OC picker should switch back to builtin picker cleanly, if it needs to (i.e. not on top of the Apple picker graphics)
- builtin picker (and password prompt) must not overwrite bios logo
- Continues to behave how @dhinakg expects: with InitialMode=Text (or Auto, if this gives Text), debug lines should just run freely over Canopy graphics, as they appear, and without any additional (automatic, or other) clear screens compared to how it was before
- Enter graphics mode tool from builtin picker then return to menu should display menu cleanly, not over previous graphics output

Not sure if it is also worth adding my summary of operation of pre-existing renderers (as they always were, i.e. after reverting any recent changes (as done in first commit)):

- Builtin renderer always starts off reporting as text (regardless of whether it runs in underlying text or graphics (or 'neither', i.e. no cc protocol) (but for reasons as above, we may want it to start off reporting as graphics; particularly if it is taking over from existing graphics - which would be InitialMode=Auto)
- System renderer:
  - In Graphics or Text, we set the requested mode, and then start off reporting as being in the mode we are actually in (but defaulting to text if there is no native console control protocol) (so note that with no InitialMode and no builtin protocol, we can only start off reporting as being in text mode)
  - Generic is somewhat different; firstly, we just use the existing mode, without trying to set either mode first; but secondly, we (attempt to) get the existing console control protocol, and _if_ it exists, we pass through GetMode, SetMode and LockStdIn to it

### DIFF
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -8107,6 +8107,19 @@ for additional options.
 \begin{enumerate}
 
 \item
+  \texttt{InitialMode}\\
+  \textbf{Type}: \texttt{plist\ string}\\
+  \textbf{Failsafe}: \texttt{Auto}\\
+  \textbf{Description}: Selects initial \texttt{ConsoleControl} mode to use before applying \texttt{TextRenderer} option.
+
+  Available values are \texttt{Auto}, \texttt{Text} and \texttt{Graphics}.
+  When \texttt{Text} or \texttt{Graphics} are specified, will attempt to set the system
+  \texttt{ConsoleControl} protocol to the specified value, and will then report being in
+  specified mode even if it could not set it, e.g. if there is no system \texttt{ConsoleControl}
+  protocol. For \texttt{Auto}, reports the existing \texttt{ConsoleControl} mode, defaulting
+  to \texttt{Text} if there is no system \texttt{ConsoleControl} protocol.
+
+\item
   \texttt{TextRenderer}\\
   \textbf{Type}: \texttt{plist\ string}\\
   \textbf{Failsafe}: \texttt{BuiltinGraphics}\\
@@ -8161,7 +8174,22 @@ for additional options.
   and \texttt{ClearScreenOnModeSwitch} are more specific, and their use
   depends on the firmware.
 
-  \emph{Note}: Some Macs, such as the \texttt{MacPro5,1}, may have incompatible
+  \emph{Note 1}: As required to work round issues with some legacy firmware,
+  both \texttt{Builtin} protocols use the system \texttt{ConsoleControl} protocol
+  to set \texttt{Text} or \texttt{Graphics} mode as specified in the option name, but
+  then behave internally (e.g. as to whether text output is shown in \texttt{Text}
+  mode or suppressed in \texttt{Graphics} mode) and report to other systems as if they
+  were in the mode present \emph{before} the protocol was applied (defaulting
+  to \texttt{Text} mode when there is no system \texttt{ConsoleControl} protocol).
+  To control this internal mode value, the \texttt{InitialMode} option should be used.
+
+  \emph{Note 2}: The \texttt{System} protocols will set and report the mode
+  specified in the option name, as long as there is a native \texttt{ConsoleControl}
+  protocol. To gain control over the reported mode whether or not there is an existing
+  system \texttt{ConsoleControl} protocol, the \texttt{InitialMode} option should
+  be used.
+
+  \emph{Note 3}: Some Macs, such as the \texttt{MacPro5,1}, may have incompatible
   console output when using modern GPUs, and thus only \texttt{BuiltinGraphics}
   may work for them in such cases. NVIDIA GPUs may require additional
   \href{https://github.com/acidanthera/bugtracker/issues/1280}{firmware upgrades}.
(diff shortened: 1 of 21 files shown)


mikebeaton:
The sequence of commits relevant to the remaining reversion which is present, with the current state of master at https://github.com/acidanthera/OpenCorePkg/commit/acecdf5e7b37da9e733908f6147c7516e7ab25d4, are:

 - https://github.com/acidanthera/OpenCorePkg/commit/2439d58412c071cb735cde8a70c9ef964a095780 - Resolved @Goldfish64's reversion [here](https://github.com/acidanthera/OpenCorePkg/commit/2439d58412c071cb735cde8a70c9ef964a095780#diff-3d0ce29384d99e6f2b0583f6a31b90f9bd40b3c9cfc71cf53d2dc2e2739c59e8R911-R917) (but not in a nice way, taking account of all the other desiderata, unlike the new PR)
 - https://github.com/acidanthera/OpenCorePkg/commit/13356ae4a4ed11db571ff34600790ace564559b4 - OcBootManagementLib: Fix unwanted flicker when launching non-text entry (introduced Fish's reversion)
 - https://github.com/acidanthera/OpenCorePkg/commit/eb5cc3ee62c06f111931a5e20ed584085ac28d1f - Just the change in OcBootManagementLib.c (part of a massive commit, sorry) - fixed what it says, but clearly (now) reverts the original black flash fix (unlike new PR, which resolves all of the various issues at once)
 - https://github.com/acidanthera/OpenCorePkg/commit/76e889f89af7ecdec0a8de191d7ded5856b338fb - OpenCanopy: Fix black splash with non-black bg

Docs/Configuration.tex
```diff
@@ -8106,6 +8106,19 @@ \subsection{Output Properties}\label{uefioutputprops}
 
 \begin{enumerate}
 
+\item
+  \texttt{InitialMode}\\
+  \textbf{Type}: \texttt{plist\ string}\\
+  \textbf{Failsafe}: \texttt{Auto}\\
+  \textbf{Description}: Selects initial \texttt{ConsoleControl} mode to use before applying \texttt{TextRenderer} option.
```
vit9696, line 8113: This description is cryptic. `Start text renderer in the specified mode` may be better.

Docs/Configuration.tex
```diff
@@ -8106,6 +8106,19 @@ \subsection{Output Properties}\label{uefioutputprops}
...
+
+  Available values are \texttt{Auto}, \texttt{Text} and \texttt{Graphics}.
```
vit9696, line 8115: In my opinion, focus on what the mode is in the first place. Something like `In text mode both text and graphics can be rendered onscreen. In graphics mode only graphics can be. Not all tools can switch to text mode if they want to print text.`

Library/OcConsoleLib/TextOutputBuiltin.c
```diff
@@ -382,26 +397,43 @@ RenderResync (
   }
 
   mConsoleGopMode = mGraphicsOutput->Mode->Mode;
-  mConsoleWidth   = (Info->HorizontalResolution / TGT_CHAR_WIDTH)  - 2 * SCR_PADD;
-  mConsoleHeight  = (Info->VerticalResolution   / TGT_CHAR_HEIGHT) - 2 * SCR_PADD;
-  mConsoleMaxPosX = 0;
-  mConsoleMaxPosY = 0;
+  MaxWidth        = Info->HorizontalResolution / TGT_CHAR_WIDTH;
+  MaxHeight       = Info->VerticalResolution / TGT_CHAR_HEIGHT;
+  if ((mUserWidth < 1) || (mUserHeight < 1)) {
```
vit9696, line 402: should be == 0, made me scratch my head for a short while.

Staging/EnableGop/README.md
```diff
@@ -3,6 +3,11 @@
 ## Releases
 
 EnableGop version (OpenCore version)
+
+### 1.4 (0.9.3)
+ - Incorporates recent updates to OpenCore console control code, but no difference in behaviour compared
+   to version 1.3 is expect on any supported systems.
```
vit9696, line 9: is expected?

vit9696 (commented):
I think this finally looks good to me. Thank you so much for spending your time on it :)

PR #455, opened 2023-05-06
EnableGop: Provide release and pre-release versions with build

mikebeaton:
Here is a PR for having a stable release and a current bleeding edge EnableGop both packaged by OC build.

The resulting directory structure within the packaged Utilities/EnableGop directory is:

```
$ ls -RF
EnableGopDirect_1.3.efi           Pre-release/
EnableGopDirect_1.3.ffs           README.md
EnableGop_1.3.efi                 UEFITool_Inserted_Screenshot.png
EnableGop_1.3.ffs                 vBiosInsert.sh*

./Pre-release:
EnableGopDirect_1.4-dev.efi  EnableGop_1.4-dev.efi
EnableGopDirect_1.4-dev.ffs  EnableGop_1.4-dev.ffs
```

The top level directory is the same layout as before, except for the version number on some filenames, and those versioned files are now taken from the stable Staging/EnableGop/Release directory in OC source.

The Pre-release directory is new and contains the current, bleeding edge built versions (the same files which would previously have been put in the top level directory, but now with an automatically inserted version number added to each filename).

As an additional improvement, it will now intentionally stop the build if the version strings in EnableGop.inf and EnableGopDirect.inf are different.


vit9696:
Looks good to me. Please merge.

PR #459, opened 2023-05-08
4G Decoding CpuIo/PciRootBridgeIo/PciIo fix for Aptio 4 / Ivy Bridge

xCuri0:
This UEFI has several bugs relating to 4G Decoding which need to be fixed
* PciRootBridgeIo Mem can't access above 4GB MMIO (Done)
* CpuIo can't access above 4GB MMIO (Done)
* PciIo can't access extended configuration. ```ResizeUsePciRbIo``` works around this for resizable BAR

Boot audio now plays on my system with this patch. Loading time is kinda slow configuring the amp but seems similar to original PciIo.

### DIFF
--- a/Include/Acidanthera/Library/OcPciIoLib.h
+++ b/Include/Acidanthera/Library/OcPciIoLib.h
@@ -13,7 +13,7 @@
 #ifndef OC_PCI_IO_LIB_H
 #define OC_PCI_IO_LIB_H
 
-#include <Protocol/PciIo.h>
+#include <Protocol/CpuIo2.h>
 
 /**
   The user Entry Point for PciIo.
@@ -25,7 +25,7 @@
   @returns Installed protocol.
   @retval NULL  There was an error installing the protocol.
 **/
-EFI_PCI_IO_PROTOCOL *
+EFI_CPU_IO2_PROTOCOL *
 OcPciIoInstallProtocol (
   IN BOOLEAN  Reinstall
   );
--- a/Include/Acidanthera/Library/OcPciIoLib.h
+++ b/Include/Acidanthera/Library/OcPciIoLib.h
@@ -1,5 +1,7 @@
 /** @file
-  Copyright (c) 2006 - 2011, Intel Corporation. All rights reserved.<BR>
+  Copyright (C) 2023, xCuri0. All rights reserved.
+
+  All rights reserved.
 
   This program and the accompanying materials
   are licensed and made available under the terms and conditions of the BSD License
--- a/Include/Acidanthera/Library/OcPciIoLib.h
+++ b/Include/Acidanthera/Library/OcPciIoLib.h
@@ -1,0 +1,33 @@
+/** @file
+  Copyright (c) 2006 - 2011, Intel Corporation. All rights reserved.<BR>
+
+  This program and the accompanying materials
+  are licensed and made available under the terms and conditions of the BSD License
+  which accompanies this distribution.  The full text of the license may be found at
+  http://opensource.org/licenses/bsd-license.php
+
+  THE PROGRAM IS DISTRIBUTED UNDER THE BSD LICENSE ON AN "AS IS" BASIS,
+  WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
+**/
+
+#ifndef OC_PCI_IO_LIB_H
+#define OC_PCI_IO_LIB_H
+
+#include <Protocol/PciIo.h>
+
+/**
+  The user Entry Point for PciIo.
+
+  This function fixes PciIo related protocols.
+
+  @param[in] Reinstall  Replace any installed protocol.
+
+  @returns Installed protocol.
+  @retval NULL  There was an error installing the protocol.
+**/
+EFI_PCI_IO_PROTOCOL *
+OcPciIoInstallProtocol (
+  IN BOOLEAN  Reinstall
+  );
+
+#endif // OC_PCI_IO_LIB_H
--- a/Library/OcPciIoLib/OcPciIoLib.c
+++ b/Library/OcPciIoLib/OcPciIoLib.c
@@ -27,7 +27,8 @@ OcPciIoInstallProtocol (
     return NULL;
   }
 
-  if (Reinstall) {
+  DEBUG ((DEBUG_INFO, "OCPIO: Reinstall %d\n", Reinstall));
+  if (!Reinstall) {
     return mCpuIo;
   } else {
     Tpl    = gBS->RaiseTPL (TPL_HIGH_LEVEL);
--- a/Library/OcPciIoLib/OcPciIoLib.c
+++ b/Library/OcPciIoLib/OcPciIoLib.c
@@ -14,29 +14,23 @@ OcPciIoInstallProtocol (
   IN BOOLEAN  Reinstall
   )
 {
-  EFI_TPL                          tpl;
+  EFI_TPL                          Tpl;
   UINTN                            Index;
-  EFI_STATUS                       Status  = 0;
-  EFI_CPU_IO2_PROTOCOL             *mCpuIo = NULL;
+  EFI_STATUS                       Status;
+  EFI_CPU_IO2_PROTOCOL             *mCpuIo;
   UINTN                            HandleCount;
   EFI_HANDLE                       *HandleBuffer;
   EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL  *PciRootBridgeIo;
 
-  if (Reinstall) {
+  mCpuIo = InitializeCpuIo2 ();
+  if (!mCpuIo) {
     return NULL;
+  }
+
+  if (Reinstall) {
+    return mCpuIo;
   } else {
-    mCpuIo = InitializeCpuIo2 ();
-
-    tpl = gBS->RaiseTPL (TPL_HIGH_LEVEL);
-    if (!mCpuIo) {
-      return NULL;
-    }
-
-    DEBUG ((DEBUG_INFO, "OCPIO: Fixing CpuIo2\n"));
-    // Override with 64-bit MMIO compatible functions
-    mCpuIo->Mem.Read  = CpuMemoryServiceRead;
-    mCpuIo->Mem.Write = CpuMemoryServiceWrite;
-
+    Tpl = gBS->RaiseTPL (TPL_HIGH_LEVEL);
     Status = gBS->LocateHandleBuffer (
                     ByProtocol,
                     &gEfiPciRootBridgeIoProtocolGuid,
@@ -50,8 +44,14 @@ OcPciIoInstallProtocol (
       goto Free;
     }
 
+    DEBUG ((DEBUG_INFO, "OCPIO: Fixing CpuIo2\n"));
+    // Override with 64-bit MMIO compatible functions
+    mCpuIo->Mem.Read  = CpuMemoryServiceRead;
+    mCpuIo->Mem.Write = CpuMemoryServiceWrite;
+
     for (Index = 0; Index < HandleCount; ++Index) {
       DEBUG ((DEBUG_INFO, "OCPIO: Fixing PciRootBridgeIo %d\n", Index));
+
       Status = gBS->HandleProtocol (
                       HandleBuffer[Index],
                       &gEfiPciRootBridgeIoProtocolGuid,
@@ -61,16 +61,14 @@ OcPciIoInstallProtocol (
         mCpuIo = NULL;
         goto Free;
       }
-
       // Override with 64-bit MMIO compatible functions
       PciRootBridgeIo->Mem.Read  = RootBridgeIoMemRead;
       PciRootBridgeIo->Mem.Write = RootBridgeIoMemWrite;
     }
-
-    gBS->RestoreTPL (tpl);
   }
 
 Free:
+  gBS->RestoreTPL (Tpl);
   FreePool (HandleBuffer);
   return mCpuIo;
 }
--- a/Library/OcPciIoLib/OcPciIoLib.c
+++ b/Library/OcPciIoLib/OcPciIoLib.c
@@ -9,32 +9,30 @@
 #include "OcPciIoU.h"
 #include <Library/OcPciIoLib.h>
 
-EFI_PCI_IO_PROTOCOL *
+EFI_CPU_IO2_PROTOCOL *
 OcPciIoInstallProtocol (
   IN BOOLEAN  Reinstall
   )
 {
   EFI_TPL                          tpl;
-  EFI_STATUS                       Status = 0;
-  EFI_CPU_IO2_PROTOCOL             *mCpuIo;
+  UINTN                            Index;
+  EFI_STATUS                       Status  = 0;
+  EFI_CPU_IO2_PROTOCOL             *mCpuIo = NULL;
   UINTN                            HandleCount;
   EFI_HANDLE                       *HandleBuffer;
   EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL  *PciRootBridgeIo;
-  UINTN                            Index;
 
   if (Reinstall) {
-    // TODO: if we need it
+    return NULL;
   } else {
-    DEBUG ((DEBUG_INFO, "OCPIO: Init CpuIo2\n"));
     mCpuIo = InitializeCpuIo2 ();
-    DEBUG ((DEBUG_INFO, "OCPIO: Fixing CpuIo2\n"));
 
     tpl = gBS->RaiseTPL (TPL_HIGH_LEVEL);
-
-    if (EFI_ERROR (Status)) {
+    if (!mCpuIo) {
       return NULL;
     }
 
+    DEBUG ((DEBUG_INFO, "OCPIO: Fixing CpuIo2\n"));
     // Override with 64-bit MMIO compatible functions
     mCpuIo->Mem.Read  = CpuMemoryServiceRead;
     mCpuIo->Mem.Write = CpuMemoryServiceWrite;
@@ -47,6 +45,11 @@ OcPciIoInstallProtocol (
                     &HandleBuffer
                     );
 
+    if (EFI_ERROR (Status)) {
+      mCpuIo = NULL;
+      goto Free;
+    }
+
     for (Index = 0; Index < HandleCount; ++Index) {
       DEBUG ((DEBUG_INFO, "OCPIO: Fixing PciRootBridgeIo %d\n", Index));
       Status = gBS->HandleProtocol (
@@ -55,7 +58,8 @@ OcPciIoInstallProtocol (
                       (VOID **)&PciRootBridgeIo
                       );
       if (EFI_ERROR (Status)) {
-        continue;
+        mCpuIo = NULL;
+        goto Free;
       }
 
       // Override with 64-bit MMIO compatible functions
@@ -63,12 +67,10 @@ OcPciIoInstallProtocol (
       PciRootBridgeIo->Mem.Write = RootBridgeIoMemWrite;
     }
 
-    FreePool (HandleBuffer);
-
     gBS->RestoreTPL (tpl);
-
-    return NULL;
   }
 
-  return NULL;
+Free:
+  FreePool (HandleBuffer);
+  return mCpuIo;
 }
--- a/Library/OcPciIoLib/OcPciIoLib.c
+++ b/Library/OcPciIoLib/OcPciIoLib.c
@@ -23,7 +23,7 @@ OcPciIoInstallProtocol (
   EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL  *PciRootBridgeIo;
 
   CpuIo = InitializeCpuIo2 ();
-  if (!CpuIo) {
+  if (CpuIo == NULL) {
     return NULL;
   }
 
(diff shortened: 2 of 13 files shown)


xCuri0:
Should I also add a fix for PciIo extended capabilities ? I'm thinking of not doing it because we already have ```ResizeUsePciRbIo``` and patching ~20 devices will add to boot time.

Library/OcPciIoLib/OcPciIoLib.c
```diff
@@ -0,0 +1,76 @@
...
+#include <Library/OcPciIoLib.h>
+
+EFI_CPU_IO2_PROTOCOL *
+OcPciIoInstallProtocol (
+  IN BOOLEAN  Reinstall
+  )
+{
+  EFI_TPL                          tpl;
+  UINTN                            Index;
+  EFI_STATUS                       Status  = 0;
```
mhaeuser, line 19: 0 is not an EFI_STATUS value, also declaration and initialisation must be separate.

Library/OcPciIoLib/OcPciIoLib.c
```diff
@@ -0,0 +1,76 @@
...
+  EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL  *PciRootBridgeIo;
+
+  if (Reinstall) {
+    return NULL;
+  } else {
+    mCpuIo = InitializeCpuIo2 ();
+
+    tpl = gBS->RaiseTPL (TPL_HIGH_LEVEL);
+    if (!mCpuIo) {
+      return NULL;
```
mhaeuser, line 32: High TPL must not escape the function scope!

Library/OcPciIoLib/OcPciIoLib.c
```diff
@@ -0,0 +1,76 @@
...
+  EFI_TPL                          tpl;
+  UINTN                            Index;
+  EFI_STATUS                       Status  = 0;
+  EFI_CPU_IO2_PROTOCOL             *mCpuIo = NULL;
+  UINTN                            HandleCount;
+  EFI_HANDLE                       *HandleBuffer;
+  EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL  *PciRootBridgeIo;
+
+  if (Reinstall) {
+    return NULL;
```
mhaeuser, line 26: What’s the point of the parameter if it just turns the function into a no-op?

Library/OcPciIoLib/OcPciIoLib.c
```diff
@@ -0,0 +1,76 @@
...
+    Status = gBS->LocateHandleBuffer (
+                    ByProtocol,
+                    &gEfiPciRootBridgeIoProtocolGuid,
+                    NULL,
+                    &HandleCount,
+                    &HandleBuffer
+                    );
+
+    if (EFI_ERROR (Status)) {
+      mCpuIo = NULL;
```
mhaeuser, line 49: This leaves mCpuIo modified, but the system is only in a partially patched state.

Library/OcPciIoLib/OcPciIoLib.c
```diff
@@ -0,0 +1,76 @@
...
+
+#include "OcPciIoU.h"
+#include <Library/OcPciIoLib.h>
+
+EFI_CPU_IO2_PROTOCOL *
+OcPciIoInstallProtocol (
+  IN BOOLEAN  Reinstall
+  )
+{
+  EFI_TPL                          tpl;
```
mhaeuser, line 17: Must start with a capital letter.

mhaeuser (changes requested):

Library/OcPciIoLib/OcPciIoLib.c
xCuri0: Changed to return mCpuIo

Include/Acidanthera/Library/OcPciIoLib.h
```diff
@@ -0,0 +1,33 @@
+/** @file
+  Copyright (c) 2006 - 2011, Intel Corporation. All rights reserved.<BR>
```
vit9696, line 2: Copyright.

Library/OcPciIoLib/OcPciIoLib.c
```diff
@@ -0,0 +1,75 @@
...
+  EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL  *PciRootBridgeIo;
+
+  mCpuIo = InitializeCpuIo2 ();
+  if (!mCpuIo) {
+    return NULL;
+  }
+
+  if (!Reinstall) {
+    return mCpuIo;
+  } else {
```
vit9696, line 32: No need for else.

Library/OcPciIoLib/OcPciIoLib.c
```diff
@@ -0,0 +1,75 @@
...
+                    ByProtocol,
+                    &gEfiPciRootBridgeIoProtocolGuid,
+                    NULL,
+                    &HandleCount,
+                    &HandleBuffer
+                    );
+
+    if (EFI_ERROR (Status)) {
+      mCpuIo = NULL;
+      goto Free;
```
vit9696, line 44: This will run free on garbage value in HandleBuffer.

Please also remove all gotos here and below.

Library/OcPciIoLib/OcPciIoLib.c
```diff
@@ -0,0 +1,75 @@
...
+    Status = gBS->LocateHandleBuffer (
+                    ByProtocol,
+                    &gEfiPciRootBridgeIoProtocolGuid,
+                    NULL,
+                    &HandleCount,
+                    &HandleBuffer
+                    );
+
+    if (EFI_ERROR (Status)) {
+      mCpuIo = NULL;
```
vit9696, line 43: No need to zero local variable CpuIo (currently mCpuIo) here and below.

Library/OcPciIoLib/OcPciIoLib.c
```diff
@@ -0,0 +1,75 @@
...
+    for (Index = 0; Index < HandleCount; ++Index) {
+      DEBUG ((DEBUG_INFO, "OCPIO: Fixing PciRootBridgeIo %d\n", Index));
+
+      Status = gBS->HandleProtocol (
+                      HandleBuffer[Index],
+                      &gEfiPciRootBridgeIoProtocolGuid,
+                      (VOID **)&PciRootBridgeIo
+                      );
+      if (EFI_ERROR (Status)) {
+        mCpuIo = NULL;
```
vit9696, line 61: This can just be break.

Library/OcPciIoLib/OcPciIoLib.c
```diff
@@ -0,0 +1,75 @@
...
+{
+  EFI_TPL                          Tpl;
+  UINTN                            Index;
+  EFI_STATUS                       Status;
+  EFI_CPU_IO2_PROTOCOL             *mCpuIo;
+  UINTN                            HandleCount;
+  EFI_HANDLE                       *HandleBuffer;
+  EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL  *PciRootBridgeIo;
+
+  mCpuIo = InitializeCpuIo2 ();
```
vit9696, line 25: This is a local variable, not global variable. Please rename `mCpuIo` to `CpuIo` in this function.

Library/OcPciIoLib/OcPciIoU.c
```diff
@@ -0,0 +1,446 @@
...
+  0,      // EfiCpuIoWidthFillUint16
+  0,      // EfiCpuIoWidthFillUint32
+  0       // EfiCpuIoWidthFillUint64
+};
+
+EFI_CPU_IO2_PROTOCOL *
+InitializeCpuIo2 (
+  )
+{
+  mCpuIo = NULL;
```
vit9696, line 51: - Please cache the value. If `mCpuIo` is not null, then return it.
- Check the return value of `gBS->LocateProtocol` and return NULL explicitly on failure.

Library/OcPciIoLib/OcPciIoU.c
```diff
@@ -0,0 +1,446 @@
...
+  8,      // EfiCpuIoWidthFifoUint64
+  0,      // EfiCpuIoWidthFillUint8
+  0,      // EfiCpuIoWidthFillUint16
+  0,      // EfiCpuIoWidthFillUint32
+  0       // EfiCpuIoWidthFillUint64
+};
+
+EFI_CPU_IO2_PROTOCOL *
+InitializeCpuIo2 (
+  )
```
vit9696, line 49: Missing VOID in braces.

Library/OcPciIoLib/OcPciIoU.c
```diff
@@ -0,0 +1,446 @@
...
+  if ((UINT32)Width >= EfiCpuIoWidthMaximum) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  //
+  // For FIFO type, the target address won't increase during the access,
+  // so treat Count as 1
+  //
+  if ((Width >= EfiCpuIoWidthFifoUint8) && (Width <= EfiCpuIoWidthFifoUint64)) {
+    Count = 1;
```
vit9696, line 116: This code overwrites Count, but the caller of this code (e.g. CpuMemoryServiceRead) is unaware of this change. Is this intentional?

Library/OcPciIoLib/OcPciIoU.c
```diff
@@ -0,0 +1,446 @@
...
+  // Select loop based on the width of the transfer
+  //
+  InStride       = mInStride[Width];
+  OutStride      = mOutStride[Width];
+  OperationWidth = (EFI_CPU_IO_PROTOCOL_WIDTH)(Width & 0x03);
+  for (Uint8Buffer = Buffer; Count > 0; Address += InStride, Uint8Buffer += OutStride, Count--) {
+    if (OperationWidth == EfiCpuIoWidthUint8) {
+      *Uint8Buffer = MmioRead8 ((UINTN)Address);
+    } else if (OperationWidth == EfiCpuIoWidthUint16) {
+      *((UINT16 *)Uint8Buffer) = MmioRead16 ((UINTN)Address);
```
vit9696, line 246: Use WriteUnalignedXX functions here and below.

Library/OcPciIoLib/OcPciIoU.c
```diff
@@ -0,0 +1,446 @@
...
+
+  //
+  // Select loop based on the width of the transfer
+  //
+  InStride       = mInStride[Width];
+  OutStride      = mOutStride[Width];
+  OperationWidth = (EFI_CPU_IO_PROTOCOL_WIDTH)(Width & 0x03);
+  for (Uint8Buffer = Buffer; Count > 0; Address += InStride, Uint8Buffer += OutStride, Count--) {
+    if (OperationWidth == EfiCpuIoWidthUint8) {
+      MmioWrite8 ((UINTN)Address, *Uint8Buffer);
```
vit9696, line 325: Use ReadUnalignedXX functions here and below.

Library/OcPciIoLib/OcPciIoU.h
```diff
@@ -0,0 +1,181 @@
...
+
+#include <Protocol/PciRootBridgeIo.h>
+#include <Protocol/CpuIo2.h>
+#include <Library/IoLib.h>
+
+#define MAX_IO_PORT_ADDRESS  0xFFFF
+
+EFI_CPU_IO2_PROTOCOL *
+InitializeCpuIo2 (
+  );
```
vit9696, line 29: Missing VOID

Library/OcPciIoLib/OcPciIoU.c
xCuri0: This code comes from EDK2, so I guess this is within spec.

Library/OcPciIoLib/OcPciIoU.c
xCuri0: There doesn't seem to be one for Uint8 though, same for below

vit9696: Well, obviously, since any area is 1-byte aligned. Only 16+ are affected by undefined behaviour in the current code.

xCuri0: Is this needed ? I can't figure this out without causing a compiler warning under GCC and MSVC. Using `WriteUnaligned32 ((UINT32 *)Address, *((UINT32 *)Uint8Buffer));` works fine under Clang/Xcode but gives a compiler warning on MSVC / GCC, so I tried using bitmask which usually fixes this error. 

`WriteUnaligned32 ((UINT32 *)(Address & 0xFFFFFFFF), *((UINT32 *)Uint8Buffer));` causes function to not work (freeze on connecting drivers). Tried `0xFFFFFFFFULL` too.

Maybe I'm missing something.

mhaeuser: Yes, this is needed. `(UINT32 *)(UINTN)Address`. Truncating the address on 64-bit is borked for obvious reasons.

xCuri0: That worked, thanks!

vit9696:
In my opinion, no big reason for it. Firmware / OS will not use it, and OPROMs start before this quirk gets applied.

xCuri0:
ignore the review request didn't mean to do that

Library/OcPciIoLib/OcPciIoLib.c
```diff
@@ -0,0 +1,73 @@
...
+  EFI_TPL                          Tpl;
+  UINTN                            Index;
+  EFI_STATUS                       Status;
+  EFI_CPU_IO2_PROTOCOL             *CpuIo;
+  UINTN                            HandleCount;
+  EFI_HANDLE                       *HandleBuffer;
+  EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL  *PciRootBridgeIo;
+
+  CpuIo = InitializeCpuIo2 ();
+  if (!CpuIo) {
```
vit9696, line 26: I think the code style for EDK is CpuIo == NULL, i.e. explicitly compare against actual value for all argument types but BOOLEAN.

Library/OcPciIoLib/OcPciIoU.c
```diff
@@ -0,0 +1,457 @@
...
+};
+
+EFI_CPU_IO2_PROTOCOL *
+InitializeCpuIo2 (
+  VOID
+  )
+{
+  EFI_STATUS  Status;
+
+  if (mCpuIo) {
```
vit9696, line 54: Same here, mCpuIo != NULL.

Library/OcPciIoLib/OcPciIoU.c
```diff
@@ -0,0 +1,457 @@
...
+  // Select loop based on the width of the transfer
+  //
+  InStride       = mInStride[Width];
+  OutStride      = mOutStride[Width];
+  OperationWidth = (EFI_CPU_IO_PROTOCOL_WIDTH)(Width & 0x03);
+  for (Uint8Buffer = Buffer; Count > 0; Address += InStride, Uint8Buffer += OutStride, Count--) {
+    if (OperationWidth == EfiCpuIoWidthUint8) {
+      *Uint8Buffer = MmioRead8 ((UINTN)Address);
+    } else if (OperationWidth == EfiCpuIoWidthUint16) {
+      *((UINT16 *)Uint8Buffer) = ReadUnaligned16 ((UINT16 *)(UINTN)Address);
```
vit9696, line 257: This should also have WriteUnaligned, just like Read.

Library/OcPciIoLib/OcPciIoU.c
```diff
@@ -0,0 +1,457 @@
...
+  // Select loop based on the width of the transfer
+  //
+  InStride       = mInStride[Width];
+  OutStride      = mOutStride[Width];
+  OperationWidth = (EFI_CPU_IO_PROTOCOL_WIDTH)(Width & 0x03);
+  for (Uint8Buffer = Buffer; Count > 0; Address += InStride, Uint8Buffer += OutStride, Count--) {
+    if (OperationWidth == EfiCpuIoWidthUint8) {
+      MmioWrite8 ((UINTN)Address, *Uint8Buffer);
+    } else if (OperationWidth == EfiCpuIoWidthUint16) {
+      WriteUnaligned16 ((UINT16 *)(UINTN)Address, *((UINT16 *)Uint8Buffer));
```
vit9696, line 338: This should have ReadUnaligned.

vit9696 (commented):
Looks generally good to me. Left few comments, including not fully addressed unaligned stuff.

Library/OcPciIoLib/OcPciIoU.c
xCuri0: There is no WriteAligned/ReadAligned or WriteAligned8/ReadAligned8 function though. Previously it was explained that it isn't needed for these because 1 byte will always be aligned.

mhaeuser: But you write a 16-bit value, not 8-bit.

Library/OcPciIoLib/OcPciIoU.c
```diff
@@ -0,0 +1,457 @@
...
+
+  //
+  // Select loop based on the width of the transfer
+  //
+  InStride       = mInStride[Width];
+  OutStride      = mOutStride[Width];
+  OperationWidth = (EFI_CPU_IO_PROTOCOL_WIDTH)(Width & 0x03);
+  for (Uint8Buffer = Buffer; Count > 0; Address += InStride, Uint8Buffer += OutStride, Count--) {
+    if (OperationWidth == EfiCpuIoWidthUint8) {
+      *Uint8Buffer = MmioRead8 ((UINTN)Address);
```
mhaeuser, line 255: Replace MmioRead with a direct access

Library/OcPciIoLib/OcPciIoU.c
mhaeuser: Not resolved, use direct access

Library/OcPciIoLib/OcPciIoLib.c
mhaeuser: Why is this done before the check below?

Library/OcPciIoLib/OcPciIoLib.c
```diff
@@ -0,0 +1,73 @@
...
+  EFI_HANDLE                       *HandleBuffer;
+  EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL  *PciRootBridgeIo;
+
+  CpuIo = InitializeCpuIo2 ();
+  if (CpuIo == NULL) {
+    return NULL;
+  }
+
+  if (!Reinstall) {
+    return CpuIo;
```
mhaeuser, line 31: This *still* is a no-op if Reinstall is FALSE, no? As far as I can see, it does not work like the other protocol overrides at all.
@vit9696 Why isn’t this a simple, conditional quirk?

Library/OcPciIoLib/OcPciIoU.c
```diff
@@ -0,0 +1,457 @@
...
+
+   Copyright (c) 2006 - 2019, Intel Corporation. All rights reserved.<BR>
+   Copyright (c) 2023, xCuri0. All rights reserved.<BR>
+   SPDX-License-Identifier: BSD-2-Clause-Patent
+
+ **/
+
+#include "OcPciIoU.h"
+
+EFI_CPU_IO2_PROTOCOL  *mCpuIo = NULL;
```
mhaeuser, line 12: STATIC?

Library/OcPciIoLib/OcPciIoU.c
```diff
@@ -0,0 +1,457 @@
...
+
+UINT8  mInStride[] = {
```
mhaeuser, line 14: STATIC CONST

mhaeuser (changes requested):

Library/OcPciIoLib/OcPciIoU.c
xCuri0: @mhaeuser but isn't `Uint8Buffer` is supposed to be 8 bit in 
```
    if (OperationWidth == EfiCpuIoWidthUint8) {
      MmioWrite8 ((UINTN)Address, *Uint8Buffer);
```
don't really see how it can be 16 bit

vit9696: It cannot be. Lines with 8 bits are ok. Please google Unaligned accesses in C/C++ it may help you to better understand things.

mhaeuser: The comments refer to the EXACT lines. The line for this comment is

```
*((UINT16 *)Uint8Buffer) = ReadUnaligned16 ((UINT16 *)(UINTN)Address);
```

not

```
if (OperationWidth == EfiCpuIoWidthUint8) {
      MmioWrite8 ((UINTN)Address, *Uint8Buffer);
```

xCuri0: So should I check the alignment first before allowing writes ?

The original code from EDK using ```MmioWrite``` would ASSERT if the memory address was not aligned with the width (like 16 bits)

mhaeuser: No, WriteUnaligned16, that is exactly what this API is for, you are not performing MMIO

xCuri0: I'm confused what's being requested here because I'm already using ReadUnaligned/WriteUnaligned for Uint16/Uint32/Uint64 and direct access for Uint8

mhaeuser: No, you are not, you literally cast to UINT16 * and deference…….

xCuri0: Think I understand it now, I've pushed changes that use WriteAlignedXX instead of dereferencing pointer.

PR #461, opened 2023-05-12
Audio: Add ExitBSWait option

xCuri0:
Useful when using longer audio files, previously this option was just set to wait always.

### DIFF
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -7822,6 +7822,15 @@ for additional options.
   \hyperref[uefiaudio]{AudioDxe} section.
 
   \item
+  \texttt{ExitBSWait}\\
+  \textbf{Type}: \texttt{plist\ boolean}\\
+  \textbf{Failsafe}: \texttt{true}\\
+  \textbf{Description}: Wait for audio to finish playing on ExitBootServices
+
+  Enabled by default, you might want to disable this if you use a longer OCEFIAudio_VoiceOver_Boot
+  or other audio file.
+
+  \item
   \texttt{MaximumGain}\\
   \textbf{Type}: \texttt{plist\ integer}\\
   \textbf{Failsafe}: \texttt{-15}\\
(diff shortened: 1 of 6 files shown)


mikebeaton:
I'm not convinced we want this one - it's a minor speed up that's going to sound like a bug when enabled!

You are welcome to raise an issue before spending time developing a new feature, to discuss whether or not a change is going to fit. (Many projects are very strict about this, we're more laissez faire, if the code is good - but just to avoid you wasting your time.)

I'm not the only voice on this - of course - and I'll leave it open for now, but that's my thoughts for what they're worth. That said, many thanks for your several and good contributions recently.

mikebeaton:
> I added this so that I can put a longer boot audio, not for boot time reasons. Already was using this in my build of OpenCore, just put my changes here in-case it's found useful.

Thanks for offering it.

I'm still not convinced, unless you have an example of a longer boot audio that is desirable to have but also desirable to cut off when you boot, without waiting for it to complete - I mean other than e.g. playing your favourite track for fun, or similar!

mikebeaton:
I'll close for now.

I'd day the status is: possibly convincable, if you can describe a use case that matches the existing purpose - a boot sound - or extends in a way that is agreed to be useful, and is likely to be useful to more than a small handful of people. (Would probably suggest changing the option name, in that case.) But there'd be a prima facie case again adding more options to the main config file, without a compelling use case.

Comments on closed issues and PRs are monitored, on this project.

vit9696:
I do wonder if this option may make sense for blind persons with voice assistant enabled when autobooting (i.e. booting without interaction). You can put some beacon sound to confirm that the UEFI environment is still running, which will stop as soon as the kernel starts and confirm that the boot process started. I wonder whether we should loop the sound file in this case though.

mikebeaton:
Well can be, definitely. Something like @xCuri0's existing PR, but with three options maybe?
 - Play and wait (as now)
 - Play and no wait (other option in @xCuri0's current PR)
 - Loop and no wait (new suggested idea)

PR #462, opened 2023-05-17
UpdateSSDT-PLUG.dsl

SeemRan7:
For me


Cam396:
This won't work as the prebuilt is designed to work on all applicable hardware, not just your hardware.

PR #463, opened 2023-05-18
Address new console text issues

mikebeaton:
This suite of commits addresses all the items mentioned by PM (except for a possible issue about multiple key press responses):

- Added Unicode font pages for some accented characters, plus box drawing and block elements, to `Builtin` text renderer
- Improved support for overlong menu entries and very narrow console modes in builtin picker (not mentioned, but fits and something I've been meaning to do)
- Made `Builtin` text renderer ignore UI Scale, when required to ensure that text mode reaches minimum supported size of 80x25
- Added save and restore of text and graphics mode round tools and failed boot entries
- Updated out-of-range cursor handling to work round minor display issue (bug?) in memtest86

Here are some after and before images. Note when looking at the 'After' images that we do not yet (in this PR) have all the accented characters which the memtest86 translations use (hence some accents are present, but some underscores are still present).

It seems that some sort of EFI/Unicode standard (at least, what memtest86 is expecting) is indeed the chars listed as the 'Extended Unicode character set' here: https://int10h.org/oldschool-pc-fonts/fontlist/font?ibm_vga_8x16 .

Currently code page 1 is the free font used by xnu (same as previous code page 0), and I took the additional glyphs (pp.74&75) from `ib8x16u.bdf` from https://github.com/farsil/ibmfonts . But the best font in that project, if we were to want all characters (as the memtest86 translations are expecting), is `iv8x16u.bdf`. You can preview these font files with `gbdfed` (available via brew). I could continue to make a reasonably sparse (i.e. memory efficient) implementation that uses all of it? (Note that it would then include Russian characters... ;) )

**After**:

**Before**:


mikebeaton:
Here are a couple of quick updates:

 - I tested memtest86 on a MacPro10,2, and it renders all the box drawing characters as blank spaces. It also does not even offer the language localisations (same install of memtest86 from the same USB does offer them, on my hack; I guess because we are not implementing an error code for characters we do not support? Might be an idea to do so? Not sure ). I do not have a newer UEFI Mac to test whether more font support is available later.
 - It's true that the font I found is not as pretty as the one we have (open license font also used by xnu); I don't think this affects box drawing chars, which can't really vary; but it affects the actual font _per se_. Not sure how to deal with that, e.g. whether there's any source of a fuller implementation of the font we have. Or whether it would even be possible/desirable to implement an optional font loader..?

PR #470, opened 2023-05-29
OcConsoleLib: Provide ConsoleFont option to load custom console font

mikebeaton:
The main serious usage of this feature is to enable non-English character set support, in those rare EFI programs which can use it (such as memtest86):

These two screenshots both use [Terminus](https://terminus-font.sourceforge.net/) under the SIL Open Font License, Version 1.1 (added in https://github.com/acidanthera/OcBinaryData/pull/11):

Of course, the feature can also be used for more or less silly - or at least decorative - alternative console fonts. (I found a reasonably nice set of 8x16 GPL'd console fonts, so made them available, updated with compatible character encoding, separately here: https://github.com/mikebeaton/epto-fonts .):

Tremolo2:

2 x Deco:

2 x Fancy:

Display:


mikebeaton:
I've gone back and forth a bit about whether to propose including 'TerminusCore' as well as 'Terminus' console font, but, well ... I prefer it, so I've added it back, in a separate commit (on this PR and on the linked PR in OcBinaryData). (I think there is no issue with providing this modified version under the SIL open font license.)

Library/OcConsoleLib/TextOutputBuiltin.c
```diff
@@ -620,6 +82,89 @@ STATIC EFI_CONSOLE_CONTROL_SCREEN_MODE      mConsoleMode = EfiConsoleControlScre
...
+      ASSERT (Success == TRUE);
+      if (!UseFallback || (Success == FALSE)) {
+        return FALSE;
+      }
+
+      Success = FALSE;
+
+      //
+      // For default font, render all unknown chars in first two pages as space (following the explicit
+      // spaces in free console font used in XNU); in other pages and in user fonts use underscore.
```
mikebeaton, line 141: Not sure that I really need this special case - unless exactly matching what e.g. Mac BIOS might render, in the (I guess) unlikely event of using these chars, really matters?

Library/OcConsoleLib/TextOutputBuiltin.c
```diff
@@ -620,6 +82,89 @@ STATIC EFI_CONSOLE_CONTROL_SCREEN_MODE      mConsoleMode = EfiConsoleControlScre
...
+      //
+      if (mIsDefaultFont && (PageNumber < 2)) {
```
mikebeaton, line 143: Note to self: If I'm not including ISO Latin page 1 - as I'm currently not in this PR - this PageNumber test should be 1, not 2. To change, if we are keeping the special case code.

Library/OcConsoleLib/TextOutputBuiltin.c
```diff
@@ -620,6 +82,89 @@ STATIC EFI_CONSOLE_CONTROL_SCREEN_MODE      mConsoleMode = EfiConsoleControlScre
 #define MIN_SUPPORTED_CONSOLE_WIDTH   (80)
 #define MIN_SUPPORTED_CONSOLE_HEIGHT  (25)
 
+BOOLEAN
```
mikebeaton, line 85: Another note to self: This may need EFI_STATUS, and a different status for 'returned usable fallback character' vs. 'could not return anything'?

Library/OcConsoleLib/TextOutputBuiltin.c
```diff
@@ -620,6 +82,107 @@ STATIC EFI_CONSOLE_CONTROL_SCREEN_MODE      mConsoleMode = EfiConsoleControlScre
...
+      ASSERT (Status == EFI_SUCCESS); ///< NB EFI_WARN_... are not errors
+      if (!UseFallback || (Status != EFI_SUCCESS)) {
+        return EFI_LOAD_ERROR;
+      }
+
+      Status = EFI_WARN_UNKNOWN_GLYPH;
+
+      //
+      // For default font, render all unknown chars in provided page(s) as space (following the explicit
+      // spaces in free console font used in XNU); in other pages and in user fonts use underscore.
```
mikebeaton, line 155: Not sure that we need the special case of rendering spaces in default font - unless exactly matching what e.g. Mac BIOS might render, in the (I guess) unlikely event of using these chars, really matters?

Library/OcConsoleLib/ConsoleFontLoader.c
```diff
@@ -0,0 +1,471 @@
...
+  CHAR8   *CurrentLine,
+  CHAR16  *Char,
+  UINT8   *LineGlyphData
+  )
+{
+  EFI_STATUS  Status;
+  UINTN       Data;
+  UINT8       Line;
+
+  if (CurrentLine[4] != ':') {
```
vit9696, line 88: Is this out of buffer access? Was this code fuzz-tested?

vit9696: Oh, I think HEX_LINE_LENGTH checks it.

Library/OcConsoleLib/TextOutputBuiltin.c
vit9696: Yeah, space is nicer because it does not immediately raise an issue that something is wrong. Maybe we want it for other fonts too really.

vit9696:
This looks ready to merge to me, but I wish we fuzz-test it eventually. Thank you!

Library/OcConsoleLib/TextOutputBuiltin.c
mikebeaton: Hi - Thank you for the review.

So, do we want any change to the current PR here?

We previously used space in the provided page of the provided font, and underscore outside it. My understanding of the use of space for non-present chars within the provided page of the provided font, in the previous code, was that it was because the original free font, as used in XNU, actually had blank char data at these code points - so not storing these zeroes, but rendering space when these chars were requested, was a minor storage optimisation.

However, I'm not sure there's any reason that anything would ever render glyphs from 0x0-0x1f, or 0x7f, which is what this applies to; and so, if this does happen, I think I'd prefer to see an underscore, and have a hint that's something's up, rather than to see a space and probably assume it was meant to be a space.

i.e. I think on balance I'd propose leaving it as is in the PR: we effectively keep the existing font as if it provided these characters, as blank glyphs, and other fonts provide just what they provide, and outside what is provided, we render underscore.

mikebeaton: Okay thanks for response by PM - will merge as is for now

PR #480, opened 2023-07-21
Port RehabMan's ProBook 4x30s EFI module(s) to OpenCore

ubihazard:
Added two ProBook 4x30s EFI modules to Staging: BIOS fan reset and BIOS Wi-Fi whitelist bypass. Built, tested and working on my ancient ProBook 4530s: successfully replaced Atheros AR9285 with non-whitelisted Broadcom BCM94352HMB Bluetooth & Wi-Fi card (Wi-Fi side of things also required kapton tape masking on a certain PCB contact on a card itself, but that‘s another story).

All credits go to @RehabMan, – I merely did the porting and separated the [original EFI module](https://github.com/RehabMan/HP-ProBook-4x30s-Fan-Reset) in two (original combines fan reset and Wi-Fi hack together).

I make it clear that this is only for 4x30s series ProBook models and for HM65 boards. If you think this is too specific, just close this PR, – I‘ll just keep my fork as a backup. This device is very old and probably wouldn‘t interest many people.


Staging/ProBook6sC200sBiosFanReset/ProBookFanReset.c
```diff
@@ -0,0 +1,304 @@
...
+ * HP ProBook 4330s, 4530s, and 4730s models only:
+ * running this code on any other laptop might brick the device.
+ */
+
+/*-
+ * Copyright (c) 2012 RehabMan
+ *               2023 Ubihazard
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
```
vit9696, line 15: Unfortunately we do not accept GPL code in OpenCore.

ubihazard: I see, it makes sense. Unfortunately, @RehabMan released his original code GPL‘d, and this can‘t be changed (and I can‘t relicense it).

I‘ll just keep it in my fork then.

vit9696: Thank you for your hard work!

PR #481, opened 2023-08-01
Build: Switch workflows from macos-latest to macos-13

dhinakg:
GitHub's macOS 13 CI runners run on new(er) hardware, providing a ~30% improvement in build times. `macos-latest` is currently macOS 12 - we can switch back to `macos-latest` once it is >= macOS 13.

Additionally, this updates the build scripts to print the error message if `curl` fails, which should provide some more clarity if network or server errors occur.

Please make sure Docker still works, as far as I understand this should not break anything but I don't use Docker.


vit9696:
Thanks! Could you also update other repositories, like audk?

Also, should we update kexts?

PR #483, opened 2023-08-30
OcVariableLib: Add option to set bs-only var ShimLockGuid:ShimRetainProtocol

mikebeaton:
If set by bootloader chained from shim before further image is loaded, instructs shim to keep its protocol and its security overrides (if enabled) present for all image loads, not just the first.

May be of more use with additional scripts to be added in ShimUtils (along with shim-to-cert), but works as stated now.

Think this should be okay as is, but not sure if anyone wants to quickly scan for anything. E.g.:
 - Could have split into a new library, but seems overkill.
 - Don't think there are any issues in including this small amount of shim-related stuff, but thought should check before pushing.


Library/OcVariableLib/Shim.c
```diff
@@ -0,0 +1,30 @@
...
+#include <Uefi.h>
+#include <ShimVars.h>
+#include <Library/UefiRuntimeServicesTableLib.h>
+
+EFI_STATUS
+OcShimRetainProtocol (
+  IN BOOLEAN  Required
+  )
+{
+  UINT8  ShimRetainProtocol = 1;
```
mhaeuser, line 17: Need to separate declaration and init

Include/Grub2/ShimVars.h
```diff
@@ -0,0 +1,18 @@
...
+  GRUB2 shim values.
+
+  Copyright (c) 2023, Mike Beaton. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-3-Clause
+**/
+
+#ifndef __SHIM_H
+#define __SHIM_H
+
+#include "Guid/ShimLock.h"
```
mhaeuser, line 11: <> instead of “”

mhaeuser (commented):
LGTM, tyvm

PR #487, opened 2023-09-22
Close file pointer before calling abort

rilysh:
Hi.
This is a nitpick, but before calling abort in `fwrite(...)`, the `UserWriteFile(...)` function doesn't close the file pointer, which will cause a memory leak. This PR adds `fclose(...)` before the abort.


PMheart:
Great reminder. Thanks!

PR #488, opened 2023-09-22
Clarify Handling of 10.6 SnowLeopard (32-bit)

dakanji:
Amends wording in documentation for the `KernelArch` setting to make clear that `x86_64` is only defaulted to on 10.6 SnowLeopard when `Auto` is set and that the `i386` options can be forced by explicitly setting them.

A lot of unrelated whitespace issues were flagged and have been put into a separate commit to be kept or discarded as wanted.

### DIFF
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -3012,15 +3012,15 @@ to install and troubleshoot such macOS installations.
       \texttt{x86\_64}, \texttt{i386}, \texttt{i386-user32}.
   \end{enumerate}
 
-  Unlike macOS~10.7 (where certain board identifiers are treated as the \texttt{i386}
-  only machines), and macOS~10.5 or earlier (where \texttt{x86\_64} is not supported
-  by the macOS kernel), macOS~10.6 is very special. The architecture choice on macOS~10.6
-  depends on many factors including not only the board identifier, but also the macOS
-  product type (client vs server), macOS point release, and amount of RAM. The detection
-  of all these is complicated and impractical, as several point releases had implementation
-  flaws resulting in a failure to properly execute the server detection in the first place.
-  For this reason, OpenCore on macOS~10.6 falls back on the \texttt{x86\_64}
-  architecture whenever it is supported by the board, as it is on macOS~10.7.
+  Kernel architecture selection on macOS~10.6 and 10.7 depends on multiple
+  factors such as the board identifier, the macOS product type (client vs server),
+  macOS point release, as well as the amount of RAM present. It is not practical
+  for OpenCore to enumerate and process all these variables; especially given
+  that several macOS point releases had implementation flaws that affect the
+  server detection process in the first place. Hence, when \texttt{Auto} is set,
+  OpenCore falls back on the \texttt{x86\_64} architecture whenever this is
+  supported by the board when loading macOS~10.6 or 10.7. The 32-bit options can
+  however be explicitly configured when required.
 
   A 64-bit Mac model compatibility matrix corresponding to actual
   EfiBoot behaviour on macOS 10.6.8 and 10.7.5 is outlined below.


Docs/Configuration.tex
```diff
@@ -3012,15 +3012,21 @@ \subsection{Scheme Properties}\label{kernelpropsscheme}
...
+  \texttt{i386} only machines), and macOS~10.5 or earlier (where
+  \texttt{x86\_64} is not supported by the macOS kernel), kernel architecture
+  selection on macOS~10.6 depends on several factors such as the board
+  identifier, the macOS product type (client vs server), macOS point release,
+  as well as the amount of RAM present.
+
+  Enumerating and processing all these variables by OpenCore to determine
+  the kernel architecture is complicated and impractical, more so given that
+  several macOS~10.6 point releases had implementation flaws that prevent
+  properly executing the server detection process in the first place. Hence,
```
mhaeuser, line 3025: Maybe this can be simplified to “prevent detecting”?

mhaeuser (commented):
Tyvm

Docs/Configuration.tex
```diff
@@ -3012,15 +3012,15 @@ \subsection{Scheme Properties}\label{kernelpropsscheme}
...
-  by the macOS kernel), macOS~10.6 is very special. The architecture choice on macOS~10.6
-  depends on many factors including not only the board identifier, but also the macOS
-  product type (client vs server), macOS point release, and amount of RAM. The detection
-  of all these is complicated and impractical, as several point releases had implementation
-  flaws resulting in a failure to properly execute the server detection in the first place.
-  For this reason, OpenCore on macOS~10.6 falls back on the \texttt{x86\_64}
-  architecture whenever it is supported by the board, as it is on macOS~10.7.
+  Kernel architecture selection on macOS~10.6 depends on several factors such
+  as the board identifier, the macOS product type (client vs server), macOS
+  point release, as well as the amount of RAM present. It is not practical for
```
mikebeaton, line 3017: It's definitely worth adding the clarification of the macOS 10.6 (and also 10.7) 32 bit options below, but we've lost useful contrasting information here about the native behaviour of macOS 10.5 and 10.7.

Docs/Configuration.tex
```diff
@@ -5170,15 +5170,15 @@ \subsection{Other Variables}\label{nvramvarsother}
 
   \begin{itemize}
   \item \texttt{boot-save-log=VALUE} --- debug log save mode for normal boot.
-  	\begin{itemize}
+    \begin{itemize}
```
mikebeaton, line 5173: All the whitespace cleanups look fine and useful to me.

Docs/Configuration.tex
dakanji: Actually, the whole section is about 10.6 SnowLeopard but fails to adequately do what it set out to do while adding superfluous detail about other versions.

10.8/newer do not have the fact that they would only ever boot into 64-bit pointed out in this section and this has now been similarly applied to 32-bit only instances (10.5/older as well as some 10.7).

The potential unexpected behaviour for 10.6 and some 10.7 is covered here as well what to ensure an expected outcome is always the case.

mikebeaton: It is not correct that "the whole section is about 10.6 SnowLeopard". The whole section is about the KernelArch parameter. The original whole paragraph included useful contrasting information about macOS 10.5 and 10.7, as well as 10.6, which has been lost. It is not correct that "10.8/newer do not have the fact that they would only ever boot into 64-bit pointed out in this section":

Screenshot 2023-09-24 at 04 47 54

As per your PR description, you want to "make clear that `x86_64` is only defaulted to on 10.6 SnowLeopard when `Auto` is set and that the `i386` options can be forced by explicitly setting them." All you need to retain the lost but useful additional information, while adding the reasonable clarification which you want to make, is to change is the last sentence of the paragraph, e.g. as follows:

```
  For this reason, when \texttt{Auto} is set OpenCore on macOS~10.6 falls back
  on the \texttt{x86\_64} architecture whenever it is supported by the board, as
  on macOS~10.7. The 32-bit \texttt{KernelArch} options can still be configured
  explicitly however.
```

dakanji: I meant the paragraph that was changed.
Not sure i fully understand the suggested text but will add it as requested alter.

Thanks

mikebeaton: Okay, I see you've already made a change to the last sentence somewhat along these lines. To be clear, I'm suggesting - for the sake of quickly making the change you want, but not losing the other relevant information in there - that you _only_ change the last sentence, and the that the rest of the wording of the paragraph doesn't need changing. Thank you.

Docs/Configuration.tex
mikebeaton: I've pushed the whitespace handling changes - tyvm. (If you want to rebase your PR branch onto master, it should automatically drop that commit.)

PR #493, opened 2023-10-26
Debug: Add support for 32-bit on 32-bit under GDB

mikebeaton:
In addition to existing support for 32-bit on 64-bit

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@ OpenCore Changelog
 #### v0.9.6
 - Updated builtin firmware versions for SMBIOS and the rest
 - Fixed hang while generating boot entries on some systems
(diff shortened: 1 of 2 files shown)


mikebeaton:
Apologies - directory names need updating for change from `Ia32` to `IA32` (and will prevent it working on case sensitive FS like Unix until changed) - will do this eve.

mikebeaton:
Sorry, running commentary, but I've switched back to `Ia32` as the canonical name (but allowing `IA32` now, e.g. same as target arch spec for EDK and OC build scripts).

mikebeaton:
Checked working in Linux (gdb 32/32, gdb 32/64, lldb 64/64) and macOS (gdb 32/32, gdb 32/64, lldb 64/64).

vit9696:
Looks good to me, merged, tyvm!

PR #495, opened 2023-10-29
OcBootManagementLib: Add guard for static mImageLoaderCaps

mikebeaton:


Library/OcBootManagementLib/ImageLoader.c
```diff
@@ -940,6 +944,13 @@ InternalEfiStartImage (
   OC_LOADED_IMAGE_PROTOCOL   *OcLoadedImage;
   EFI_LOADED_IMAGE_PROTOCOL  *LoadedImage;
 
+  if (  (mImageLoaderConfigure  != NULL)
+     && (mImageLoaderCapsHandle != ImageHandle))
+  {
+    ASSERT (mImageLoaderCapsHandle == ImageHandle);
```
mhaeuser, line 950: If you really want an `ASSERT`, make it `ASSERT(FALSE)`, but why not just use a print statement which can also output the handles?

Library/OcBootManagementLib/ImageLoader.c
```diff
@@ -835,6 +836,7 @@ InternalEfiLoadImage (
  #else
   mImageLoaderCaps = OC_KERN_CAPABILITY_K64_U64;
  #endif
+  mImageLoaderCapsHandle = NULL;
```
mhaeuser, line 839: I would set this way earlier. While it would normally make sense to only set when we actually load, setting it earlier would allow to catch bugged or weird Load/Start mismatches that would otherwise be masked.

Library/OcBootManagementLib/ImageLoader.c
```diff
@@ -773,6 +774,8 @@ InternalEfiLoadImage (
   VOID        *AllocatedBuffer;
   UINT32      RealSize;
 
+  mImageLoaderCapsHandle = NULL;
```
mikebeaton, line 777: Here?

Library/OcBootManagementLib/ImageLoader.c
```diff
@@ -940,6 +945,13 @@ InternalEfiStartImage (
   OC_LOADED_IMAGE_PROTOCOL   *OcLoadedImage;
   EFI_LOADED_IMAGE_PROTOCOL  *LoadedImage;
 
+  if (  (mImageLoaderConfigure  != NULL)
+     && (ImageHandle != mImageLoaderCapsHandle))
+  {
+    DEBUG ((DEBUG_ERROR, "OCB: load/start unsupported ordering, %p != %p\n", ImageHandle, mImageLoaderCapsHandle));
```
mikebeaton, line 951: ERROR, or WARNING?

PR #498, opened 2023-11-02
OcPeCoffExtLib: Add support for FixupAppleEfiImages quirk

mikebeaton:
Fixes https://github.com/acidanthera/bugtracker/issues/2233

For the purposes of viewing the changes, I strongly recommend https://github.com/mikebeaton/OpenCorePkg/commit/b6d5baf1e5af61b1c4b0888cd133ea02b9ca6b9f (i.e. commit 2/3), since this makes it very clear how OcPeCoffFixup.c relates to PeCoffInit.c which it came from.
 - Commit 1/3 imports the two files from BasePeCoffLib2 with no changes (except renaming PeCoffInit.c -> OcPeCoffFixup.c)
  - Commit 2/3 all the actual changes, making it easy to see what changed between PeCoffInit.c -> OcPeCoffFixup.c
  - Commit 3/3 just the required uncrustify changes to OcPeCoffFixup.c

Since all three commits are valid builds of OpenCore (with only Uncrustify errors, only in one file, in 1/3 and 2/3), then I suggest it might be worth keeping this three-commit structure in the main branch (i.e. not accepting as a merge commit, but just pushing the tip of this branch to master to accept the changes and auto-close the PR), so that it keeps the reference commit where it's easy to see the changes made between PeCoffInit.c and OcPeCoffFixup.c.

### DIFF
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -1611,6 +1611,34 @@ To view their current state, use the \texttt{pmset -g} command in Terminal.
   Refer to the \texttt{OCABC: MAT support is 1/0} log entry to determine whether MAT is supported.
 
 \item
+  \texttt{FixupAppleEfiImages}\\
+  \textbf{Type}: \texttt{plist\ boolean}\\
+  \textbf{Failsafe}: \texttt{false}\\
+  \textbf{Description}: Fix errors in early Mac OS X boot.efi images.
+
+  Modern secure PE loaders will refuse to load \texttt{boot.efi} images from
+  Mac OS X 10.4 and 10.5 due to these files containing \texttt{W\^{}X} errors
+  and illegal overlapping sections.
+
+  This quirk detects these issues and pre-processes such images in memory,
+  so that a modern loader can accept them.
+
+  Pre-processing in memory is incompatible with secure boot, as the image loaded
+  is not the image on disk, so you cannot sign files which are loaded in this way
+  based on their original disk image contents.
+  Certain firmware will offer to register the hash of new, unknown images - this would
+  still work. On the other hand, it is not particularly realistic to want to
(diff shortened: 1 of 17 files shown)


mikebeaton:
Worth pointing out that the functionality I've implemented seems to overlap somewhat (but not quite enough, I think) with this unused functionality: https://github.com/acidanthera/OpenCorePkg/pull/498/files#diff-a9f15dff840fe93b2f4cad53afc58083d8b3d8638439812968501119926b88a5R913-R917

Docs/Configuration.tex
```diff
@@ -1610,6 +1610,34 @@ \subsection{Quirks Properties}\label{booterpropsquirks}
...
+  still work. On the other hand, it is not particularly realistic to want to
+  start such early, insecure images with secure boot anyway.
+
+  \emph{Note 1}: The quirk is only applied to Apple-specific `fat' (both 32-bit and 64-bit
+  versions in one image) \texttt{.efi} files, and is never applied during the Apple secure
+  boot path for newer macOS.
+
+  \emph{Note 2}: The quirk is only needed for loading Mac OS X 10.4 and 10.5 in 32-bit
+  mode, and even then only if the firmware itself includes a modern, more secure PE COFF
+  image loader.
```
mikebeaton, line 1639: I'm right about thjis, right? So in this case, not HyperV and not VMWare, but basically just OVMF compiled from either audk or UDK, I believe?

vit9696: Well, better to mention OpenDuet I think…

Also 64-bit ones are also broken.

vit9696:
Code looks perfectly correct to me.

mikebeaton:
> Code looks perfectly correct to me.

Good code design, obviously! ;)

PR #499, opened 2023-11-07
OpenDuet/OcBootManagementLib: Disable W^X auto-fix in Duet, and move fix for all Apple images to FixupAppleEfiImages quirk

mikebeaton:
Duet then matches behaviour of OVMF, and quirk can be used in both Duet and OVMF to load legacy Apple images.


OpenDuetPkg.dsc
```diff
@@ -259,7 +259,7 @@
   gEfiMdePkgTokenSpaceGuid.PcdFixedDebugPrintErrorLevel|0x0
   gEfiMdePkgTokenSpaceGuid.PcdReportStatusCodePropertyMask|0x0
   gOpenCorePkgTokenSpaceGuid.PcdCanaryAllowRdtscFallback|TRUE
-  gEfiMdePkgTokenSpaceGuid.PcdImageLoaderRemoveXForWX|TRUE
+  gEfiMdePkgTokenSpaceGuid.PcdImageLoaderRemoveXForWX|FALSE
```
vit9696, line 262: I think we can just drop the var, but perhaps safer to explicitly have it.

vit9696 (commented):
Looks good to me.

OpenDuetPkg.dsc
mikebeaton: Thank you for review. Well... drop I guess. Will do and merge.

PR #500, opened 2023-11-08
Utilities/OcPeCoffExtLib: Update AppleEfiSignTool to work with new PE COFF loader

mikebeaton:
- Adds PE COFF fixup path which initialises context as if fixups were made but does not modify image, so that signing information can be found then used.

This PR is more intrusive than the previous Apple legacy signing related PR, but I believe is more or less the minimum required to allow AppleEfiSignTool to work with new PE COFF loader.


Utilities/AppleEfiSignTool/AppleEfiSignTool.c
```diff
@@ -208,7 +227,7 @@ LLVMFuzzerTestOneInput (
       CopyMem (Copy, Data, Size);
 
       NewSize = (UINT32)Size;
-      PeCoffVerifyAppleSignature (Copy, &NewSize);
+      PeCoffVerifyAppleSignature (Copy, &NewSize, FALSE);
```
mikebeaton, line 230: Here and just above, if commented in, needs more work to cover new code path accessed with TRUE.

mikebeaton:
Just realised that the actual ASSERTs here are related to recent audk change. Something like this should still be required, but let me update wording etc. as required.

Utilities/AppleEfiSignTool/AppleEfiSignTool.c
mikebeaton: On reflection, the path accessed with TRUE is not enabled except in userspace builds, so this is probably still enough to test code that will be enabled in production.

mikebeaton:
Assuming userspace Pcd is changed then without this new code we simply cannot look inside 32-bit slices (because of their overlapping section errors). But they no longer crash the tool, and no longer prevent it looking inside e.g. following 64-bit slice, which it can still understand.

However, when we do look inside the 32-bit images, with these changes, all the tool tells us is that none of them are signed (and none of them are apfs drivers) - which we kind of knew anyway.

**Disadvantages**:
 - Over-complex for small gain?

**Advantages**:
 - Restores ability to look inside all the images which AppleEfiSignTool could previously look inside
 - The ability to verify the (lack of!) Apple signature for section overlap images, and more generally to set up usable image context info without modifying the image itself, might end up being useful elsewhere??
 - The changes add the ability in this tool to see the fixups which would be made for a legacy Apple efi image (without having to load it in OC with debugging on, for instance)
   - However, this could be achieved with a small separate new tool, without this new code

mikebeaton:
Now that the Pcds are fixed, I only need the 'create image context for broken images without modifying the image in memory' code in order to verify the signature of files with section overlap errors and a signature. Since there are actually no such files in the wild, I can rewrite this without that new code. Will do so.

PR #502, opened 2023-11-09
Utilities/OcPeCoffExtLib: Update AppleEfiSignTool to work with new PE COFF loader

mikebeaton:
Simpler than previous approach and no need for a special 'fixup image context without fixing image' codepath (because there are no signed images with section overlaps, which is where that would really be needed).


Include/Acidanthera/Library/OcPeCoffExtLib.h
```diff
@@ -27,6 +28,26 @@ PeCoffVerifyAppleSignature (
...
+
+  @retval EFI_SUCCESS on success.
+**/
+EFI_STATUS
+InternalPeCoffVerifyAppleSignatureFromContext (
+  IN OUT PE_COFF_LOADER_IMAGE_CONTEXT  *ImageContext,
+  IN OUT  UINT32                       *ImageSize
+  );
+
+#endif
```
mikebeaton, line 49: Here an just below: possibly slightly janky way of making these internals visible to user space utils only, or possibly okay?

Utilities/AppleEfiSignTool/AppleEfiSignTool.c
```diff
@@ -44,34 +45,43 @@ PrintHelp (
...
 {
-  EFI_STATUS   Status;
-  UINT32       OrgImageSize;
-  CONST CHAR8  *Slice;
+  EFI_STATUS                    Status;
+  EFI_STATUS                    ApfsStatus;
+  UINT32                        OrgImageSize;
+  CONST CHAR8                   *Slice;
+  PE_COFF_LOADER_IMAGE_CONTEXT  Context;
+  RETURN_STATUS                 ContextStatus;
```
mikebeaton, line 83: Possibly better to use EFI_STATUS type in EFI context? (I understand EFI_STATUS is superset of RETURN_STATUS.)

mhaeuser: But why? For the most part, there is nothing EFI about image processing.

Include/Acidanthera/Library/OcPeCoffExtLib.h
mhaeuser: I think it’s ok, otherwise there only other option is a User-specific header. Declaring the function in the consumer code should not be done, as not all compilers will notice de-sync between prototypes without a shared header.

Utilities/AppleEfiSignTool/AppleEfiSignTool.c
mikebeaton: @mhaeuser - Well, because this very RETURN_STATUS is processed as EFI_STATUS in existing OC code at https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcPeCoffExtLib/OcPeCoffExtLib.c#L394 and https://github.com/acidanthera/OpenCorePkg/blob/master/Library/OcPeCoffExtLib/OcPeCoffExtLib.c#L483, and although processing RETURN_STATUS as RETURN_STATUS seemed more sane to me, I wanted to check I wasn't missing anything!

Include/Acidanthera/Library/OcPeCoffExtLib.h
mikebeaton: > as not all compilers will notice de-sync between prototypes without a shared header

Yes

PR #504, opened 2023-11-16
OcBootManagementLib: Store ImageLoaderCaps in protocol.

mikebeaton:
The recently added "OCB: load/start unsupported ordering" error log message is unfortunately being hit on real macs when loading recovery.

This means the assumption of the original code (i.e. before the recent changes) that LoadImage and StartImage are nicely paired, at least in the cases we care about, never was being met. Storage of image caps, to transfer them between LoadImage and StartImage when needed, is now moved from a shared static variable to a protocol instance on the image handle.

Tested that it resolves the issue.

Additionally, UnloadImage was never being called when StartImage fails. Now added.

### DIFF
--- a/Library/OcBootManagementLib/BootEntryManagement.c
+++ b/Library/OcBootManagementLib/BootEntryManagement.c
@@ -2548,6 +2548,10 @@ OcLoadBootEntry (
       // Unload dmg if any.
       //
       InternalUnloadDmg (&DmgLoadContext);
+      //
+      // Unload image.
+      //
+      gBS->UnloadImage (EntryHandle);
     }
   } else {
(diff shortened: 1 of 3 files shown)


Library/OcBootManagementLib/BootEntryManagement.c
```diff
@@ -2548,6 +2548,10 @@ OcLoadBootEntry (
       // Unload dmg if any.
       //
       InternalUnloadDmg (&DmgLoadContext);
+      //
+      // Unload image.
+      //
+      gBS->UnloadImage (EntryHandle);
```
mikebeaton, line 2554: Not sure if this gBS->UnloadImage call needs further wrapping, e.g. make a new Context->UnloadImage, or if that's only worth adding if there's anything special for it to do, above and beyond just the gBS call.

PR #507, opened 2023-11-24
User: Update PcdImageProtectionPolicy to match 8c3970100da6e217d959289a908722285eda1531

mikebeaton:
Think this update should be non-controversial, but also wanted to ask:

### DIFF
--- a/User/Library/UserPcd.c
+++ b/User/Library/UserPcd.c
@@ -50,4 +50,4 @@ BOOLEAN  _gPcd_FeatureFlag_PcdFatReadOnlyMode                        = _PCD_VALU
 UINT32   _gPcd_BinaryPatch_PcdSerialRegisterStride                   = 0;
 UINT8    _gPcd_FixedAtBuild_PcdUefiImageFormatSupportNonFv           = 0x00;
 UINT8    _gPcd_FixedAtBuild_PcdUefiImageFormatSupportFv              = 0x03;
-UINT32   _gPcd_FixedAtBuild_PcdImageProtectionPolicy                 = 0x00;
+UINT32   _gPcd_FixedAtBuild_PcdImageProtectionPolicy                 = 0x00000003;


User/Library/UserPcd.c
```diff
@@ -50,4 +50,4 @@ BOOLEAN  _gPcd_FeatureFlag_PcdFatReadOnlyMode                        = _PCD_VALU
 UINT32   _gPcd_BinaryPatch_PcdSerialRegisterStride                   = 0;
 UINT8    _gPcd_FixedAtBuild_PcdUefiImageFormatSupportNonFv           = 0x00;
```
mikebeaton, line 51: Should this line be modified to match the default (0x02) which I believe is used unmodified in OC.

mikebeaton:
Should this https://github.com/acidanthera/OpenCorePkg/blob/master/Utilities/TestPeCoff/PeCoff.c#L203 also be modified to `UefiImageOriginUserImage`?

(Perhaps in an ideal world, both the PcdImageProtectionPolicy value and this `UefiImageOrigin...` value would come from LoadConfig()?)

User/Library/UserPcd.c
```diff
@@ -50,4 +50,4 @@ BOOLEAN  _gPcd_FeatureFlag_PcdFatReadOnlyMode                        = _PCD_VALU
...
 UINT8    _gPcd_FixedAtBuild_PcdUefiImageFormatSupportFv              = 0x03;
```
MikhailKrichanov, line 52: These PCDs are primarily used for BaseTools construction and [determine](https://github.com/acidanthera/audk/blob/master/MdePkg/Library/BaseUefiImageLib/UefiImageLib.c#L31-L44) supported formats (PE/UE). Because for them [`UEFI_IMAGE_FORMAT_SUPPORT_SOURCES=0x02`](https://github.com/acidanthera/audk/blob/master/BaseTools/Source/C/Makefiles/header.makefile#L147l), `_gPcd_FixedAtBuild_PcdUefiImageFormatSupportNonFv` will change nothing.

User/Library/UserPcd.c
```diff
@@ -50,4 +50,4 @@ BOOLEAN  _gPcd_FeatureFlag_PcdFatReadOnlyMode                        = _PCD_VALU
...
-UINT32   _gPcd_FixedAtBuild_PcdImageProtectionPolicy                 = 0x00;
```
MikhailKrichanov, line 53: This PCD is also used by BaseTools and such a modification will break [EDK2_REDUCE_FW_SIZE](https://github.com/acidanthera/audk/blob/master/BaseTools/Conf/tools_def.template#L50-L54) builds, because BaseTools will reject PE files with SectionAlignment < 4KBi.

MikhailKrichanov (commented):
Conclusion: reject.

MikhailKrichanov:
> Should this https://github.com/acidanthera/OpenCorePkg/blob/master/Utilities/TestPeCoff/PeCoff.c#L203
> 
> ```
>   Status = PeCoffInitializeContext (&Context, FileBuffer, FileSize, UefiImageOriginFv);
> ```
> 
> also be modified to `UefiImageOriginUserImage`?
> 
> (Perhaps in an ideal world, both `PcdImageProtectionPolicy` and this `UefiImageOrigin...` value would come from `LoadConfig ()`, in this test?)

Regarding fuzzing, it should not really matter.

User/Library/UserPcd.c
mikebeaton: I'm confused, these are the versions of the Pcds used when building userspace tools, which I'm taking in general (unless there's a reason not too, for fuzzing or something else) should match the values used in the main OC build. I'm not clear how these affect or relate to EDK-II/UDK BaseTools?

mikebeaton: (Btw, my specific reason for _wanting_ to change this, was to make AppleEfiSignTool be less surprising, by matching what it does and doesn't accept to what Apple signature checking in live OC accepts.)

User/Library/UserPcd.c
mikebeaton: Same comment as below, I'm trying to ensure that OC userspace tools are unsurprising by matching the Pcd configuration of OC itself - unless in any given case there's a good reason not to.

User/Library/UserPcd.c
MikhailKrichanov: These PCDs are used by [ImageTool](https://github.com/acidanthera/audk/blob/master/BaseTools/ImageTool/GNUmakefile#L26). Maybe it is time to eradicate audk dependency on OpenCore.

mikebeaton: Oh! Yes, there seem to be two quite different purposes here, with just one set of Pcds. Thank you.

vit9696: - We should assign desired values in ImageTool and possibly other tools in OpenCore.
- Then we should update default values here.

For the latter we should decide on the defaults.

* I think supporting only PE and not UE by default is fair.
* Also enabling image protection policy is a good idea, but I vote for all sources (i.e. not just fw, option roms but also user supplied images).

PR #513, opened 2024-01-02
Platform: Add Firmware Settings entry driver

mikebeaton:
Would address some of https://github.com/acidanthera/bugtracker/issues/2209.

Requires labels in https://github.com/acidanthera/OcBinaryData/pull/12, would require audio.


mikebeaton:
Worth noting (cc @CobanRamo) that you can already achieve basically the same with a `Tools` entry using the existing `ResetSystem.efi` tool, e.g. as below.

cc @vit9696 The main (arguable...) advantage to adding a driver is it would force settling on an agreed 'flavour' name for this feature, which might be worth doing if enough authors are still updating icon sets, and enough users will use this feature.

---

```
			<dict>
				<key>Comment</key>
				<string></string>
				<key>RealPath</key>
				<false/>
				<key>Name</key>
				<string>UEFI Firmware Settings</string>
				<key>TextMode</key>
				<false/>
				<key>Enabled</key>
				<true/>
				<key>Arguments</key>
				<string>firmware</string>
				<key>Flavour</key>
				<string>FirmwareSettings</string>
				<key>Path</key>
				<string>ResetSystem.efi</string>
				<key>Auxiliary</key>
				<true/>
				<key>FullNvramAccess</key>
				<false/>
			</dict>
```

vit9696:
Design-wise this looks good to me.

CobanRamo:
I tested both versions, works great. thanks Mike...

OcTools

mikebeaton:
> Design-wise this looks good to me.

Will we go ahead? If so, could @Andrey1970AppleLife add audio for `FirmwareSettings` audio file, saying "Firmware Settings", in supported languages?

PR #515, opened 2024-01-02
BuiltinPicker: Avoid responding to possibly multiple keys pressed before menu is read out, if using audio assist

mikebeaton:
Only just noticed that this happens, I would say it is a definite improvement to avoid this. Due to using a different approach to fetching key presses, Canopy already behaves like this.


mikebeaton:
Suspect the behaviour fixed by this PR may well have been introduced in the first-place in the keyboard handling updates of a couple of years ago, though would need to go back and double-check to be sure.

PR #516, opened 2024-01-02
OpenCanopy: Fixed intro animation getting stuck...

mikebeaton:
...in OpenCanopy if an entry which returns to the menu is selected before the animation ends.

When triggered, this:
 - Looks wrong
 - Disables ESC and SPACE
 - Can hide the menu completely (if triggered early)

### DIFF
--- a/Platform/OpenCanopy/OcBootstrap.c
+++ b/Platform/OpenCanopy/OcBootstrap.c
@@ -102,8 +102,14 @@ OcShowMenuByOc (
 
   *ChosenBootEntry = NULL;
   OcSetInitialCursorOffset ();
-  mGuiContext.BootEntry            = NULL;
-  mGuiContext.ReadyToBoot          = FALSE;
+  mGuiContext.BootEntry   = NULL;
+  mGuiContext.ReadyToBoot = FALSE;
+  //
+  // Re-run intro animation on each entry into menu, to avoid stuck animation
+  // which happens otherwise, if a menu item which returns to the menu is
+  // selected before the animation ends.
+  //
+  mGuiContext.DoneIntroAnimation   = FALSE;
   mGuiContext.HideAuxiliary        = BootContext->PickerContext->HideAuxiliary;
   mGuiContext.Refresh              = FALSE;
   mGuiContext.PickerContext        = BootContext->PickerContext;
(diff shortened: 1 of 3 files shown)


Platform/OpenCanopy/OcBootstrap.c
```diff
@@ -248,13 +249,6 @@ OcShowPasswordByOc (
 
   mDrawContext.TimeOutSeconds = 0;
 
-  //
-  // Do not play intro animation for blind.
-  //
-  if (Context->PickerAudioAssist) {
-    mGuiContext.DoneIntroAnimation = TRUE;
-  }
-
```
mikebeaton, line 257: mGuiContext.DoneIntroAnimation was never used or checked in the GUI password display (the fade in just happens always).

Platform/OpenCanopy/OcBootstrap.c
```diff
@@ -104,6 +104,7 @@ OcShowMenuByOc (
   OcSetInitialCursorOffset ();
   mGuiContext.BootEntry            = NULL;
   mGuiContext.ReadyToBoot          = FALSE;
+  mGuiContext.DoneIntroAnimation   = FALSE;
```
mikebeaton, line 107: This is the fundamental change, it re-runs the menu intro animation every time back into the menu. This is _possibly_ not ideal (i.e. I understand previously it was only supposed to run the first time in), but a) it looks fine, b) it also works, and this is fiddly to fix!

Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -1772,6 +1772,10 @@ InternalBootPickerAnimateIntro (
       );
   }
 
+  if (CurrentTime == 0) {
+    PrevSine = 0;
+  }
+
```
mikebeaton, line 1778: This is a hack, but having a static PrevSine value is (arguably?) a hack anyway.

Platform/OpenCanopy/OcBootstrap.c
mhaeuser: No comment in the code :(

mikebeaton: This?

mhaeuser: LGTM

PR #519, opened 2024-01-03
OpenCanopy: Provide OC_ATTR_REDUCE_MOTION to disable non-needed OpenCanopy animiation

mikebeaton:
Addresses https://github.com/acidanthera/bugtracker/issues/2040


mikebeaton:
Plan to resolve password fade-in (i.e. should be disabled for visually impaired users; which it never was) in separate PR, will make it obey this too.

mhaeuser:
That option sounds bizarrely specific to me, should it maybe disable all animations? Or all with motion (like iOS’ Reduce Motion)?

mikebeaton:
How about renaming the attr, and having:

| | Password fade-in | Menu fade-in/ease-in | Selected item pulse | Password incorrect animation |
|---|---|---|---|---|
| Normal | On | On | On | On |
| OC_ATTR_REDUCE_MOTION | Off | Off | On | On |
| AudioAssist | Off | Off | Off | Off |

mhaeuser:
Sounds good, tyvm

mikebeaton:
> Sounds good, tyvm

This should be it. I don't know if you want to quickly re-scan for any issues? (Password shake was already disabled on audio assist.)

mhaeuser:
LGTM, tyvm

PR #520, opened 2024-01-18
OcApfsLib: Always load JumpStart APFS directly

mikebeaton:
apfs.efi has W^X errors which require fixup for strict loaders, however we were only passing the post-sanitised image to OC-wrapped platform loader, therefore the wrapper in ImageLoader.c cannot detect that it is an Apple signed binary which should be fixed up.

Fixes issue reported here: https://github.com/acidanthera/bugtracker/issues/2363#issuecomment-1896210279


vit9696 (approved):

Ampernic:
Thank you. Now it finaly booting.

mikebeaton:
> Thank you. Now it finaly booting.

Thank you for the feedback, and for reporting this issue.

PR #522, opened 2024-01-20
OpenDuet: Apply UE loader Pcds which are required to load current Linux EFI stubs

mikebeaton:
Should address https://github.com/acidanthera/bugtracker/issues/2371


vit9696:
This is acceptable for OpenDuet. Please merge.

PR #523, opened 2024-02-27
Support both python 2 and 3

corpnewt:
f-strings are python 3 only.  This commit uses .format() instead which works on both python 2 and 3.  The shebang remains python3 as that's preferred.

### DIFF
--- a/Utilities/macrecovery/macrecovery.py
+++ b/Utilities/macrecovery/macrecovery.py
@@ -58,7 +58,7 @@ def run_query(url, headers, post=None, raw=False):
             return response
         return dict(response.info()), response.read()
     except HTTPError as e:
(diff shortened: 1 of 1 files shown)


vit9696:
Where do we need Python 2? Transition to Python 3 was intended.

corpnewt:
I had worked with a user recently who was intending to download 10.7 on a machine currently running Vista - which they were unable to install python 3 to.  The changes in this PR don't affect python 3's functionality - but don't exclude python 2 support.

vit9696:
Not sure we want to support this really. The change makes the code less readable. Sorry.

corpnewt:
Not a problem.  Thank you for looking at it.

PR #526, opened 2024-03-06
Tweak GHA configs

Goooler:
- Bump GH actions, old ones based on Node 16 are deprecated.
- Use Apple silicon images, this will speed up builds on Mac, see https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source
- Merge steps.
- ~~Switch back to ubuntu-latest, now it's `22.04`.~~
- Reuse env variables.

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -17,7 +17,7 @@ jobs:
     env:
       JOB_TYPE: ANALYZE
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Install Dependencies
         run: |
@@ -65,7 +65,7 @@ jobs:
     env:
       JOB_TYPE: ANALYZE
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Install Dependencies
         run: |
@@ -77,7 +77,7 @@ jobs:
           python3 -m prospector . -P ./profile.yml | tee prospector_result.txt || exit 1
 
       - name: Upload prospector result to Artifacts
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         if: failure()
         with:
           name: Prospector Artifacts
@@ -88,7 +88,7 @@ jobs:
     runs-on: ubuntu-22.04
     if: github.repository_owner == 'acidanthera' && github.ref_name == 'master'
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Install Dependencies
         run: |
@@ -106,7 +106,7 @@ jobs:
       TOOLCHAINS: GCC
     if: github.repository_owner == 'acidanthera' && github.event_name != 'pull_request'
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Install Dependencies
         run: |
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -62,7 +62,7 @@ jobs:
 
   analyze-python-scripts:
     name: Python Scripts
-    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/analyze.yml
```diff
@@ -61,11 +62,11 @@ jobs:
 
   analyze-python-scripts:
     name: Python Scripts
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
```
vit9696, line 65: Please leave 22.04 for stability reasons, we should better upgrade it manually once 24.04 is released and tested.

.github/workflows/build.yml
```diff
@@ -46,13 +43,13 @@ jobs:
...
       - name: Upload to Artifacts
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: macOS XCODE5 Artifacts
           path: Binaries/*.zip
       - name: Upload to Release
         if: github.event_name == 'release'
-        uses: svenstaro/upload-release-action@e74ff71f7d8a4c4745b560a485cc5fdb9b5b999d
+        uses: svenstaro/upload-release-action@v2
```
vit9696, line 52: Please use a commit corresponding to their latest tag (release). We cannot entirely trust third-party repositories (non-GitHub) are not compromised and get their tags overwritten.

vit9696:
The changes look good to me, would you update other repos or is there no big need in them?

.github/workflows/analyze.yml
Goooler: 655d58a24595e868112db1cf603572d10e633b80

.github/workflows/build.yml
Goooler: fa9af0fee6b40c5b5399672c697ab032f968225c

Goooler:
There are so many repos. I believe we can set up @renovate-bot to keep things up to date for them.

mikebeaton:
Hi - Just as a reminder, if we switch to ARM build runner we should probably remove this: https://github.com/acidanthera/OpenCorePkg/blob/master/build_oc.tool#L328-L336

(Or, if we don't remove it, we will be shipping AARCH64 versions of these couple of BaseTools, which is probably less useful and may confuse people.)

Goooler:
It could be addressed in a follow-up.

vit9696:
Ouch, should we really remove that or rather make the tools FAT? In my opinion, should merge after we have a fix for it.

mikebeaton:
Okay, look for better fix subsequently, for the issue I raised, agreed.

mikebeaton:
@Goooler - Right, Vit's suggestion was for me to fix the issue I mentioned, first (which I can do and test on local ARM build, ofc) - and then to merge your PR when that is done. So I will look at that, the plan is to do both after next release. Thank you.

mikebeaton:
I think this should be ready to install, following https://github.com/acidanthera/audk/commit/33a40156214d8cd44f13ae2e7c9fa2b37a6f3ae5 .

Unfortunately the `brew install shellcheck` line from `analyze.yml` workflow is now [failing](https://github.com/acidanthera/OpenCorePkg/actions/runs/8169632268/job/23279227360) on the M1 runner - presumably this will hit the Intel runner in due course too - because we are hitting [this](https://github.com/orgs/Homebrew/discussions/3404).

I am currently [not managing to replicate the error](https://github.com/orgs/Homebrew/discussions/3404#discussioncomment-8965615), in order to fix it, using a local, non-container install of brew, even though brew, python3 and pip3 are all the same versions that are causing this to fail in the M1 runner.

Adding `--break-system-packages` is a possible, but I think strongly non-preferred, work-around.

mikebeaton:
Fixed: https://github.com/acidanthera/OpenCorePkg/commit/ab5aeec67a22438b826fcd9460bd838b6d67551d

PR #527, opened 2024-03-06
Docs: Updates to styling

soyeonswife63:
i tried to clean up/update the README as good as i could!! it's not perfect but i certainly think it's an improvement

im open to any suggestions!! thanks in advance <33

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,7 +5,7 @@ OpenCore Changelog
 - Modified `Launchd.command` to recreate its log file if deleted
 - Updated `Launchd.command` to work with macOS Sonoma (re-run `./Launchd.command install` after upgrading to Sonoma)
 - Fixed an incorrectly labelled MacBookPro11,3 model code in `macserial`, thx @Macschrauber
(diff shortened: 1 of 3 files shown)


soyeonswife63:
there was a typo in the changelog as well, lemme fix that up rq

mikebeaton:
Hi - Thank you for taking the time to do this. I did sound out a second opinion, and on balance we would say that we don't feel we need this 'happy' styling, though again thank you for taking the time and offering it. Corrections to factual inaccuracies or poor English are still welcome of course, and more likely to be accepted. (I had already spotted and flagged up the recent Changelog typo, so I have just gone ahead and fixed that, but again, thanks for the extra pair of eyes.)

soyeonswife63:
> Hi - Thank you for taking the time to do this. I did sound out a second opinion, and on balance we would say that we don't feel we need this 'happy' styling, though again thank you for taking the time and offering it. Corrections to factual inaccuracies or poor English are still welcome of course, and more likely to be accepted. (I had already spotted and flagged up the recent Changelog typo, so I have just gone ahead and fixed that, but again, thanks for the extra pair of eyes.)

aah okay ty ^^;;

PR #532, opened 2024-04-15
OpenDuet: Debug updates

mikebeaton:
For reasons currently unknown, unoptimized Duet will not start in qemu (continually restarts, before any debugging output). However it seems potentially extremely useful to have a version of Duet which can be run in qemu with all debug output streaming to serial, and with ASSERTs enabled, once they can be. It seams reasonable to make NOOPT Duet permanently like this?

### DIFF
--- a/OpenDuetPkg.dsc
+++ b/OpenDuetPkg.dsc
@@ -112,9 +112,16 @@
   OcStringLib|OpenCorePkg/Library/OcStringLib/OcStringLib.inf
   #
   # To save size, use NULL library for DebugLib and ReportStatusCodeLib.
-  # If need status code output, do library instance overriden.
+  # If need status code output, do library instance override.
   #
+!if ($(TARGET) == NOOPT)
+  #
+  # NOOPT equivalent to DEBUG_ON_SERIAL_PORT in OvmfPkg.
+  #
+  DebugLib|OpenCorePkg/Library/OcDebugLibSerial/OcDebugLibSerial.inf
+!else
   DebugLib|OpenCorePkg/Library/OcDebugLibNull/OcDebugLibNull.inf
+!endif
   DebugPrintErrorLevelLib|MdePkg/Library/BaseDebugPrintErrorLevelLib/BaseDebugPrintErrorLevelLib.inf
   ReportStatusCodeLib|MdePkg/Library/BaseReportStatusCodeLibNull/BaseReportStatusCodeLibNull.inf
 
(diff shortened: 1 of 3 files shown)


OpenDuetPkg.dsc
```diff
@@ -256,9 +263,19 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdEmuVariableNvModeEnable|TRUE
   gEfiMdeModulePkgTokenSpaceGuid.PcdMaxVariableSize|0x10000
   gEfiMdeModulePkgTokenSpaceGuid.PcdVariableStoreSize|0x10000
+!if ($(TARGET) == NOOPT)
+  # DEBUG_ASSERT_ENABLED | DEBUG_PRINT_ENABLED | DEBUG_CODE_ENABLED | CLEAR_MEMORY_ENABLED | ASSERT_DEADLOOP_ENABLED
+  ##gEfiMdePkgTokenSpaceGuid.PcdDebugPropertyMask|0x2f
+  # DEBUG_PRINT_ENABLED | DEBUG_CODE_ENABLED | CLEAR_MEMORY_ENABLED | ASSERT_DEADLOOP_ENABLED
+  gEfiMdePkgTokenSpaceGuid.PcdDebugPropertyMask|0x2e
```
mikebeaton, line 270: We currently cannot add DEBUG_ASSERT_ENABLED, because we are hitting an ASSERT - which was presumably happening silently anyway.

build_duet.tool
```diff
@@ -176,7 +176,7 @@ if [ "${INTREE}" != "" ]; then
   imgbuild "${TARGETARCH}"
 else
   if [ "$TARGETS" = "" ]; then
-    TARGETS=(DEBUG RELEASE)
+    TARGETS=(DEBUG RELEASE NOOPT)
```
mikebeaton, line 179: Seems useful to confirm in CI that NOOPT Duet builds.

OpenDuetPkg.dsc
```diff
@@ -279,7 +296,7 @@
   MSFT:DEBUG_*_*_CC_FLAGS    = -D OC_TARGET_RELEASE=1 /FAcs -Dinline=__inline -DMDEPKG_NDEBUG /GS /kernel
   MSFT:RELEASE_*_*_CC_FLAGS  = -D OC_TARGET_RELEASE=1 /FAcs -Dinline=__inline -DMDEPKG_NDEBUG /GS /kernel
 
-  XCODE:NOOPT_*_*_CC_FLAGS   = -D OC_TARGET_RELEASE=1 -fno-unwind-tables -O0 -fstack-protector-strong -ftrivial-auto-var-init=pattern
+  XCODE:NOOPT_*_*_CC_FLAGS   = -D OC_TARGET_RELEASE=1 -fno-unwind-tables -flto -Os -fstack-protector-strong -ftrivial-auto-var-init=pattern
```
MikhailKrichanov, line 299: This change transformed NOOPT into DEBUG. As now there is not difference, NOOPT is useless.

mikebeaton: No, NOOPT is now capable of symbolic debugging, and you can actually start it.

mikebeaton: The current difference with NOOPT, compared to DEBUG, is that all the debug macros are on, also non-null serial debug lib is on.

mikebeaton: Please see https://github.com/acidanthera/bugtracker/issues/2397 - thank you

PR #537, opened 2024-04-24
Docs: Fix alphabetical order on some entries in the Configuration file.

leon9078:
When I was reading the Sample.plist file in Xcode, I noticed that the cascading entries are perfectly ordered in alphabetical order and caps-lock first. Nice.

When I switched to the official Configuration document, however, I found that some entries are not listed like the Sample file, and this made me very confusing, because I had to scroll up/down in the document different times to find the specific entry description. So I fixed the order by Cut/Paste the wrong order sections.

This are the sections involved:
- Misc –> Properties
- Misc –> Boot Properties
- Misc –> Security Properties
- PlatformInfo –> Properties
- PlatformInfo –> Generic Properties
- PlatformInfo –> DataHub Properties
- PlatformInfo –> PlatformNVRAM Properties
- PlatformInfo –> SMBIOS Properties
- UEFI –> AppleInput Properties
- UEFI –> Drivers Properties
- UEFI –> Output Properties
- UEFI –> Quirks Properties
- UEFI –> ReservedMemory Properties

So here is the result, better readability and better sense when reading the whole Configuration document step-by-step.

If everything's fine, can you rebuild the .pdf? Thanks. CC @vit9696 @Andrey1970AppleLife @mikebeaton

### DIFF
--- a/Docs/Configuration.tex
+++ b/Docs/Configuration.tex
@@ -152,6 +152,7 @@ with the full list available in
 \subsection{Generic Terms}\label{generic-terms}
 
 \begin{itemize}
+
 \item
   \texttt{plist} --- Subset of ASCII Property List format written in
   XML, also know as XML plist format version 1. Uniform Type Identifier
@@ -220,6 +221,7 @@ with the full list available in
 \subsection{Configuration Terms}\label{configuration-terms}
 
 \begin{itemize}
+
 \item
   \texttt{OC\ config} --- OpenCore Configuration file in \texttt{plist}
   format named \texttt{config.plist}. It provides an extensible way
@@ -891,6 +893,7 @@ Please note however, that suggested solutions from third parties may be outdated
 \subsection{Properties}\label{acpiprops}
 
 \begin{enumerate}
+
 \item
   \texttt{Add}\\
   \textbf{Type}: \texttt{plist\ array}\\
@@ -931,6 +934,7 @@ Please note however, that suggested solutions from third parties may be outdated
 \subsection{Add Properties}\label{acpipropsadd}
 
 \begin{enumerate}
+
 \item
   \texttt{Comment}\\
   \textbf{Type}: \texttt{plist\ string}\\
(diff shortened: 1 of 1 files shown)


mikebeaton:
Thank you for making the effort to create this - it may take some time to review, as it is easy in such a large change to intentionally or unintentionally (though with no prior reason at all to suspect intentionally!) lose something, move something to the wrong place, move something which changes meaning, etc., etc.

If anyone else who has shown an interest in improving the docs has time to add a review of these changes, it would be welcome.

leon9078:
Thanks in advance! Yeah, I know it's a large change. I hope it could be ready (after the review) for the 1.0.0 May release!

mikebeaton:
```
git diff --color-moved=blocks master..leon -- Docs/Configuration.tex
```

is reassuring that this commit is, indeed, a sequence of carefully made moves. Thank you.

However, it does not currently build.

@leon9078 - Please could you download the dependencies of Docs/BuildDocs.tool (in particular, `mactex`) and resolve this?

leon9078:
> @leon9078 - Please could you download the dependencies of Docs/BuildDocs.tool (in particular, `mactex`) and resolve this?

@mikebeaton  So I tried myself your advice, but I've all the dependencies already installed.

And still, doesn't build for me. So I tried to carefully edit small parts of the whole commit change and rebuild (every time I run: bash ./Docs/BuildDocs.tool to check if it hangs).

It seems there are some issues re-ordering some entries in PlatformInfo –> DataHub Properties, but I really can't understand why... There's some formatting change that could be applied in the Tex document to fix those build issues?!

Again, thanks

mikebeaton:
The issue is pinned down to latexdiff not coping with this large change; in two or three instances it puts several lines of changes, including new paragraphs, inside `texttt`, which is invalid LaTeX. We will fix or work around it.

mikebeaton:
@leon9078 - Thanks for the changes.

Due to latexdiff not quite coping with this large change, the current plan is to manually fix up Differences.tex, which in the end produces a correct Differences.pdf (as was done in https://github.com/acidanthera/OpenCorePkg/commit/2356f787434658c46a1631bbf28dddbf47e84d93), for this and any remaining Configuration.tex changes for this release.

Once PreviousConfiguration.tex is updated, after the version is bumped, automatic handling of smaller changes should be restored!

mikebeaton:
@leon9078 Thank you for having produced this helpful PR 😎

PR #539, opened 2024-04-28
CrScreenshotDxe: Work around TPL-related ASSERT within firmware compiled to assert

mikebeaton:
(such as recently updated DEBUG version of OpenDuet)

### DIFF
--- a/Platform/CrScreenshotDxe/CrScreenshotDxe.c
+++ b/Platform/CrScreenshotDxe/CrScreenshotDxe.c
@@ -166,6 +166,16 @@ TakeScreenshot (
   UINTN                          Index;
   UINT8                          Temp;
 
+  //
+  // This is required to avoid assert (only noticeable on firmware compiled to
+  // assert) from gBS->RaiseTPL(TPL_CALLBACK) within the file system accesses
+  // below. Makes explicit what was happening anyway, which is that we're
+  // effectively lowering the TPL for a long running task, during a keyboard
+  // interrupt.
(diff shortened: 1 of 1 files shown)


vit9696:
cc @NikolajSchlej

vit9696 (approved):

mikebeaton:
In the context of debug build of Ovmf there is no such issue, so needs further investigation as to why.

mikebeaton:
I can confirm that the same issue arises in current EDK-II OvmfPkg. It seems very hard to compile OvmgPkg with ASSERTs enabled and get anywhere - it hits lots of ASSERTs - but you can single step from where FatPkg tries to acquire its lock, and see exactly the same issue.

mikebeaton:
In case it helps to clarify what I am seeing (when debug prints and asserts are enabled in firmware), this visible assert message is:

```
FATAL ERROR - RaiseTpl with OldTpl(0x10) > NewTpl(0x8)
ASSERT [DxeCore] Tpl.c(66): ((BOOLEAN)(0==1))
```

and the call stack after tripping this is:

Screenshot 2024-04-28 at 10 55 24

mikebeaton:
FWIW Issue is confirmed present with OC 0.8.7 and DEBUG OVMF, so is not caused by any recent updates, just showing up because we recently updated OpenDuet DEBUG version to assert - and [fixed all the other asserts](https://github.com/acidanthera/bugtracker/issues/2397) we have found so far.

vit9696 (approved):
Please make a PR in https://github.com/LongSoft/CrScreenshotDxe as well. We need to backport it there.

PR #540, opened 2024-05-15
OcAfterBootCompatLib: Add `FixupAppleEfiImages`‎ quirk to debug output. Also fixed alphabetical order.

leon9078:
When I was reading the OC debug file I found `FixupAppleEfiImages` quirk was missing. Then, I fixed the alphabetical order of some entries, to follow the Sample.plist file.

[opencore-2024-05-15-082116.txt](https://github.com/acidanthera/OpenCorePkg/files/15319718/opencore-2024-05-15-082116.txt)


mikebeaton:
FixupAppleEfiImages is not an after boot compatibility setting, therefore it is not correct to copy it into OC_ABC_SETTINGS.

I don't see that much need for re-ordering the debug output either. (As mentioned in various other tickets, the debug output does not aim to be user-friendly, just to enable debugging.)

PR #542, opened 2024-05-21
macrecovery: Progress bar & cleanup

soyeonswife63:
- Remove leftover Python 2 code
    - There was already a lot of f-strings which were introduced in Python 3.6, which means it wasn't compatible from the start
    - Python 2 was sunsetted 4 years ago, no need to support it at all
- Cleanup code
    - There were many small cleanups, so please check diff
- Add progress bar for downloads
    - It turned out quite nice, I think :) examplegif
    - It uses the `Content-Length` header to do this. If the header isn't available, it will fall back to the old download 
- Switch `RECENT_MAC` model from `iMacPro1,1` to `MacPro7,1`
    - `iMacPro1,1` discontinuation: March 5, 2021
    - `MacPro7,1` discontinuation: June 5, 2023

*(Previous PR was closed because I had pushed around 12 commits T-T)*


Utilities/macrecovery/macrecovery.py
```diff
@@ -205,27 +199,41 @@ def save_image(url, sess, filename='', directory=''):
...
+                break
         size = 0
         while True:
             chunk = response.read(2**20)
             if not chunk:
                 break
             fh.write(chunk)
             size += len(chunk)
-            print(f'\r{size / (2**20)} MBs downloaded...', end='')
+            if totalsize + 1:
```
mikebeaton, line 226: `if totalsize + 1:` possibly unecessarily clever, `if totalsize > 0:` probably preferable (easier to read, for same functionality).

Utilities/macrecovery/macrecovery.py
```diff
@@ -7,28 +7,30 @@
...
     from urllib.request import Request, HTTPError, urlopen
     from urllib.parse import urlparse
 except ImportError:
-    from urllib2 import Request, HTTPError, urlopen
-    from urlparse import urlparse
+    print('ERROR: Python 2 is not supported, please use Python 3')
+    sys.exit(1)
 
 SELF_DIR = os.path.dirname(os.path.realpath(__file__))
+TERMINAL_SIZE = os.get_terminal_size().columns
```
mikebeaton, line 27: Thanks for progress bar work. Imo it looks good. But it does not handle well, a) resizing screen while downloading, nor b) screen too narrow.
Should probably refetch terminal size every time it needs it, should also revert to old method below a certain terminal width (min. width at which progress bar works).

mikebeaton:
@soyeonswife63 - I hope my suggested commit handles window resizing stably. It's a bit fiddly to stop it looking horrible, so by the time I'd convinced myself it was possibly to get it working cleanly, I thought I might as well add the commit! What do you think?

vit9696:
This is a nice one, thank you :-)

soyeonswife63:
> > Thanks for progress bar work. Imo it looks good. But it does not handle well, a) resizing screen while downloading, nor b) screen too narrow.
> 
> @soyeonswife63 - I think my suggested commit handles window resizing stably. It's a bit fiddly to stop it looking horrible when resizing, so by the time I'd convinced myself it was possible to get it working cleanly as I asked you to, I thought I might as well add the commit! What do you think?

the changes look very nice! thank you for taking a look and fixing up some stuff ^^ i totally forgot about resizing

PR #544, opened 2024-06-01
Add space whenever downloading a new file.

AdityaGarg8:
It makes reading the progress easier.

### DIFF
--- a/Utilities/macrecovery/macrecovery.py
+++ b/Utilities/macrecovery/macrecovery.py
@@ -205,7 +205,7 @@ def save_image(url, sess, filename='', directory=''):
     if filename.find(os.sep) >= 0 or filename == '':
         raise RuntimeError('Invalid save path ' + filename)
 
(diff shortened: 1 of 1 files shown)


mikebeaton:
Thanks for taking the time and offering the suggestion, but IMO this change isn't needed.

AdityaGarg8:
> Thanks for taking the time and offering the suggestion, but IMO this change isn't needed.

Hi @mikebeaton 

Sorry for 2 more commits after this comment. I saw it late.

Anyways, I do agree its not a very required thing, but it seems to make the output looks a bit better and less cluttered:

**Before: **

image


**After: **

Screenshot from 2024-06-01 15-25-52

mikebeaton:
Two or three pairs of eyes have already looked at this recently and felt it was okay. Note the previous update to this tool was not just cosmetic, but involved a fair amount of research and careful work, and functional improvements.

PR #546, opened 2024-06-16
Go back to macos-latest

Goooler:
Now macos-latest is moved to Arm64.

See https://github.com/actions/runner-images?tab=readme-ov-file#available-images.

Follow up #526.


mikebeaton:
Tyvm

PR #548, opened 2024-07-07
OcBootManagementLib: Revert changes to fat slice secure boot logic from 125d8e8df36bd7ca12085d9670dca9060cf49c76

mikebeaton:
Note: EFI_SUCCESS does not always mean a signature was found and verified (which cannot happen for a fat slice), but simply that no further verification of SB status is required and OC image loader should be used.

Resolves: https://github.com/acidanthera/bugtracker/issues/2414

### DIFF
--- a/Library/OcBootManagementLib/ImageLoader.c
+++ b/Library/OcBootManagementLib/ImageLoader.c
@@ -969,7 +969,7 @@ InternalEfiLoadImage (
   // Load the image ourselves in secure boot mode.
   //
   if (SecureBootStatus == EFI_SUCCESS) {
-    if ((SourceBuffer != NULL) && (OcImageLoaderCaps == NULL)) {
+    if (SourceBuffer != NULL) {
       Status = OcImageLoaderLoad (
                  FALSE,
                  ParentImageHandle,
@@ -980,10 +980,16 @@ InternalEfiLoadImage (
                  );
     } else {
       //
-      // We verified the image, but contained garbage, or we are trying to secure boot a Fat slice.
-      // This should not happen.
+      // SecureBootStatus can be EFI_SUCCESS without a signature being
+      // verified, if SecureBootModel is Disabled and the boot file is from
+      // a DMG. This effectively just means: no more SB checks required, use
+      // our loader.
+      // SourceBuffer will be NULL if filtering for supported architecture
+      // failed.
+      // All this can't happen in a secure boot chain, and should not happen
+      // (with valid boot files) in a non-secure boot chain. But if it does,
+      // just abort.
       //
-      ASSERT (FALSE);
       Status = EFI_UNSUPPORTED;
     }
   } else {
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,7 @@ OpenCore Changelog
 - Added support for manual build of i386/x64 10.6+ versions of userspace tools via `FATBIN32=1 make`
 - Disabled `XcpmExtraMsrs MSR_MISC_PWR_MGMT` patch on macOS 12+ due to non-existence
 - Fixed `ThirdPartyDrives` quirk on macOS 14.4 and above
+- Resolved issue booting recovery for OS X 10.8 and earlier since 0.9.7
 
 #### v1.0.0
 - Updated builtin firmware versions for SMBIOS and the rest


Library/OcBootManagementLib/ImageLoader.c
```diff
@@ -969,7 +969,7 @@ InternalEfiLoadImage (
   // Load the image ourselves in secure boot mode.
   //
   if (SecureBootStatus == EFI_SUCCESS) {
-    if ((SourceBuffer != NULL) && (OcImageLoaderCaps == NULL)) {
+    if (SourceBuffer != NULL) {
```
vit9696, line 972: Do you remember what was the purpose for `OcImageLoaderCaps == NULL` check?

Library/OcBootManagementLib/ImageLoader.c
```diff
@@ -980,10 +980,9 @@ InternalEfiLoadImage (
                  );
     } else {
       //
-      // We verified the image, but contained garbage, or we are trying to secure boot a Fat slice.
-      // This should not happen.
+      // We verified the image, but contained garbage.
+      // This should not happen, just abort.
       //
-      ASSERT (FALSE);
```
vit9696, line 986: Why did you remove the `ASSERT`? Is this code reachable from now on?

Library/OcBootManagementLib/ImageLoader.c
mikebeaton: Yes. Go to the commit I'm (proposing) reverting a small part of: https://github.com/acidanthera/OpenCorePkg/commit/125d8e8df36bd7ca12085d9670dca9060cf49c76

And forward search in the page for 'fat slice'. This leads to changes I made (see lines 944 and 958 of the revised ImageLoader.c file in the result of that diff). I am reverting to your original code.

I believe I added the fat slice check because it seemed to me (wrongly, it now turns out) that there was no way that a fat slice could sanely have success in secure boot status (since they date from before signing).

Library/OcBootManagementLib/ImageLoader.c
mikebeaton: I am removing the assert because I added it - I don't think would do any harm to keep it, however. I can say with fair certainty that I added it so that if my added check ([see above](https://github.com/acidanthera/OpenCorePkg/pull/548#discussion_r1668144594)) was wrong, it would get flagged up sooner rather than later and be easy to identify...

Library/OcBootManagementLib/ImageLoader.c
vit9696: Got it, ok.

Library/OcBootManagementLib/ImageLoader.c
vit9696: Just perhaps if it is an invariant, we should replace `Status` assignment with a `CpuDeadLoop()` call and keep the assert. If it is not, then your change makes sense.

mikebeaton: I don't think it is an invariant - we could get get to here without really verifying the file (when it's a dmg, as we are seeing in this issue), but then the file could generate an error when extracting a fat slice.

vit9696: Ok

Library/OcBootManagementLib/ImageLoader.c
```diff
@@ -980,10 +980,9 @@ InternalEfiLoadImage (
                  );
     } else {
       //
-      // We verified the image, but contained garbage, or we are trying to secure boot a Fat slice.
-      // This should not happen.
+      // We verified the image, but contained garbage.
+      // This should not happen, just abort.
```
vit9696, line 984: Reword this comment that this can happen when a trusted binary failed to load for some reason.

mikebeaton: I agree, was just thinking up some wording.

Library/OcBootManagementLib/ImageLoader.c
```diff
@@ -980,10 +980,16 @@ InternalEfiLoadImage (
...
-      // This should not happen.
+      // SecureBootStatus can be EFI_SUCCESS without a signature being
+      // verified, if SecureBootModel is Disabled and the boot file is from
+      // a DMG. This effectively just means: no more SB checks required, use
+      // our loader.
+      // SourceBuffer will be NULL if filtering for supported architecture
+      // failed.
+      // All this can't happen in a secure boot chain, and should not happen
+      // (with valid boot files) in a non-secure boot chain. But if it does,
+      // just abort.
```
mikebeaton, line 991: Is this about right?

vit9696: Yes

PR #549, opened 2024-08-06
OpenNetworkBoot: Add PXE and HTTP(S) Boot support

mikebeaton:
Addresses https://github.com/acidanthera/bugtracker/issues/2386.

Supports launching PXE and HTTP(S) boot options from the OpenCore menu if these are supported by the underlying firmware, or by additional drivers loaded by OpenCore. Additionally supports loading `.dmg` files and their associated `.chunklist` files over HTTP(S) Boot, allowing macOS Recovery to be started over HTTP(S) Boot.

This includes some changes which could be separate commits:

 - cc @Goldfish64 I have changed EXTERNAL_SYSTEM, ExternalSystem, etc. naming from https://github.com/acidanthera/OpenCorePkg/commit/bb44e89addb6cfa8141c483537b1a2eb224cc062 to UNMANAGED_BOOT, UnmanagedBoot, etc. - I was finding it hard to keep track of external tool, external OS and external system in our code, and I believe (hope) the suggested new naming accurately captures what that code (which I was adding to and modifying for HTTP/PXE boot) is for - so just wanted to check that this change looks okay to you!
 - cc @savvamitrofanov I have updated the work done in https://github.com/acidanthera/OpenCorePkg/commit/7e69946021c8d8662dc9fb55b51ca90b7b84ce60 to add IPv6 and HTTPS support
 - I have added a `UEFI/Unload` config option to unload existing firmware drivers - because it was necessary in a few cases to unload an existing firmware driver before loading the standard EDK 2 one provided with OC (since e.g. vendor HttpBootDxe makes ours think it is already started, but vendor driver is trying to use vendor UI which we do not have, therefore causes only non-working (blank screen) HTTP Boot options; and/or can be that vendor HttpDxe is locked to https:// but user wants to use http://)


OpenCorePkg.dsc
```diff
@@ -375,8 +386,10 @@
   #
   !include NetworkPkg/NetworkComponents.dsc.inc
 
+  CryptoPkg/Library/IntrinsicLib/IntrinsicLib.inf
+
 [LibraryClasses]
-  NULL|OpenCorePkg/Library/OcCompilerIntrinsicsLib/OcCompilerIntrinsicsLib.inf
+  NULL|CryptoPkg/Library/IntrinsicLib/IntrinsicLib.inf
```
mikebeaton, line 392: cc @savvamitrofanov @vit9696 - I believe it is very important to support HTTPS boot as well as HTTP boot, therefore we need to also ship TlsDxe. However this depends on CryptoPkg/Library/IntrinsicLib/IntrinsicLib.inf. We cannot just specify this library in the `!if $(NETWORK_TLS_ENABLE) == TRUE` code block above, because it conflicts without our library (defines some of the same externs), and our intrinsics library is set to be added to everything.

I have currently changed to use the CryptoPkg intrinsics library for everything, instead. All seems to work fine, AFAICT, but might perhaps be running slower than necessary in order (e.g.) to be making constant time guarantees, or something. Or might be other negatives I am not aware of.

We could try to update our intrinsics library to add the couple of externs which are missing for it to link with CryptoPkg, but suspect this would be a bad idea (for same reason, i.e. CryptoPkg library might be addressing e.g. constant time security concerns which we are not otherwise bothered with).

I could otherwise update to adding our library manually, wherever it is actually needed, rather than globally; so that CrypoPkg library can be used just in CryptoPkh, without conflicting with ours.

mikebeaton:
@vit9696 - I have not (yet) implemented what we discussed, about loading a DMG file directly into an OC-format DMG ramdisk. I think this could be done by adding a protocol of ours, which would be understood by a new custom OpenHttpBootDxe.

Currently a DMG would be loaded into contiguous memory by HttpBootDxe.efi, then transferred from there to a not-necessarily-contiguous OC DMG ramdisk. Obviously, and as discussed, reading direct to a not-necessarily-contiguous ramdisk of the right size will work faster, and on machines with considerably less RAM.

This would be something to be added later.

Platform/OpenNetworkBoot/README.md
```diff
@@ -0,0 +1,396 @@
...
+filename is `{filename}.dmg` or `{filename}.chunklist` then the other
+file of this pair will be automatically loaded, in order to allow DMG
+chunklist verification, and both files will be used for OpenCore DMG booting.
+
+### MIME types
+
+In order for `.dmg` and `.chunklist` files to be loaded by standard HTTP boot
+drivers (including the `HttpBootDxe` driver currently shipped with OpenCore),
+they must be served with the MIME type `application/efi`, otherwise they will
+be treated as having an unknown file type and will not be loaded.
```
mikebeaton, line 261: If we are creating a custom OpenHttpBootDxe (i.e. to support a new protocol, allowing loading DMGs directly into the required OC-format DMG ramdisk), then at the same time we might as well implement support for recognising `.dmg` and `.chunklist` files directly (rather than requiring users to faff around setting not-really-correct MIME types to serve them).

mikebeaton: It's occurred to me that since I already trap an HTTP callback, it may be possible to forcibly fake the MIME type that HttpBootDxe sees for these files. In which case these files could be recognised by extension, and no special MIME type would need to be set on the web server, even for standard HttpBootDxe. I will look into this...

mikebeaton:
Will need to update to take account of https://github.com/acidanthera/bugtracker/issues/2421

Library/OcMainLib/OpenCoreUefiUnloadDrivers.c
```diff
@@ -0,0 +1,414 @@
...
+                    &gEfiLoadedImageProtocolGuid,
+                    (VOID **)&LoadedImage
+                    );
+    if (!EFI_ERROR (Status)) {
+      FileName = NULL;
+      Name     = GetStringNameFromHandle (HandleBuffer[Index], NULL); ///< Do not free this one.
+      if (Name == NULL) {
+        FileName = FindLoadedImageFileName (LoadedImage);
+        Name     = FileName;
+      }
```
mikebeaton, line 303: The choice of driver name here (to list in SysReport and use as the name to unload by) was prompted by the cMP5,1 (and I believe other similar era macs), where the long English component name is the only name available.

It is too slow to check every possible name, so we have to give priority to one or other name. I am currently giving the long English name priority, as I thought that was available everywhere - but it turns out that on many systems, the long component name is available of most but not all drivers.

Therefore it might make more sense to use the short image file name (which typically corresponds to the driver name, e.g. `HttpDxe`) as 'the' name, only falling back to the long name on systems where the driver name isn't available.

Platform/OpenNetworkBoot/README.md
mikebeaton: Have implemented via callback, 'fake' MIME type no longer needed.

mikebeaton:
> Will need to update to take account of [acidanthera/bugtracker#2421](https://github.com/acidanthera/bugtracker/issues/2421)

Done and committed here. See also https://github.com/acidanthera/audk/pull/66 .

OpenCorePkg.dsc
```diff
@@ -160,7 +160,7 @@
   UefiBootServicesTableLib|MdePkg/Library/UefiBootServicesTableLib/UefiBootServicesTableLib.inf
   UefiBootManagerLib|MdeModulePkg/Library/UefiBootManagerLib/UefiBootManagerLib.inf
   UefiDriverEntryPoint|OpenCorePkg/Library/OcDriverEntryPoint/UefiDriverEntryPoint.inf
-  UefiHiiServicesLib|MdeModulePkg/Library/UefiHiiServicesLib/UefiHiiServicesLib.inf
+  UefiHiiServicesLib|OpenCorePkg/Library/OcHiiServicesLib/OcHiiServicesLib.inf
```
mikebeaton, line 163: Suspect this is another change which makes sense anyway. Resolves same issue as in 13a1c62446a9bea66a4ff11771f166b7ee31464f - namely that the EDK 2 HiiServicesLib asserts when officially _optional_ protocols are not present.

mikebeaton:
Resolved various dependencies (HiiDatabase; plus another new edk2-stable202405 requirement, on Hash2 protocol) which were stopping HTTP Boot working on MacPro5,1. So now gets much further, and visibly starts HTTP Boot, but still won't actually load anything. Returns eventually, so not sure why not. Not sure if this will be easy to resolve or not. Also not quite sure whether to make OpenNetworkBoot Staging or Platform? Feel like it's useful (if you want network boot) and stable on non-Apple hardware, but a shame it won't currently work on Apple.

OpenCorePkg.dsc
mikebeaton: As discussed, will try modifying so that everything uses more or less our intrinsics lib, however updated with missing required functions, and moved as core library to MdePkg in audk.

mikebeaton:
Also as discussed, I will convert the various separable supporting changes here to separate PRs.

mikebeaton:
Closing in favour of separated commits https://github.com/acidanthera/OpenCorePkg/pull/552 https://github.com/acidanthera/OpenCorePkg/pull/553 https://github.com/acidanthera/OpenCorePkg/pull/554

PR #550, opened 2024-08-06
macrecovery: Fix ValueError

mkorje:
When running the script in a builder, I was getting the error that `Image verification failed. (Sign not allowed in string format specifier)`. If `os.get_terminal_size().columns` is less than 2 then the align left padding has a negative value, giving the error `ValueError: Sign not allowed in string format specifier`. I just added a check before the value's use to ensure that it isn't negative.


mikebeaton:
Thank you for pointing out the issue when running headless.

Can you confirm that it works fine for you with my suggested additional commit?

PR #552, opened 2024-09-02
OcBootManagementLib: Convert naming from External System Boot to Unmanaged Boot

mikebeaton:
External is already used for:

 - External picker (as opposed to builtin)
 - External file system (as in, which icon to show in Canopy, or whether to mark entry “ (external)” in builtin picker)
 - External Tool (user defined tool)
 - External OS (user defined entry; it is also used for entries from boot entry protocol*)
 - External Boot System

This PR suggests changing the last one to Unmanaged Boot This is a PR split off from https://github.com/acidanthera/OpenCorePkg/pull/549. I found it easier to work on that after these changes, which is why it was included in there in the first place.

**EDIT**: See additional explanation in comment below.

(*I only spotted this starred point while separating out this PR. I wonder if these should really have been Other OS instead of External OS all along? Like non-macOS, non-Windows blessed entries?)

### DIFF
--- a/Platform/OpenCanopy/Views/BootPicker.c
+++ b/Platform/OpenCanopy/Views/BootPicker.c
@@ -1495,7 +1495,7 @@ BootPickerEntriesSet (
         }
 
         break;
-      case OC_BOOT_EXTERNAL_SYSTEM:
+      case OC_BOOT_UNMANAGED:
         if (OcAsciiStriStr (Entry->Flavour, OC_FLAVOUR_WINDOWS) != NULL) {
           Status = CopyLabel (&VolumeEntry->Label, &GuiContext->Labels[LABEL_WINDOWS]);
         } else {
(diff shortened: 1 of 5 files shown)


Platform/OpenCanopy/Views/BootPicker.c
```diff
@@ -1495,7 +1495,7 @@ BootPickerEntriesSet (
         }
 
         break;
-      case OC_BOOT_EXTERNAL_SYSTEM:
+      case OC_BOOT_UNMANAGED:
```
mikebeaton, line 1498: Note that OC_BOOT_EXTERNAL_TOOL and OC_BOOT_SYSTEM are present a few lines above, if you expand upwards. (IMO while the previous name OC_BOOT_EXTERNAL_SYSTEM was similar to both these, the meaning is not similar.)

vit9696: Yes, that feels ok to me.

PR #553, opened 2024-09-02
OcMainLib: Add UEFI/Unload config option to unload existing firmware drivers

mikebeaton:
This is split out from https://github.com/acidanthera/OpenCorePkg/pull/549.

I don't know of any current use for this apart from network boot, but it is definitely useful there; e.g. HttpBootDxe may be present but produce non-working entries when used within the context of OC (e.g. because it is modified to work in a GUI); HttpDxe may be present but locked to https:// only, when the user wishes to use http://.


Library/OcMainLib/OpenCoreUefiUnloadDrivers.c
```diff
@@ -0,0 +1,352 @@
...
+//
+// Perform action for all driver binding protocols which are
+// present on the same handle as loaded image protocol and
+// where the handle has a usable name.
+//
+// Note: In terms of choosing a name to identify the driver, while the loaded image
+// section name is shorter and corresponds better to the driver's file name, and the
+// firmware volume GUID from the loaded image path identifies the driver more uniquely,
+// these are both not always available for a loaded image (across various firmware),
+// whereas the driver component name normally is.
```
mikebeaton, line 211: This is one of two main choice points within this PR: there are various other names that would seem preferable (more compact, arguably more definitive) that the long English language (usually...) component names. But the other names are often not available; component name seems to be widely available (i.e. in almost all firmware).

Library/OcMainLib/OpenCoreUefiUnloadDrivers.c
```diff
@@ -0,0 +1,352 @@
...
+  if (EFI_ERROR (Status)) {
+    return;
+  }
+
+  for (Index = 0; Index < HandleCount; Index++) {
+    Status = gBS->HandleProtocol (
+                    HandleBuffer[Index],
+                    &gEfiLoadedImageProtocolGuid,
+                    (VOID **)&LoadedImage
+                    );
```
mikebeaton, line 242: The other issue is which drivers we try to list (in SysReport) and allow unloading.

We currently only support drivers which have loaded image and driver binding protocol, and a component name. This is only a small subset of the drivers in any given firmware - it does capture the network stack, which is what https://github.com/acidanthera/OpenCorePkg/pull/549 is concerned with, and which seem to be cleanly designed to allow correct unloaded. I believe many/most of the other drivers which this system won't currently list or unload don't support unloading anyway.

mikebeaton:
It is worth adding that when deciding on this approach, I did initially look at trying to disconnect relevant existing drivers (e.g. perhaps triggered via certain optional arguments to this driver).

However it turns out it is very hard - approaching impossible, afaict - to work out which handles need disconnecting. Certainly there is no particular PCI device to find and disconnect, as there is with audio or video. Instead, each driver binds to handles created by previous drivers in the stack, and because of IP 4 vs 6, most drivers also create two driver binding protocol instances as well (one on the loaded image handle and one on a new handle).

I spent quite a lot of time hunting around all this in UEFI Shell trying to figure out what to disconnect and eventually became fairly convinced it was not possible, whereas unloading cleanly removes what was set up by a given driver - assuming it correctly supports unloading, as the network stack drivers do.

Docs/Configuration.tex
```diff
@@ -7560,6 +7560,37 @@ \subsection{Properties}\label{uefiprops}
...
+  \texttt{Drivers} section.
+  This setting is typically only required if a user-provided driver is a variant of
+  an existing system firmware driver, and if the new driver would detect itself
+  as partially loaded, or otherwise fail to operate correctly, if the old
+  driver is not unloaded first.
+
+  \textbf{Warning}: Unloading system firmware drivers is usually not required
+  and not recommended.
+
+  \emph{Note 1}: The list of driver which this option can attempt to unload
```
mhaeuser, line 7580: List of driver**s**

Docs/Configuration.tex
```diff
@@ -7560,6 +7560,37 @@ \subsection{Properties}\label{uefiprops}
...
+  can be found in \texttt{SysReport/Drivers/DriverImageNames.txt}.
+  The relevant name is the driver component name. Drivers are only listed if they
+  implement \texttt{DriverBindingProtocol} and \texttt{LoadedImageProtocol},
```
mhaeuser, line 7583: LoadedImage is installed by DxeCore, no?

Docs/Configuration.tex
```diff
@@ -7560,6 +7560,37 @@ \subsection{Properties}\label{uefiprops}
...
+
+  To be filled with \texttt{plist\ string} entries containing
+  the names of firmware drivers to unload before loading the
+  \texttt{Drivers} section.
+  This setting is typically only required if a user-provided driver is a variant of
+  an existing system firmware driver, and if the new driver would detect itself
+  as partially loaded, or otherwise fail to operate correctly, if the old
+  driver is not unloaded first.
+
+  \textbf{Warning}: Unloading system firmware drivers is usually not required
```
mhaeuser, line 7577: Maybe drop a sentence about buggy drivers? Iirc OpenUsbKbDxe be used on Aptio IV because it’s USB drivers have borked unload.

Library/OcMainLib/OpenCoreUefiUnloadDrivers.c
```diff
@@ -0,0 +1,352 @@
...
+  Status = gBS->OpenProtocol (
+                  TheHandle,
+                  &gEfiComponentName2ProtocolGuid,
+                  (VOID **)&CompNameStruct,
+                  gImageHandle,
+                  NULL,
+                  EFI_OPEN_PROTOCOL_GET_PROTOCOL
+                  );
+  if (!EFI_ERROR (Status)) {
+    BestLang = GetBestLanguageForDriver (CompNameStruct->SupportedLanguages, Language, FALSE);
```
mhaeuser, line 120: Hmm, why not always pass NULL?

Library/OcMainLib/OpenCoreUefiUnloadDrivers.c
mhaeuser: Yes, I don’t think there is much else you can do. The “proper” names are FFS metadata not used for loading iirc.

Library/OcMainLib/OpenCoreUefiUnloadDrivers.c
mhaeuser: Correct - DriverBinding drivers are UEFI drivers and must be able to unload. The rest are DXE drivers and may or may not be unloadable, you cannot really know from the outside. Further, as per your other comment, we cannot really identify them very well anyway.

Docs/Configuration.tex
mikebeaton: Huh? No. It's in MdeModulePkg/Core/Dxe for all loaded images, e.g.: https://github.com/acidanthera/audk/blob/master/MdeModulePkg/Core/Dxe/Image/Image.c#L841-L871

Docs/Configuration.tex
mikebeaton: > ...OpenUsbKbDxe be used..

cannot be used? must be used?

I gather we cannot unload OpenUsbKbDxe on that firmware... but then... we have to/can/must...?

Docs/Configuration.tex
mhaeuser: That *is* DxeCore, so yes? :)

mikebeaton: We have an OC one, for our loader. It's not that one!

Docs/Configuration.tex
mhaeuser: Cannot, whoops. No, the driver we want to replace with OpenUsbKbDxe cannot be unloaded. And no, point is unloading may be very broken and it depends on the driver unloaded whether it works.

Docs/Configuration.tex
mhaeuser: I thought that was only used for booters? And even if not, point is that it’s an environment thing (DxeCore, OC, whatever), not a property of the driver itself.

mikebeaton: I'm going mad - I read OpenCore. Sorry. Okay. So then? Sorry! This was just double-checking?

Docs/Configuration.tex
mikebeaton: So just strengthening this warning. Okay.

mikebeaton: > Cannot, whoops. No, the driver we want to replace with OpenUsbKbDxe cannot be unloaded. And no, point is unloading may be very broken and it depends on the driver unloaded whether it works.

Btw I already found some drivers relevant to network boot on Apple which at least refuse to unload with Unsupported, which is better than borking everything. So i.e. it can be different levels of broken, but yes, certainly can be completely broken. (I was initially trying to avoid this on the basis that it probably wouldn't work half the time - before, see other comment, deciding it probably was out best bet.)

Docs/Configuration.tex
mikebeaton: > I thought that was only used for booters? And even if not, point is that it’s an environment thing (DxeCore, OC, whatever), not a property of the driver itself.

The reason for checking for DXE Core gEfiLoadedImageProtocolGuid is because those are the handles which support gBS->UnloadImage, which is exactly what we want to use to unload these network stack drivers.

mhaeuser: No, you confused me with OcLoadedImage and I confused you back… everything is clear now! :)

Library/OcMainLib/OpenCoreUefiUnloadDrivers.c
mikebeaton: Sorry, for what? Language? If so, in fact we do - but GetStringNameFromHandle is from UefiHandleParsingLib.c, and takes Language - to which I always pass NULL.

mhaeuser: No, for BestLang. Clearly, when the callee returns NULL , it is passed to GetDriverName (so it’s allowed? On the phone, cannot check). I have two concerns: 1) Reducing calls to anything HII where-ever possible and 2) keeping the keys invariant of the system configuration, i.e. if the UEFI language is changed, the driver name keys used in OC should not suddenly become invalid (because the “BestLang” is suddenly French instead of English).

mikebeaton: BestLang definitely cannot be NULL: https://github.com/acidanthera/audk/blob/master/MdePkg/Include/Protocol/ComponentName.h#L40 https://github.com/acidanthera/audk/blob/master/MdePkg/Include/Protocol/ComponentName2.h#L49

GetBestLanguageForDriver, as called, basically chooses the first language which _the driver itself_ lists as available (this list never changes, for a given driver). So this call to GetBestLanguageForDriver is already intended (i.e. by me) not to change if the firmware language is changed (in fact, precisely by me not passing in any specified/desired language in the Language parameter).

mhaeuser: If it cannot be NULL, then why call the function when it is NULL? That's where I drew that conclusion from.

Right, the callee does not retrieve PlatformLang, the caller does - good. I do wonder about the empty string though, maybe we should default to US English in case there are bugs?

mhaeuser: Well, as I said, in case there are bugs - e.g. SupportedLanguages is empty or otherwise borked. But can just wait whether there are any such reports, I was just thinking aloud.

mikebeaton: Also, I didn't write any of this, including the empty string thing. Also it's not calling a protocol even then, it's calling linked MdePkg UefiLib code, so I guess we know there isn't a bug?

mikebeaton: > Well, as I said, in case there are bugs - e.g. SupportedLanguages is empty or otherwise borked. But can just wait whether there are any such reports, I was just thinking aloud.

Sounds like you're more or less okay with this PR then, I think, once I've added something to the docs mentioning that one reason to avoid unloading images, normally, is borked drivers which crash when you try to unload them?

Docs/Configuration.tex
```diff
@@ -7560,6 +7560,40 @@ \subsection{Properties}\label{uefiprops}
...
+  and not recommended. Poorly written drivers may crash when
+  unloaded, or cause subsequent crashes (e.g by allowing themselves to be
+  unloaded even though they have active dependencies). However standard UEFI
+  network stack drivers should unload cleanly.
```
mikebeaton, line 7581: @mhaeuser Hopefully this is about okay?

mhaeuser: Yes, tyvm.

PR #554, opened 2024-09-06
Network boot

mikebeaton:
Assuming the [Unload](https://github.com/acidanthera/OpenCorePkg/pull/553) and [rename](https://github.com/acidanthera/OpenCorePkg/pull/552) changes were merged, this is what the network boot PR would look like.

I haven't changed the approach to intrinsic lib here yet. I'm planning that that can be addressed on a separate PR to audk, with appropriate very small changes here when finalised. (I'm actually wondering whether the CryptoPkg IntrinsicLib might not be the best one to use; and also wondering whether moving it to MdePkg really makes sense - if it's just going to make future rebases on EDK 2 harder? Anyway, I'll make a PR or two to audk as a place to finalise the intrinsic lib issue.) ~There are also a couple of other minor PRs related to the network boot PR that I want to make on audk (e.g. to downgrade some unecessarily verbose logging in the network stack), but they don't stop this working.~ (Done.)


mikebeaton:
Closed in favour of https://github.com/acidanthera/OpenCorePkg/pull/562.

PR #555, opened 2024-09-07
TEST COMMIT: Update OcCompilerIntrinsicsLib so that it can build the network stack

mikebeaton:
@vit9696 - This is a test commit to show my current thinking on the intrinsics library.

This commit has the changes required to update our intrinsics library so that it can build the network stack. Note that we end up importing [several things](https://github.com/acidanthera/OpenCorePkg/blob/intrinsics-test/Library/OcCompilerIntrinsicsLib/OcCompilerIntrinsicsLib.inf#L26-L29) from the CryptoPkg IntrinsicLib (that is, the EDK 2 intrinsics library which I have used entirely instead of ours in the versions so far of the network boot PR).

This makes me wonder whether the CryptoPkg IntrinsicLib might not be the best one to base on (or rather just use as is, since it already builds our packages and the rest of EDK 2 fine) for a shared intrinsics lib, rather than ours, because the CryptoPkg library already covers the imported things and [multiple other cases](https://github.com/acidanthera/audk/blob/master/CryptoPkg/Library/IntrinsicLib/IntrinsicLib.inf#L29-L49). (Albeit that, as you pointed out, the EDK 2 CryptoPkg IntrinsicLib has some bugs which we already [had to fix](https://github.com/acidanthera/audk/commit/9252fc0e1a9050fc07eaeb862c3943bfbbf0e201).)

**EDIT**: *Okay, it seems this commit has only the changes to our intrinsics library necessary to build the network stack on Xcode - as can be seen in the CI, we would need to import yet more of what is already in CryptoPkg (EDK 2) IntrinsicLib to update our intrinsics lib enough to build the network package on all the architectures. Which I think probably just confirms the point above.*

My next step of reasoning, if the CryptoPkg IntrinsicLib _is_ the best intrinsics lib to use, is that possibly moving it to MdePkg is not such a good idea. This is because it is already quite widely used as an intrinsics library all around EDK 2, despite by its name sounding as if it wouldn't be:

```
./CryptoPkg/CryptoPkgMbedTls.dsc:  IntrinsicLib|CryptoPkg/Library/IntrinsicLib/IntrinsicLib.inf
./CryptoPkg/CryptoPkg.dsc:  IntrinsicLib|CryptoPkg/Library/IntrinsicLib/IntrinsicLib.inf
./CryptoPkg/CryptoPkg.dsc:  CryptoPkg/Library/IntrinsicLib/IntrinsicLib.inf
[...]

PR #556, opened 2024-09-07
OpenCorePkg/OpenDuetPkg: Remove non-needed presence of libraries in Components section

mikebeaton:
Spotted while working on https://github.com/acidanthera/OpenCorePkg/pull/555.

It looks like these lines have always been harmless but not needed.


mikebeaton:
I found this https://github.com/acidanthera/audk/blob/master/CryptoPkg/CryptoPkg.dsc#L354-L361:

```
#
# If profile is ALL, then do verification build of all library instances.
#
!if $(CRYPTO_SERVICES) == ALL
[Components]
  #
  # Build verification of all library instances
  #
.
.
.
```

which seems relevant - though I cannot see what is verified except libraries that would not otherwise have been used, or libraries that would have been used and verified anyway.

Searching for ` +[A-Za-z]*/.*Library/` in `*.dsc` shows a lot of instance of libraries (?) in `Components` section. Some are DXE_DRIVER, PEIM, DXE_RUNTIME_DRIVER, etc., but quite a few are BASE, as applies in this change - so am unsure if I'm missing something obvious about why these do something useful.

mikebeaton:
Okay, closing this again, on the general advice that there are 'reasons' for this.... not least allowing testing of libraries which a package provides, but may not consume.

PR #562, opened 2024-10-08
OpenNetworkBoot: Add PXE and HTTP(S) Boot support

mikebeaton:
This replaces #554, now that the previous two changes #552 #553 are merged.

I was surprised how much of network boot is done in the OVMF UI package. Hence needing to take (sometimes with small mods) several bits of that, in the files imported from EDK 2; and then needing to reimplement the rest of it, in the rest of this driver. Doing it that way means it works with existing network boot drivers, up to the top level of the network boot stack; and means it works, full stop, i.e. no need to try to work out from scratch the correct usage patterns for the lower level network libraries, which are very low level (e.g. you can't 'just' issue a GET or a POST in http, very far from it).

The three commits here are: 1. import three edk 2 files; 2. clean them down to just what I kept and modified; 3. everything else, so if you review just the [third commit](https://github.com/acidanthera/OpenCorePkg/pull/562/commits/c293fbf45cd522d3fa91c4c5aa3a25bc6f2b604b), it is much easier to see what I changed from the originals in BmBoot.c, BmBootDescription.c and TlsAuthConfigImpl.c.

~As before, the biggest change for the rest of the project is switching to use the CryptoPkg intrinsics lib. Maybe that change per se could be discussed independent of reviewing the rest of this? (REFS: https://github.com/acidanthera/OpenCorePkg/pull/555; https://github.com/acidanthera/audk/pull/67; https://github.com/tianocore/edk2/pull/6169 - based on these refs, I say a) we should use it, and b) we should leave it where it is (not move to MdePkg) for now.)~ **EDIT**: Now rebased and updated to use new MdePkg IntrinsicLib location.


Platform/OpenNetworkBoot/BmBoot.c
```diff
@@ -12,2653 +12,535 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
-      }
-
-      EfiBootManagerFreeLoadOption (&BootOption);
-    }
-  }
+  EFI_STATUS                Status;
+  EFI_HANDLE                Handle;
+  EFI_BLOCK_IO_PROTOCOL     *BlockIo;
+  VOID                      *Buffer;
+  EFI_DEVICE_PATH_PROTOCOL  *TempDevicePath;
```
mikebeaton: The diff for this file (here and onwards), at this second commit, is very misleading; this commit is _only_ simple deletes of lines, as can be seen e.g. with `git diff --minimal network-boot~2 network-boot~` at the command line.

mikebeaton:
Rebased and updated to use new MdePkg IntrinsicLib location.

Is there any progress reviewing or testing this?

mikebeaton:
https://www.youtube.com/watch?v=xo_Pb-3thS4

mikebeaton:
@vit9696 @MikhailKrichanov - Updated the PR so that we now ship the three network stack drivers which are required only for PXE Boot (not for HTTP(S) Boot), including the driver UefiPxeBcDxe which I missed out before in the PXE Boot drivers section of the README. Apologies for omitting that, before today I never tested PXE Boot with only our network stack, since up to now we only shipped the drivers required for HTTP(S) Boot.

MikhailKrichanov:
@vit9696 would also like you to add to OpenCore CI tests for PXE and HTTP(S) Boot as a separate pull request.

mikebeaton:
> To date I tested only PXE Boot (works fine in QEMU), which needs not only `UefiPxeBcDxe` but also `VirtioNetDxe` or some other network card driver. `VirtioNetDxe` is worth mentioning in README and providing links to OEM sites, where network card drivers can be downloaded also seems to be a good idea.

I missed `UefiPxeBcDxe` because the [original network stack commit](https://github.com/acidanthera/OpenCorePkg/commit/7e69946021c8d8662dc9fb55b51ca90b7b84ce60) didn't include drivers for PXE boot at all, only HTTP boot - so up to the recent [update](https://github.com/acidanthera/OpenCorePkg/pull/562?notification_referrer_id=NT_kwDOALZKbbQxMjc2MzE4MTMzNDoxMTk0NjYwNQ#issuecomment-2481356637) to this PR above, I assumed we were only aiming at HTTP boot. Now added.

Rather similarly, the standard EDK II and audk builds of OvmfPkg include `VirtioNetDxe` even if `NETWORK_ENABLE` compilation flag is 0, but I believe you are using a different, more cut down build. It is probably a mistake on their part to include it in this case, I might even look at doing a PR, but it explains why I missed it. SNP seems to be what this installs, and presumably is what the hardware device drivers would install.

On the hardware side, a similar issue is that almost all real hardware already includes at least PXE boot ... unless it's using special, cut-down, firmware... - so again I never encountered this, as the relevant hardware drivers are almost always present on most normal machines. But again, I do agree it should be mentioned. Do you have any links to relevant h/ware device drivers?

MikhailKrichanov:
> On the hardware side, a similar issue is that almost all real hardware already includes at least PXE boot ... unless it's using special, cut-down, firmware... - so again I never encountered this, as the relevant hardware drivers are almost always present on most normal machines. But ofc I agree it should be mentioned. Do you have any links to relevant h/ware device drivers?

1. https://winraid.level1techs.com/t/efi-lan-bios-intel-gopdriver-modules/33948
2. https://www.intel.com/content/www/us/en/download/15755/intel-ethernet-connections-boot-utility-preboot-images-and-efi-drivers.html

mikebeaton:
> > On the hardware side, a similar issue is that almost all real hardware already includes at least PXE boot ... unless it's using special, cut-down, firmware... - so again I never encountered this, as the relevant hardware drivers are almost always present on most normal machines. But ofc I agree it should be mentioned. Do you have any links to relevant h/ware device drivers?
> 
> 1. https://winraid.level1techs.com/t/efi-lan-bios-intel-gopdriver-modules/33948
> 2. https://www.intel.com/content/www/us/en/download/15755/intel-ethernet-connections-boot-utility-preboot-images-and-efi-drivers.html

I have added those links to the README:
https://github.com/acidanthera/OpenCorePkg/tree/network-boot/Platform/OpenNetworkBoot#firmware-and-option-roms

MizhiDR:
> > > 在硬件方面，类似的问题是几乎所有实际硬件都已至少包含 PXE 启动...除非它使用特殊的、精简的固件... - 所以我再一次从未遇到过这种情况，因为相关的硬件驱动程序几乎总是存在于大多数普通机器上。但我当然同意应该提到这一点。您有相关硬件设备驱动程序的链接吗？
> > 
> > 
> > 
> > 1. https://winraid.level1techs.com/t/efi-lan-bios-intel-gopdriver-modules/33948
> > 2. https://www.intel.com/content/www/us/en/download/15755/intel-ethernet-connections-boot-utility-preboot-images-and-efi-drivers.html
> 
> 我已将这些链接添加到 README： [https ://github.com/acidanthera/OpenCorePkg/tree/network-boot/Platform/OpenNetworkBoot#firmware-and-option-roms](https://github.com/acidanthera/OpenCorePkg/tree/network-boot/Platform/OpenNetworkBoot#firmware-and-option-roms)

Have you closed this branch?

PR #566, opened 2024-11-16
Update SSDT-PLUG.dsl

4lvl1l2:


stevezhengshiqi:
Your CPU path is not suitable for everyone. Please read https://github.com/acidanthera/OpenCorePkg/blob/master/Docs/AcpiSamples/Source/SSDT-PLUG.dsl#L4-L10

PR #568, opened 2024-11-24
CreateVault: Fix `sign.command` and update signing docs

mikebeaton:
Fixes https://github.com/acidanthera/bugtracker/issues/2443.

Fix operation of `sign.command` when printable characters occur immediately before `=BEGIN OC VAULT=`. `strings` finds the location of the first printable character in such a sequence. `hexdump` automatically works on 16 byte boundaries, so still finds the correct offset.

Use `BASE_ALIGNAS` to enforce the required alignment, which will not be correct on all builds unless enforced (note alignment is required purely for locating the structure correctly from external script as above, not for reading in C).

Remove struct packing, since structs had better be naturally packed anyway (if not, reading from them without arbitrary-alignment-safe code, as we do, would be undefined behaviour). Add static asserts to confirm correct naturally packed layout.

Update the docs to refer to `sign.command` rather than to include the signing commands explicitly - otherwise we have two places that need to be kept in sync for signing commands, and note that the commands in the two places were already out of sync.


Library/OcMainLib/OpenCoreVault.c
```diff
@@ -14,24 +14,21 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
...
   OC_RSA_PUBLIC_KEY_2048    VaultKey;
   CHAR8                     EndMagic[16];
 } OC_BUILTIN_VAULT_KEY;
 
-#pragma pack(pop)
-
+BASE_ALIGNAS (16)
 STATIC
 OC_BUILTIN_VAULT_KEY
-  mOpenCoreVaultKey = {
```
mikebeaton, line 34: Uncrustify removed these spaces, I think it may be confused by the ALIGNAS.

mikebeaton: Tried various combinations, and I think this is the best we are going to get.

Library/OcMainLib/OpenCoreVault.c
```diff
@@ -44,6 +41,19 @@ OcGetVaultKey (
   UINT32   Index;
   BOOLEAN  AllZero;
 
+  STATIC_ASSERT (
+    sizeof (OC_RSA_PUBLIC_KEY_2048) == sizeof (((OC_RSA_PUBLIC_KEY_2048 *)NULL)->Hdr) + sizeof (((OC_RSA_PUBLIC_KEY_2048 *)NULL)->Data),
+    "OC_RSA_PUBLIC_KEY_2048 is not naturally packed"
+    );
```
mikebeaton, line 47: Hmmm, maybe I should add another(?) assert that the size of this is 528, as expected by the script.

mikebeaton: Okay, actually just added this.

mhaeuser: Actually that should be the only ASSERT. Yours is more “obvious” why that size is chosen, but less readable. :( I think we use only the size constant everywhere.

mikebeaton: Which one are you saying I should keep? (Btw I added the one for 528 because the script uses 528 hard coded - expand down a bit from where I edited the script in the PR.)

mhaeuser: 528

mhaeuser: Yes, sgtm.

PR #571, opened 2024-12-28
read MSR_IA32_TSC_ADJUST only if CPU supports

t0rr3sp3dr0:
This patch addresses a bug that would cause OpenCore to freeze on a Virtualization Framework VM. Now, MSR_IA32_TSC_ADJUST is only read when the CPU supports it.

Fixes: https://github.com/acidanthera/bugtracker/issues/2463


t0rr3sp3dr0:
OpenCore running in a Virtualization Framework VM through [vfkit](https://github.com/crc-org/vfkit):

image

vit9696:
Thank you!

PR #573, opened 2025-02-23
Add ClearTaskSwitchBit booter quirk

Goldfish64:
This quirk is needed to boot macOS 10.7 and older when using a 32-bit kernel on a 64-bit UEFI firmware that makes uses of FPU or SSE instructions in runtime services (such as Hyper-V).


vit9696:
This looks good to me, let's get this merged!

PR #574, opened 2025-03-10
Build: Add drivers for QEMU virtio devices into build

savvamitrofanov:
Last week I spent some time on a task related to testing the network stack, namely, finding scenarios to do it correctly.
The most suitable scenario is to test network boot in a minimal UEFI environment, namely OpenDuet.

As a result, I found that for the virtio devices provided by QEMU to work correctly, their drivers need to be loaded, which are missing from our package.
(those drivers are bundled by default with OVMF, that is why we have no problem with it out of QEMU-seabios boot scenario)

To facilitate automated network boot testing in OpenDuet as part of CI, we could provide the missing drivers along with our bootloader.

### DIFF
--- a/Docs/Sample.plist
+++ b/Docs/Sample.plist
@@ -1535,6 +1535,78 @@
 				<key>Arguments</key>
 				<string></string>
 				<key>Comment</key>
+				<string></string>
+				<key>Enabled</key>
+				<false/>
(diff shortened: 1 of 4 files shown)


mikebeaton (approved):

mikebeaton:
This looks sane and useful to me. Thanks for the update.

PR #575, opened 2025-05-11
ocvalidate: Prevent uint32 roll over if *FixedSize > OC_STORAGE_SAFE_PATH_MAX

corpnewt:
If the `BundlePath` is greater than `OC_STORAGE_SAFE_PATH_MAX`, then subtracting the `*FixedSize` will always result in a would-be negative number, which rolls the print statement's `should not exceed` value over to the uint32 max.

This commit uses `0` if `*FixedSize` is greater than `OC_STORAGE_SAFE_PATH_MAX`, or the difference of the two values.

### DIFF
--- a/Utilities/ocvalidate/ValidateKernel.c
+++ b/Utilities/ocvalidate/ValidateKernel.c
@@ -231,7 +231,7 @@ CheckKernelAdd (
         "Kernel->Add[%u]->ExecutablePath (length %u) is too long (should not exceed %u)!\n",
         Index,
         AsciiStrLen (ExecutablePath),
-        OC_STORAGE_SAFE_PATH_MAX - ExecutableFixedSize
+        ExecutableFixedSize > OC_STORAGE_SAFE_PATH_MAX ? 0 : OC_STORAGE_SAFE_PATH_MAX - ExecutableFixedSize
         ));
       ++ErrorCount;
     }
@@ -247,7 +247,7 @@ CheckKernelAdd (
         "Kernel->Add[%u]->PlistPath (length %u) is too long (should not exceed %u)!\n",
         Index,
         AsciiStrLen (PlistPath),
-        OC_STORAGE_SAFE_PATH_MAX - PlistFixedSize
+        PlistFixedSize > OC_STORAGE_SAFE_PATH_MAX ? 0 : OC_STORAGE_SAFE_PATH_MAX - PlistFixedSize
         ));
       ++ErrorCount;
     }


corpnewt:
It appears to fail to build for the IA32 arch - though I'm not sure that's due to my changes?  Locally, I see the following:
```
In file included from /Users/corpnewt/desktop/opencorepkg/UDK/OpenCorePkg/Library/OcCompressionLib/zlib/adler32.c:8:
/Users/corpnewt/desktop/opencorepkg/UDK/OpenCorePkg/Library/OcCompressionLib/zlib/zutil.h:170:11: error: 'OS_CODE' macro redefined [-Werror,-Wmacro-redefined]
  170 | #  define OS_CODE 19
      |           ^
/Users/corpnewt/desktop/opencorepkg/UDK/OpenCorePkg/Library/OcCompressionLib/zlib/zutil.h:141:11: note: previous definition is here
  141 | #  define OS_CODE  7
      |           ^
1 error generated.
make: *** [/Users/corpnewt/desktop/opencorepkg/UDK/Build/OpenCorePkg/DEBUG_XCODE5/IA32/OpenCorePkg/Library/OcCompressionLib/OcCompressionLib/OUTPUT/zlib/adler32.obj] Error 1


build.py...
 : error 7000: Failed to execute command
	make tbuild [/Users/corpnewt/desktop/opencorepkg/UDK/Build/OpenCorePkg/DEBUG_XCODE5/IA32/OpenCorePkg/Library/OcCompressionLib/OcCompressionLib]


build.py...
 : error F002: Failed to build module
	/Users/corpnewt/desktop/opencorepkg/UDK/OpenCorePkg/Library/OcCompressionLib/OcCompressionLib.inf [IA32, XCODE5, DEBUG]

- Failed -
Build end time: 15:43:59, May.11 2025
Build total time: 00:01:24
```

The [build failure](https://github.com/acidanthera/OpenCorePkg/actions/runs/14959289946/job/42019130877?pr=575) listed in the PR checks above also seems to fail with `adler32` from what I can tell - though I admit that I'm not tremendously adept with C.

If I've missed something, I can certainly go back through things though.

-CorpNewt

corpnewt:
In an effort to better clarify some of the output, I made some changes to the original PR.

`ExecutablePath` and `PlistPath` are not checked if `BundlePath` exceeds the max path length - as they're guaranteed to fail if it does.  I also attempted to clarify that the `ExecutablePath` and `PlistPath` include `BundlePath` and an additional `\` character - though I'm not sure the wording makes the most sense.

Currently, example output for a rejected `BundlePath` at index 6, and rejected `ExecutablePath` and `PlistPath` at index 7 looks as follows:
```
Kernel->Add[6]->BundlePath (length 152) is too long (should not exceed 122)!
Kernel->Add[7]->BundlePath + \ + ExecutablePath (length 136) is too long (should not exceed 122)!
Kernel->Add[7]->BundlePath + \ + PlistPath (length 136) is too long (should not exceed 122)!
```

-CorpNewt

mikebeaton:
Actually, having played around with the issue myself, I'm coming round to your way of thinking. I'll add some comment to your commits.

Utilities/ocvalidate/ValidateKernel.c
```diff
@@ -218,38 +218,40 @@ CheckKernelAdd (
...
-        "Kernel->Add[%u]->ExecutablePath (length %u) is too long (should not exceed %u)!\n",
-        Index,
-        AsciiStrLen (ExecutablePath),
-        OC_STORAGE_SAFE_PATH_MAX - ExecutableFixedSize
-        ));
-      ++ErrorCount;
-    }
+    } else {
+      //
+      // Only check the ExecutablePath and PlistPath if BundlePath did not trigger an error.
```
mikebeaton, line 223: I believe it is nicer to display all the error messages, and your code display them fine without this 'else'. (Of course you are quite correct that the second and third error messages must be there, if the first one is.)

Utilities/ocvalidate/ValidateKernel.c
```diff
@@ -215,39 +215,39 @@ CheckKernelAdd (
...
         ));
       ++ErrorCount;
     }
 
     //
     // There is one missing '\\' after the concatenation of BundlePath and ExecutablePath. Append one.
     //
-    ExecutableFixedSize   = L_STR_LEN (OPEN_CORE_KEXT_PATH) + AsciiStrLen (BundlePath) + 1;
-    ExecutablePathSumSize = ExecutableFixedSize + AsciiStrSize (ExecutablePath);
+    ExecutableFixedSize   = AsciiStrLen (BundlePath) + 1 + AsciiStrLen (ExecutablePath);
```
mikebeaton, line 226: I think `ExecutableFixedSize` is the wrong name here. In the original code it meant "the size of the part of the executable path which can't be avoided" (i.e. "Kexts\" + BundlePath). Also in the OC code Size strictly means the string size including the zero character, and Length means the string length not including the zero character. You're now using the var to hold the user string length (BundlePath + "\"+ ExecutablePath), so I suggest this var should be renamed ExecutableUserLength.

Utilities/ocvalidate/ValidateKernel.c
```diff
@@ -215,39 +215,39 @@ CheckKernelAdd (
...
         ));
       ++ErrorCount;
     }
 
     //
     // There is one missing '\\' after the concatenation of BundlePath and PlistPath. Append one.
     //
-    PlistFixedSize   = L_STR_LEN (OPEN_CORE_KEXT_PATH) + AsciiStrLen (BundlePath) + 1;
-    PlistPathSumSize = PlistFixedSize + AsciiStrSize (PlistPath);
+    PlistFixedSize   = AsciiStrLen (BundlePath) + 1 + AsciiStrLen (PlistPath);
```
mikebeaton, line 242: For the same reason as above, suggest `PlistFixedSize` as now used should be renamed `PlistUserLength`.

Utilities/ocvalidate/ValidateKernel.c
```diff
@@ -215,39 +215,39 @@ CheckKernelAdd (
...
+    PlistPathSumSize = L_STR_LEN (OPEN_CORE_KEXT_PATH) + PlistFixedSize;
```
mikebeaton, line 243: This is actually off by 1. It is a size (number of chars including zero char), or at least the following test makes sense if it is, but you've calculated the length (without zero char). It needs ` + 1`.

Utilities/ocvalidate/ValidateKernel.c
```diff
@@ -215,39 +215,39 @@ CheckKernelAdd (
...
       ++ErrorCount;
     }
 
     //
     // There is one missing '\\' after the concatenation of BundlePath and ExecutablePath. Append one.
     //
-    ExecutableFixedSize   = L_STR_LEN (OPEN_CORE_KEXT_PATH) + AsciiStrLen (BundlePath) + 1;
-    ExecutablePathSumSize = ExecutableFixedSize + AsciiStrSize (ExecutablePath);
+    ExecutableFixedSize   = AsciiStrLen (BundlePath) + 1 + AsciiStrLen (ExecutablePath);
+    ExecutablePathSumSize = L_STR_LEN (OPEN_CORE_KEXT_PATH) + ExecutableFixedSize;
```
mikebeaton, line 227: This is actually off by 1. It is a size (number of chars including zero char), or at least the following test makes sense if it is, but you've calculated the length (without zero char) - as becomes more clear after the suggested variable rename. This needs ` + 1`.

Utilities/ocvalidate/ValidateKernel.c
```diff
@@ -215,39 +215,39 @@ CheckKernelAdd (
...
     if (ExecutablePathSumSize > OC_STORAGE_SAFE_PATH_MAX) {
       DEBUG ((
         DEBUG_WARN,
-        "Kernel->Add[%u]->ExecutablePath (length %u) is too long (should not exceed %u)!\n",
+        "Kernel->Add[%u]->BundlePath + ExecutablePath (length %u) is too long (should not exceed %u)!\n",
```
mikebeaton, line 231: Also coming round to your way of thinking. How about printing `BundlePath + '\' + ExecutablePath` in the message (so adding `+ '\\' ` to the format string here, almost what you had before)?

Utilities/ocvalidate/ValidateKernel.c
corpnewt: Thank you for the clarification - I'll update the PR

PR #579, opened 2025-08-10
OcConsoleLib: Normalize console font path

al3xtjames:
OcLoadConsoleFont currently loads the specified console path from `Resources\Font\\`, which can cause vault failures if vault.plist doesn't include the duplicate path separator. Remove it to match other paths.


mikebeaton:
Thanks

PR #582, opened 2025-09-14
Patches: Allow DEBUG OpenShell to start with too many file systems under Mac EFI

mikebeaton:
Some EFI such as Mac return Invalid Parameter if a shell environment variable is too long (> 1,000 chars). Since all of root path, \efi\tools and \efi\boot are added for each file system found, the path variable exceeds this limit when there are greater than 25 file systems.

With this change DEBUG shell behaves like RELEASE shell: due to the implementation of SetEnv, when there is an error the variable gets deleted and not recreated, thus the final path variable only contains paths for file systems after the one which caused the error. While this is not an ideal fix for the underlying problem, it does simply recreate the longstanding behaviour of the RELEASE shell, and is preferable to the DEBUG shell being unusable on such systems.

Addresses https://github.com/acidanthera/bugtracker/issues/2005


mikebeaton:
I did a bit more research to confirm the underlying cause. If you attempt to set an NVRAM var with a size greater than 2006 bytes (go figure?) it is ignored and the previous value if any stays set. This applies even setting NVRAM vars from the macOS terminal with the `nvram` command.

I still think this is probably a good enough fix for a patch, as it makes the DEBUG shell work the same way the RELEASE shell always has on these systems.

A slightly better fix (possibly worthy of an audk commit) might be @joevt's second suggestion [here](https://github.com/acidanthera/bugtracker/issues/2005#issue-1225034313), of reverting to the previous value if setting the new value fails.

vit9696:
Let's do it for the time being, thank you!

PR #583, opened 2025-10-05
Local base tools

mikebeaton:
Fix two very minor nits in local BaseTools compilation:

 - Avoid `-s is obsolete` warning when compiling local BaseTools on macOS
 - Update build message to say 'additional BaseTools'
    - Avoids some potential confusion as this step is not compiling the well-known audk/edk-2 BaseTools, which are indeed compiled elsewhere in the build.

This is two separate commits. If okay, might be marginally nicer to commit as two commits rather than one merged.


mikebeaton:
Note sure if I need to worry about _not_ removing `-s` if building on older macOS? I don't think so, as I believe we only support building OC on reasonably recent toolchains (long after `-s` was deprecated)?

PR #599, opened 2026-04-16
fix: add Windows-friendly macrecovery invocation

aayushprsingh:
## Summary
- add a macrecovery.bat wrapper for Windows users
- document the Windows invocation in README.md
- clarify in 
ecovery_urls.txt that ./macrecovery.py is shell syntax and should be replaced on Windows

## Why
The release zip currently includes 
ecovery_urls.txt examples like ./macrecovery.py ..., which work in Unix-like shells but fail in Command Prompt with:

`	ext
'.' is not recognized as an internal or external command,
operable program or batch file.
`

This keeps the existing Unix workflow intact while giving Windows users a zero-guess wrapper that works from Command Prompt and PowerShell.

## Validation
- ran Utilities\\macrecovery\\macrecovery.bat -h on Windows successfully

### DIFF
--- a/Utilities/macrecovery/macrecovery.bat
+++ b/Utilities/macrecovery/macrecovery.bat
@@ -1,0 +1,20 @@
+@echo off
+setlocal
+
+set "SCRIPT_DIR=%~dp0"
+
+where py >nul 2>nul
+if %ERRORLEVEL% EQU 0 (
+  py -3 "%SCRIPT_DIR%macrecovery.py" %*
+  exit /b %ERRORLEVEL%
+)
+
+where python >nul 2>nul
+if %ERRORLEVEL% EQU 0 (
+  python "%SCRIPT_DIR%macrecovery.py" %*
+  exit /b %ERRORLEVEL%
+)
+
+echo Python launcher 'py' or 'python' was not found in PATH. 1>&2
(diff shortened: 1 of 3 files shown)


Utilities/macrecovery/macrecovery.bat
```diff
@@ -0,0 +1,20 @@
...
+
+set "SCRIPT_DIR=%~dp0"
+
+where py >nul 2>nul
+if %ERRORLEVEL% EQU 0 (
+  py -3 "%SCRIPT_DIR%macrecovery.py" %*
+  exit /b %ERRORLEVEL%
+)
+
+where python >nul 2>nul
```
vit9696, line 12: Could you please clarify whether on Windows you always have python and not python3 like on Linux?

mikebeaton: Some web searching - or rather use of similar but more recent evil developments... - seems to indicate that checking `py -3` followed by `python` is better than checking `python3` on Windows. Roughly speaking, the idea seems to be be that `py -3` is the closest Windows equivalent to `python3` (cf https://peps.python.org/pep-0397/#command-line-handling).

i.e. I believe the PR is right.

vit9696: Right. A bit weird, but we are not here to judge OS design approach.

PR #600, opened 2026-04-20
OcAppleKernelLib: System KC loading and cross-KC dependency resolution

MattJackson:
## What this PR does

Adds optional System Kernel Collection (System KC) loading to
`OcAppleKernelLib` and wires it through `OcMainLib`, so OpenCore can
resolve Boot-KC kext `OSBundleLibraries` dependencies against classes
that live in the System KC (for example `IOGraphicsFamily`,
`IOUSBFamily`, `AGPM`). When no System KC is staged on the OpenCore
ESP, behaviour is identical to today's injection flow.

## Why

Since macOS 11 the System KC hosts a large fraction of kext libraries
that used to be in the prelinkedkernel. OpenCore only injects into the
Boot KC, so any injected kext whose `OSBundleLibraries` names a
System-KC class fails at prelink with `library kext ... not found`.
Teaching OpenCore to optionally consult the System KC closes that gap
for any project whose kexts statically link against System-KC classes.

The feature is strictly opt-in: presence of
`EFI/OC/SystemKernelExtensions.kc` on the ESP activates it, and the
file is version-locked to the installed macOS build.

## Test plan

Reviewer can reproduce with the following minimal flow. Pointers
indicate what output to expect in the OpenCore DEBUG log
(`DisplayLevel=0x80000042`, i.e. `DEBUG_INFO | DEBUG_WARN`).

1. **Unit-level (no macOS required).**
   ```sh
   cd OpenCorePkg
   git am 0001-OcAppleKernelLib-Add-System-KC-context-loading.patch
   git am 0002-OcAppleKernelLib-Resolve-cross-KC-dependencies.patch
   git am 0003-OcAppleKernelLib-Translate-System-KC-symbols-at-link.patch
   git am 0004-OcMainLib-Stage-System-KC-for-prelinked-injection.patch
   git am 0005-Changelog-Note-System-KC-loading-support.patch

   make -C Utilities/TestProcessKernel
   Utilities/TestProcessKernel/ProcessKernel /path/to/prelinkedkernel
   ```
   Expect: exit 0, `out.bin` byte-identical to a run without the
   patches when no System KC is present. No new warnings.

2. **Integration-level (requires a macOS 11+ guest or host).**
   ```sh
   # On macOS host
   sudo cp /System/Library/KernelCollections/SystemKernelExtensions.kc \
     /Volumes/EFI/EFI/OC/

   # Build OpenCore with the patches and install OpenCore.efi to the ESP.
   # Boot with DEBUG_XCODE5 build + DisplayLevel=0x80000042.
   ```
   Expect these log lines (order matters, format verbatim):
   ```
   OC: System KC loaded from EFI partition (NNN bytes)
   OCAK: System KC ready (NNN bytes, NNN chained fixups)
   OCAK: System KC resolved com.apple.iokit.IOGraphicsFamily (VA 0x...)
   OCAK: System KC resolved com.apple.driver.AGPM (VA 0x...)
   ```
   The `OC: Prelinked status - Success` line that usually terminates
   OcKernelFileOpen's cache work must remain. macOS boots normally.

3. **Regression test (no System KC staged).**
   Remove `EFI/OC/SystemKernelExtensions.kc`, rebuild image or simply
   delete the file, reboot. Expect the DEBUG log to contain NONE of
   the `OC: System KC ...` or `OCAK: System KC ...` lines and
   macOS to still boot.

4. **Negative test (corrupt file).**
   Stage a file named `SystemKernelExtensions.kc` that is not a
   Mach-O (`echo not-a-macho > SystemKernelExtensions.kc`). Expect:
   ```
   OC: System KC loaded from EFI partition (N bytes)
   OCAK: System KC header is not a valid Mach-O
   OC: System KC parse failed - Invalid Parameter
   ```
   No panic, no crash; Boot KC injection continues normally.

## Review notes

A few specific spots that might benefit from review eyes:

- `PrelinkedContext.c` `InternalApplyKernelCollectionFixups ()`:
  only the x86_64 and arm64e kernel-cache pointer formats are
  handled. Formats we have not observed in shipping kernel
  collections are deliberately skipped with `continue` - please
  confirm the list is complete enough for your intended targets.
- `PrelinkedKext.c` `InternalFindSystemKCDependency ()`: the fallback
  at the end of `InternalCachedPrelinkedKext ()` still materialises a
  kernel-stub `PRELINKED_KEXT` when a bundle identifier is in
  neither collection. This preserves today's behaviour for non-KC
  dependencies (weak test symbols, etc.); open to changing that if
  you'd prefer the fallback path to disappear when
  `IsKernelCollection` is true.
- `Link.c`: the translation in `InternalSolveSymbolNonWeak ()` uses
  the same constant offsets as `KcFixupValue ()`. If a future
  macOS changes the shared-KC KASLR convention, both sites would need
  updating in sync; consider whether they should share a helper.
- `OpenCoreKernel.c` reads the System KC through `mOcStorage`, which
  anchors to the OpenCore ESP root. If a reviewer prefers a config
  knob (e.g. `Kernel/Scheme/SystemKernelCollectionPath`) I'm happy
  to add one; the current hard-coded filename felt like the minimal
  surface for a first cut.

## Log snippet

With the patches applied and
`EFI/OC/SystemKernelExtensions.kc` staged from macOS 15.7.5:

```
OC: Request is usr-apfs (0), 0(0) for custom Hook
OC: System KC loaded from EFI partition (365821088 bytes)
OCAK: System KC ready (365821088 bytes, 81247 chained fixups)
OCAK: System KC resolved com.apple.iokit.IOGraphicsFamily (VA 0xffffff8149b20000, 1590 fixups)
OCAK: System KC resolved com.apple.driver.AppleGraphicsDeviceControl (VA 0xffffff8149c80000, 56 fixups)
OC: Prelinked status - Success
```

## Checklist

- [x] Code compiles cleanly on `XCODE5 DEBUG` (Xcode 16.2, EDK2
      stable202502).
- [x] `Utilities/TestProcessKernel` runs unchanged in the no-System-KC
      case.
- [x] Changelog entry added.
- [x] Each commit is independently buildable for bisection.
- [ ] Reviewed arm64e pointer-format behaviour end-to-end on physical
      Apple-silicon hardware. (Pointer-format path compiles; no
      Apple-silicon hardware on hand for a real boot test.)

## Follow-up after #606

This PR contains its own inline chained-fixup walker code in
`InternalApplyFilesetKextFixups` (`PrelinkedKext.c`) and
`InternalKcTranslateSystemKCFixups` (`PrelinkedContext.c`). Both walkers
predate the public `KcWalkChainedFixupsInImage` API introduced by #603
and refined by #606 to be alignment-safe.

After #606 merges, I will follow up with a refactor PR that:

- Replaces both private walkers with calls to
  `KcWalkChainedFixupsInImage` and a per-call visitor callback.
- Adopts the same alignment-safe contract (`CONST UINT8 *` metadata,
  `UINT8 *FixupLoc` visitor parameter) so the same UB classes
  @vit9696 flagged on #603 (cast-and-deref of `MACH_DYLD_CHAINED_STARTS_IN_SEGMENT *`
  on a potentially-unaligned base, `UINT64 *` writes through unaligned
  fixup-slot pointers) are resolved here too.
- Eliminates ~40 lines of duplicate walker logic in OcAppleKernelLib.

Sequencing #606 first keeps that PR's review focused on the API +
walker alignment fix, and lets #600 land on its own merits without
blocking on the consumer-side refactor.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 OpenCore Changelog
 ==================
 #### v1.0.7
+- Added System KC loading and cross-KC dependency resolution in OcAppleKernelLib
 - Improved `XhciPortLimit` compatibility with macOS Tahoe, thx @laobamac
 - Updated builtin firmware versions for SMBIOS and the rest
 - Migrated to edk2-stable202511
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,7 @@ OpenCore Changelog
 - Added option to hide verbose output from any driver, thx @ilikesn0w
 
 #### v1.0.7
+- Added System KC loading and cross-KC dependency resolution in OcAppleKernelLib
 - Improved `XhciPortLimit` compatibility with macOS Tahoe, thx @laobamac
 - Updated builtin firmware versions for SMBIOS and the rest
 - Migrated to edk2-stable202511
--- a/Include/Acidanthera/Library/OcAppleKernelLib.h
+++ b/Include/Acidanthera/Library/OcAppleKernelLib.h
@@ -340,6 +340,30 @@ typedef struct {
   // Prelinked is 32-bit.
   //
   BOOLEAN                                Is32Bit;
+  //
+  // System KC buffer used to resolve cross-KC kext dependencies
+  // (e.g. IOGraphicsFamily). Optional; set via PrelinkedContextLoadSystemKC.
+  // Ownership is held by the context and released in PrelinkedContextFree.
+  //
+  UINT8                                  *SystemKC;
+  //
+  // System KC buffer size in bytes.
+  //
+  UINT32                                 SystemKCSize;
+  //
+  // Mach-O context covering the System KC outer header.
+  //
+  OC_MACHO_CONTEXT                       SystemKCMachContext;
+  //
+  // Mach-O context covering the System KC inner kernel when present,
+  // otherwise a copy of SystemKCMachContext. Used as the symtab source
+  // for fileset-entry kexts.
+  //
+  OC_MACHO_CONTEXT                       SystemKCInnerMachContext;
+  //
+  // TRUE once a System KC has been parsed and fixed up successfully.
+  //
+  BOOLEAN                                SystemKCValid;
 } PRELINKED_CONTEXT;
 
 //
@@ -820,6 +844,34 @@ EFI_STATUS
 PrelinkedDependencyInsert (
   IN OUT  PRELINKED_CONTEXT  *Context,
   IN      VOID               *Buffer
+  );
+
+/**
+  Associate a System Kernel Collection buffer with a prelinked context.
+
+  On macOS 11+ common kext libraries such as IOGraphicsFamily live in
+  the System KC rather than the Boot KC. When this function succeeds,
+  InternalCachedPrelinkedKext () can satisfy such dependencies by
+  walking the System KC's LC_FILESET_ENTRY commands.
+
+  LC_DYLD_CHAINED_FIXUPS on the outer collection are applied in place
+  so subsequent pointer reads observe resolved virtual addresses.
+
+  @param[in,out] Context       Initialized prelinked context.
+  @param[in]     SystemKC      System KC buffer. Ownership transfers to
+                               the context and is released by
+                               PrelinkedContextFree ().
+  @param[in]     SystemKCSize  System KC buffer size in bytes.
+
+  @retval EFI_SUCCESS            System KC associated successfully.
+  @retval EFI_INVALID_PARAMETER  Buffer is not a valid Mach-O kernel
+                                 collection.
+**/
+EFI_STATUS
+PrelinkedContextLoadSystemKC (
+  IN OUT  PRELINKED_CONTEXT  *Context,
+  IN      UINT8              *SystemKC,
+  IN      UINT32             SystemKCSize
   );
 
 /**
--- a/Library/OcAppleKernelLib/Link.c
+++ b/Library/OcAppleKernelLib/Link.c
@@ -386,7 +386,30 @@ InternalSolveSymbolNonWeak (
                     OcGetSymbolFirstLevel
                     );
   if (ResolveSymbol != NULL) {
-    InternalSolveSymbolValue (Context->Is32Bit, ResolveSymbol->Value, Symbol);
+    UINT64  ResolvedValue;
+
+    ResolvedValue = ResolveSymbol->Value;
+
+    //
+    // System KC kexts expose symbols with raw fileset virtual addresses
+    // (e.g. 0x149xxxxx, in the 1 MB .. 512 MB range). These must be
+    // translated into the kernel address space before the link phase
+    // emits RIP-relative relocations, otherwise the 32-bit displacements
+    // cannot reach them. The translation mirrors KcFixupValue (), which
+    // performs the same adjustment for vtable pointers; on macOS 11+
+    // all kernel collections share a single KASLR slide.
+    //
+    // Kernel-space values, zero, and weak test placeholders must not
+    // be translated.
+    //
+    if (  !Context->Is32Bit
+       && (ResolvedValue >= BASE_1MB)
+       && (ResolvedValue < BASE_1MB * 512ULL))
+    {
+      ResolvedValue = ResolvedValue + KERNEL_FIXUP_OFFSET + KERNEL_ADDRESS_BASE;
+    }
+
+    InternalSolveSymbolValue (Context->Is32Bit, ResolvedValue, Symbol);
   }
 
   return TRUE;
@@ -577,6 +600,12 @@ InternalCalculateDisplacementIntel64 (
   Difference   = ABS (Displacement);
 
   if (Difference >= X86_64_RIP_RELATIVE_LIMIT) {
+    DEBUG ((
+      DEBUG_INFO,
+      "OCAK: RIP-relative displacement overflow: target=0x%Lx diff=0x%Lx\n",
+      Target,
+      Difference
+      ));
     return FALSE;
   }
 
--- a/Library/OcAppleKernelLib/Link.c
+++ b/Library/OcAppleKernelLib/Link.c
@@ -386,7 +386,30 @@ InternalSolveSymbolNonWeak (
                     OcGetSymbolFirstLevel
                     );
   if (ResolveSymbol != NULL) {
-    InternalSolveSymbolValue (Context->Is32Bit, ResolveSymbol->Value, Symbol);
+    UINT64  ResolvedValue;
+
+    ResolvedValue = ResolveSymbol->Value;
+
+    //
+    // System KC kexts expose symbols with raw fileset virtual addresses
+    // (e.g. 0x149xxxxx, in the 1 MB .. 512 MB range). These must be
+    // translated into the kernel address space before the link phase
+    // emits RIP-relative relocations, otherwise the 32-bit displacements
+    // cannot reach them. The translation mirrors KcFixupValue (), which
+    // performs the same adjustment for vtable pointers; on macOS 11+
+    // all kernel collections share a single KASLR slide.
+    //
+    // Kernel-space values, zero, and weak test placeholders must not
+    // be translated.
+    //
+    if (  !Context->Is32Bit
+       && (ResolvedValue >= BASE_1MB)
+       && (ResolvedValue < BASE_1MB * 512ULL))
+    {
+      ResolvedValue = ResolvedValue + KERNEL_FIXUP_OFFSET + KERNEL_ADDRESS_BASE;
+    }
+
+    InternalSolveSymbolValue (Context->Is32Bit, ResolvedValue, Symbol);
   }
 
   return TRUE;
--- a/Library/OcAppleKernelLib/PrelinkedContext.c
+++ b/Library/OcAppleKernelLib/PrelinkedContext.c
@@ -520,9 +520,9 @@ PrelinkedContextFree (
 STATIC
 EFI_STATUS
 InternalApplyKernelCollectionFixups (
-  IN  UINT8              *Buffer,
-  IN  OC_MACHO_CONTEXT   *MachContext,
-  OUT UINT32             *FixupCountOut  OPTIONAL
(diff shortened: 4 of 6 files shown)


vit9696:
Thank you for your patch. It will take us some time to review, but could you first explain whether we absolutely need to copy the .kc file? The problem with this architecture is .kc desync, which happens:

* after macOS upgrade;
* when macOS loads recovery;
* during macOS upgrade/install;
* with multiple macOS installations.

I understand the problem with kext injection exists, but it feels like we ideally want a better approach than copying.

MattJackson:
Thanks for the early read.

The copy-to-ESP design is driven by a constraint rather than a preference: the sealed APFS System volume is not readable from EFI. Only Apple's `boot.efi` can follow the firmlink from the preboot volume to `/System/Library/KernelCollections/SystemKernelExtensions.kc`. From the OpenCore driver stack (APFS JumpStart + the Apple partition driver) we can reach the Preboot volume, but the sealed System snapshot that actually stores the `.kc` is not surfaced. I did not find a pre-`ExitBootServices` path to the live file, which is why `OcMainLib/OpenCoreKernel.c` pulls it from `EFI/OC/` via `mOcStorage`.

Your desync concerns are all real:

1. Post-upgrade — the staged `.kc` is version-locked to the build it was copied from; booting a newer System volume against an older staged KC will mis-resolve symbols. The current patch does not validate build UUID against the running kernel; a hash/UUID check in `PrelinkedContextLoadSystemKC()` before trusting the context would be straightforward to add.
2. Recovery — Recovery uses its own `.kc`, so the ESP copy is wrong by construction. The simplest containment is to skip `PrelinkedContextLoadSystemKC()` when the selected boot entry is Recovery; I can wire that through `OcKernelProcessPrelinked()` without touching the library core.
3. In-progress install/upgrade — same class as (1); a UUID mismatch check catches it.
4. Multi-install — one ESP, N System volumes, N `.kc` files. This is the hardest case; any per-install staging scheme leaks into the config schema.

If you would prefer a design that avoids ESP staging entirely, the two directions I can see are: (a) an APFS-snapshot-aware reader on top of the existing Apple partition driver that can open the sealed System volume read-only at EFI time; (b) deferring cross-KC resolution until after `boot.efi` has mapped the System KC, which changes the patch's shape considerably. Happy to go either way — which would you like me to pursue, or would a UUID-gated version of the current approach (with Recovery skipped) be acceptable as a first step?

vit9696:
Hmmm. Thank you for your analysis.

I think option (a) can be considered a fallback if we can implement it cleanly, but I have doubts we can. It requires verifying image seal, and I am afraid EFI driver is not ready for this.

I also think option (b) is not feasible as I believe EfiBoot does not load system kc at all. I may be wrong, but I have always thought it is part of XNU OSKext APIs, namely loadKCFileSet, which are not open-source. Are you sure EfiBoot does it?

I honestly wonder whether it makes sense to try option (c) revive https://github.com/acidanthera/Lilu/pull/86/changes and work in mixed model: OC prepares, Lilu injects later stage kexts.

MattJackson:
Thanks so much for digging into this — the analysis on all three
directions is really helpful and saved me from chasing paths that
wouldn't have worked out.

**(a)** — you're right that verifying the APFS image seal from an EFI
driver is a real concern, and I don't think there's a clean way to get
there without duplicating a non-trivial chunk of APFS's sealed-snapshot
logic into the Apple partition driver. The maintenance cost of tracking
APFS format changes over time would outweigh the benefit, so I'd agree
with your read that this isn't the direction to push.

**(b)** — you're correct, and I double-checked: `boot.efi` does not
load the System KC. System KC loading happens in XNU after VFS comes
up, via `OSKext::loadKCFileSet(path, kc_kind_t)` in
`libkern/c++/OSKext.cpp` (public header lives in the xnu source tree).
EfiBoot only maps the Boot KC. So option (b) was architecturally
misplaced on my part — cross-KC resolution is fundamentally an
XNU-time concern, not an EFI-time one. Appreciate you catching that.

**(c)** — thanks for pointing at Lilu/pull/86, that's a useful
reference. We're coming at this from a downstream context where a
Lilu-side solution wouldn't land on our own path, so we're probably
not the right folks to carry that revival ourselves. If someone in
the Lilu community ends up picking it up on its own merits, that'd
be great to see.

**On this PR — I don't think it has to be all-or-nothing.**

Going back through the series with the System-KC-loading parts
peeled off, there are a few pieces I think could potentially stand
on their own merits regardless of how (or whether) the broader
cross-KC story gets resolved:

- A small handful of pure observability / diagnostic improvements
  (a `DEBUG_INFO` log line on dependency-scan failures, a
  RIP-relative-displacement overflow log in
  `InternalCalculateDisplacementIntel64`, a missing include).
  Roughly 14 LOC of pure cleanup; happy to submit as a trivial
  separate PR if it'd be useful.
- A latent bug fix nested inside one of the System-KC commits:
  in `InternalCachedPrelinkedKext`, when a referenced kext isn't
  found, the current behaviour returns NULL immediately, which
  silently breaks symbol-only dependents (weak references, build-time
  links). The patch adds a kernel-stub `PRELINKED_KEXT` fallback for
  that case. ~10 LOC, independent of System KC; could extract as
  part of the same trivial PR.
- The chained-fixup pointer-format handlers (X86_64 + arm64e
  variants) inside `InternalApplyKernelCollectionFixups`. These are
  potentially reusable as generic helpers for any caller working with
  chained-fixup buffers, but you've been pretty consistent that
  `OcAppleKernelLib` doesn't take infrastructure with no in-tree user.
  I could pair the helpers with a `Utilities/TestProcessKernel`
  subcommand that exercises them, but at that point we're back to a
  meatier PR, ~150 LOC, and I'd want your read on whether that shape
  is worth the review cycles.

So three concrete paths, in increasing ambition:

- **B. Trivial cleanup PR.** ~25 LOC: the 3 diagnostic / include nits
  plus the symbol-only-dependent stub fallback. No System-KC story,
  no library-without-caller concerns. Withdraw the rest of this PR.
- **C. Cleanup + chained-fixup helpers.** ~150 LOC: B plus the
  pointer-format extraction with a `TestProcessKernel` driver.
  Withdraw the rest of this PR.
- **D. Pause this PR entirely.** Leave the System-KC work on our
  downstream fork as reference, and let whoever cracks the Lilu
  side (or some future XNU-time approach) come back to it on their
  own timeline. We don't take any more of your review cycles on
  this thread until/unless that happens.

My instinct is **B**, but I'd genuinely rather defer to your read on
which (if any) of these is worth the round-trip from your side.
Happy to rebase to whichever you pick, or to just close this with
**D** and move on.

Thanks again for the careful review.

vit9696:
Right. Lilu solution is rather heavy, and we are not yet in the state we can pursue it. On the other side given macOS 26 is the last one supporting Intel, I believe there is room for compromise.

I personally think we can definitely implement B + C. This will be helpful for other use cases of kernel injection. As for system KC loading — I am yet to decide whether we want it.

Could you please tell me more about your use case? I wonder whether System KC loading in your scenario is practically the only choice?

MattJackson:
Thanks again for taking the time on this — really glad to hear B + C sound useful on their own merits. Happy to spin those off as a standalone PR on whatever cadence works for you; no rush on our end.

On the use case for the System KC loading half: the work originated in a graphics context. We were injecting a couple of small kexts that hook into the IOKit graphics stack — concretely, their `OSBundleLibraries` references `com.apple.iokit.IOAcceleratorFamily2`, `com.apple.iokit.IOSurface`, and `com.apple.iokit.IOGraphicsFamily`, all of which live in `SystemKernelExtensions.kc` rather than the boot KC. With only the boot KC visible to OpenCore's prelinker, those dependencies fail to resolve at prelink time with the "library kext … not found" error, even though the same kext loads cleanly post-boot if the kernel itself can see the system KC. Teaching OpenCore to consult a staged System KC was the most direct way we found to make those graphics-stack-dependent kexts prelink without changing the guest's actual boot path or shipping a modified XNU.

It's not the only conceivable choice — a Lilu-side late-stage injection would reach the same end state, and you raised that alternative earlier in the thread re: Lilu PR #86. For our particular needs the OpenCore-prelink approach was the cleanest fit, since we're not the right folks to drive the Lilu side and we wanted the resulting boot to look as close to a stock OpenCore boot as possible.

That said: we're heads-down on a few other things right now and this isn't blocking us in any way. We mainly wanted to get the code into your tree as a contribution if it's useful to anyone else hitting the same prelink wall on graphics-stack kexts, and we're entirely happy to wait, iterate on whatever shape you'd prefer, or just close the System-KC half and go forward with B (or B + C) if you'd rather not carry the cross-KC machinery. Whichever way you'd like to take it is fine by us — please don't feel any pressure to merge on our account.

Thanks again for the review.

MattJackson:
Per your endorsement of B + C: opened standalone PRs as #602 (B path: ~25 LOC, prelink dependency-scan + RIP-relative-overflow `DEBUG_INFO` traces, plus the kernel-stub `PRELINKED_KEXT` fallback for symbol-only dependents in `InternalCachedPrelinkedKext`, no SystemKC machinery) and #603 (C path: ~150 LOC of code plus header/doc, depends on #602; reusable `KcWalkChainedFixupsInSegment` / `KcWalkChainedFixupsInImage` helpers with correct per-format stride handling, plus a `TestProcessKernel --test-fixup-walk` subcommand exercising both walks).

`TestKextInject` and `TestProcessKernel` both build clean with `WERROR=1 USE_SHARED_OBJS=1` on each branch.

The originally proposed missing-include item turned out to be a non-issue on master (`PrelinkedInternal.h` already pulls `<IndustryStandard/AppleMachoImage.h>` into `PrelinkedKext.c` transitively), so I dropped it rather than invent a fix.

The arm64e stride bug from the original PR #600 commit B is fixed in the C-PR helpers: `MACH_DYLD_CHAINED_PTR_64_KERNEL_CACHE` is stride 4 (`Fixup->Next * 4`), not 1 byte. The synthetic test in `--test-fixup-walk` exercises both stride layouts and asserts the walker produces identical visit counts.

This PR's SystemKC machinery is deferred per your "yet to decide" comment — happy to leave it open for now and close it once #602 and #603 land, or to close it sooner if you'd prefer the cleaner queue. Equally happy to rebase or squash either side at your discretion.

PR #601, opened 2026-04-26
OpenCore: Switch from UefiDriverEntryPoint.h to UefiApplicationEntryPoint.h

ilikesn0w:
Since OpenCore is no longer a driver as of 2021, it must use UefiApplicationEntryPoint.h


mikebeaton:
Looks correct.

mikebeaton (approved):

PR #603, opened 2026-04-26
OcAppleKernelLib: chained-fixup pointer-format helpers + TestProcessKernel driver

MattJackson:
**Depends on #602** (the "B" cleanup path). This is the "C" path from
the PR #600 discussion. Adds reusable static-lib helpers for walking
`LC_DYLD_CHAINED_FIXUPS` chains in a kernel collection, with correct
per-format stride handling, and exercises them from a new
`TestProcessKernel --test-fixup-walk` subcommand so the helpers land
with an in-tree user.

## Helpers

`KcWalkChainedFixupsInSegment` and `KcWalkChainedFixupsInImage` in
`Library/OcAppleKernelLib/KernelCollection.c` (declared in
`Include/Acidanthera/Library/OcAppleKernelLib.h`). Both take a
caller-provided `KC_CHAINED_FIXUP_VISIT` callback invoked per fixup
slot, so the same walker drives translation passes (rewrite the slot
in place) and validation passes (count / observe). Per
Apple's `mach-o/fixup-chains.h`, `Fixup->Next` is a stride count, not
a byte offset. The walker computes the correct stride per format:

| PointerFormat                       | Value | Stride |
|-------------------------------------|------:|-------:|
| `X86_64_KERNEL_CACHE`               |    11 |   1 B  |
| `64_KERNEL_CACHE`                   |     8 |   4 B  |
| `64_OFFSET`                         |     6 |   4 B  |
| `ARM64E_KERNEL`                     |     7 |   4 B  |

Other formats (ARM64E with auth, 32-bit variants, userland 64) and
`START_MULTI` pages return 0 / are skipped defensively.

## Test driver

`Utilities/TestProcessKernel/ProcessKernel.c` gains a
`--test-fixup-walk` subcommand that builds a synthetic single-page
chain and walks it as `X86_64_KERNEL_CACHE`, `64_KERNEL_CACHE`, and
`ARM64E` (unsupported), verifying expected slot counts. Both supported
strides encode the same byte distance between slots so the walker
must yield identical visit counts. Local run on x86_64+arm64e darwin:

```
[OK] X86_64_KERNEL_CACHE walk visited 4 fixups (stride 1)
[OK] 64_KERNEL_CACHE walk visited 4 fixups (stride 4)
[OK] unsupported-format guard returned 0
```

## Notes

`Library/OcAppleKernelLib/KernelCollection.c` does not currently
contain an `InternalApplyKernelCollectionFixups` to refactor — that
function only exists in PR #600's part A. So C ships the helpers
fresh rather than as an extraction; behaviour of every existing
in-tree caller is unchanged. When future work needs a fixup walker
(SystemKC loading, fileset entry rebasing, or any other use case
mentioned in #600), it can route through these helpers rather than
hand-rolling another walker.

`TestKextInject` and `TestProcessKernel` both build clean with
`WERROR=1 USE_SHARED_OBJS=1` against this tree.

### DIFF
--- a/Library/OcAppleKernelLib/KernelCollection.c
+++ b/Library/OcAppleKernelLib/KernelCollection.c
@@ -835,3 +835,123 @@ KcFixupValue (
 
   return Value;
 }
+
+UINTN
+KcWalkChainedFixupsInSegment (
+  IN     UINT8                                *Buffer,
+  IN     MACH_DYLD_CHAINED_STARTS_IN_SEGMENT  *StartsSeg,
+  IN     KC_CHAINED_FIXUP_VISIT               Visitor OPTIONAL,
+  IN OUT VOID                                 *VisitorContext OPTIONAL
+  )
+{
+  MACH_DYLD_CHAINED_PTR_64_KERNEL_CACHE_REBASE  *Fixup;
+  UINT64                                        *FixupLoc;
+  UINT32                                        PageIdx;
+  UINT32                                        Stride;
+  UINTN                                         Count;
+
+  ASSERT (Buffer != NULL);
+  ASSERT (StartsSeg != NULL);
+
+  //
+  // Per Apple's mach-o/fixup-chains.h, the Fixup->Next field is a stride
+  // count, not a byte offset. The stride depends on the pointer format:
+  //
+  //   X86_64_KERNEL_CACHE (11): stride 1 byte  (Fixup->Next * 1)
+  //   64_KERNEL_CACHE      (8): stride 4 bytes (Fixup->Next * 4)
+  //
+  // The 64_OFFSET (6) and ARM64E_KERNEL (7) variants also use stride 4
+  // and share the same per-pointer rebase layout for the unauth case.
+  // Other formats (ARM64E with auth, 32-bit variants, userland 64) are
+  // not produced by current macOS kernel caches and are skipped here.
+  //
+  switch (StartsSeg->PointerFormat) {
+    case MACH_DYLD_CHAINED_PTR_X86_64_KERNEL_CACHE:
+      Stride = 1;
+      break;
+    case MACH_DYLD_CHAINED_PTR_64_KERNEL_CACHE:
+    case MACH_DYLD_CHAINED_PTR_64_OFFSET:
+    case MACH_DYLD_CHAINED_PTR_ARM64E_KERNEL:
+      Stride = 4;
+      break;
+    default:
+      return 0;
+  }
+
+  Count = 0;
+
+  for (PageIdx = 0; PageIdx < StartsSeg->PageCount; ++PageIdx) {
+    if (StartsSeg->PageStart[PageIdx] == MACH_DYLD_CHAINED_PTR_START_NONE) {
+      continue;
+    }
+
+    //
+    // STARTS_IN_SEGMENT entries with the START_MULTI bit set carry their
+    // chain heads via a separate ChainStarts table appended after
+    // PageStart[PageCount]. macOS kernel caches do not emit this layout
+    // for the formats handled above; skip such pages defensively rather
+    // than mis-walking foreign data.
+    //
+    if ((StartsSeg->PageStart[PageIdx] & MACH_DYLD_CHAINED_PTR_START_MULTI) != 0) {
+      continue;
+    }
+
+    FixupLoc = (UINT64 *)(
+                          Buffer + StartsSeg->SegmentOffset
+                          + (UINT64)PageIdx * StartsSeg->PageSize
+                          + StartsSeg->PageStart[PageIdx]
+                          );
+
+    while (TRUE) {
+      Fixup = (MACH_DYLD_CHAINED_PTR_64_KERNEL_CACHE_REBASE *)FixupLoc;
+
+      if (Visitor != NULL) {
+        Visitor (FixupLoc, VisitorContext);
+      }
+
+      ++Count;
+
+      if (Fixup->Next == 0) {
+        break;
+      }
+
+      FixupLoc = (UINT64 *)(
+                            (UINT8 *)FixupLoc + (UINTN)Fixup->Next * Stride
+                            );
+    }
+  }
+
+  return Count;
+}
+
+UINTN
+KcWalkChainedFixupsInImage (
+  IN     UINT8                              *Buffer,
+  IN     MACH_DYLD_CHAINED_STARTS_IN_IMAGE  *Starts,
+  IN     KC_CHAINED_FIXUP_VISIT             Visitor OPTIONAL,
+  IN OUT VOID                               *VisitorContext OPTIONAL
+  )
+{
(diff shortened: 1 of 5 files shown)


Library/OcAppleKernelLib/KernelCollection.c
```diff
@@ -835,3 +835,123 @@ KcFixupValue (
...
+    case MACH_DYLD_CHAINED_PTR_ARM64E_KERNEL:
+      Stride = 4;
+      break;
+    default:
+      return 0;
+  }
+
+  Count = 0;
+
+  for (PageIdx = 0; PageIdx < StartsSeg->PageCount; ++PageIdx) {
```
vit9696: Here you seem to trust input, which I am not positive why. To me it seems buffer overflow may happen for faulty `PageCount`.

Library/OcAppleKernelLib/KernelCollection.c
```diff
@@ -835,3 +835,123 @@ KcFixupValue (
...
+    // chain heads via a separate ChainStarts table appended after
+    // PageStart[PageCount]. macOS kernel caches do not emit this layout
+    // for the formats handled above; skip such pages defensively rather
+    // than mis-walking foreign data.
+    //
+    if ((StartsSeg->PageStart[PageIdx] & MACH_DYLD_CHAINED_PTR_START_MULTI) != 0) {
+      continue;
+    }
+
+    FixupLoc = (UINT64 *)(
```
vit9696: Here you also seem to trust input, e.g. I believe`StartsSeg->SegmentOffset` may be wrong or calculation may overflow.

Library/OcAppleKernelLib/KernelCollection.c
```diff
@@ -835,3 +835,123 @@ KcFixupValue (
...
+        Visitor (FixupLoc, VisitorContext);
+      }
+
+      ++Count;
+
+      if (Fixup->Next == 0) {
+        break;
+      }
+
+      FixupLoc = (UINT64 *)(
```
vit9696: Here you trust input too.

Library/OcAppleKernelLib/KernelCollection.c
```diff
@@ -835,3 +835,123 @@ KcFixupValue (
...
+  MACH_DYLD_CHAINED_STARTS_IN_SEGMENT  *StartsSeg;
+  UINT32                               SegIdx;
+  UINTN                                Count;
+
+  ASSERT (Buffer != NULL);
+  ASSERT (Starts != NULL);
+
+  Count = 0;
+
+  for (SegIdx = 0; SegIdx < Starts->NumSegments; ++SegIdx) {
```
vit9696: Checks are needed here too. In fact both functions should also get size limit passed to work correctly.

vit9696 (commented):
Thank you for a PR. Sorry for being rather slow at review. Commented on a few issues.

MattJackson:
Thanks for the review — addressed in dbf2ba22.

Both helpers now take explicit size limits and treat every field read from `MACH_DYLD_CHAINED_STARTS_IN_{SEGMENT,IMAGE}` as untrusted input.

**Updated signatures:**

```c
UINTN
KcWalkChainedFixupsInSegment (
  IN     UINT8                                *Buffer,
  IN     UINTN                                BufferSize,
  IN     MACH_DYLD_CHAINED_STARTS_IN_SEGMENT  *StartsSeg,
  IN     UINTN                                StartsSegSize,
  IN     KC_CHAINED_FIXUP_VISIT               Visitor OPTIONAL,
  IN OUT VOID                                 *VisitorContext OPTIONAL
  );

UINTN
KcWalkChainedFixupsInImage (
  IN     UINT8                              *Buffer,
  IN     UINTN                              BufferSize,
  IN     MACH_DYLD_CHAINED_STARTS_IN_IMAGE  *Starts,
  IN     UINTN                              StartsSize,
  IN     KC_CHAINED_FIXUP_VISIT             Visitor OPTIONAL,
  IN OUT VOID                               *VisitorContext OPTIONAL
  );
```

**Per your four comments:**

1. **`PageCount` (line 883)** — Now bounded against `StartsSeg->Size`, which is itself bounded against `StartsSegSize`. The check `PageCount > (Size - StructHeaderSize) / sizeof (UINT16)` rejects any `PageCount` that would walk the `PageStart[]` array past the metadata region.

2. **`SegmentOffset` arithmetic (line 899)** — All page-base math goes through `BaseOverflowMulAddU64` (`SegmentOffset + PageIdx * PageSize`) and `BaseOverflowAddU64` (`+ PageSize`), with a final `PageEnd <= BufferSize` check. `SegmentOffset >= BufferSize` is rejected up front. `PageStart` is also constrained to be `< PageSize` so the chain head can't start outside its own page.

3. **Chain step `Fixup->Next * Stride` (line 918)** — Each step computes `NextStep = Fixup->Next * Stride` via `BaseOverflowMulU64`, adds it via `BaseOverflowAddU64`, and rejects any next-slot offset that would extend past the current page. A per-page iteration cap of `PageSize / sizeof (UINT64)` (= 512 for a 4 KB page) defeats self-referencing chains where overlapping byte-stepping could otherwise visit more slots than the page-end check alone would catch.

4. **`KcWalkChainedFixupsInImage` (line 944)** — Now takes `StartsSize`. Validates the fixed `NumSegments` header fits, `SegInfoOffset[NumSegments]` array fits, each `SegInfoOffset < StartsSize`, and forwards `StartsSize - SegOffset` as the per-segment `StartsSegSize` so `InSegment` validates the rest.

**Test coverage** — `TestProcessKernel --test-fixup-walk` extended with seven negative tests proving each defense fires:

```
[OK] X86_64_KERNEL_CACHE walk visited 4 fixups (stride 1)
[OK] 64_KERNEL_CACHE walk visited 4 fixups (stride 4)
[OK] unsupported-format guard returned 0
[OK] PageCount oversize rejected
[OK] Size > StartsSegSize rejected
[OK] SegmentOffset out-of-buffer rejected
[OK] PageStart >= PageSize rejected
[OK] long-chain bounded at iteration cap (512)
[OK] chain step past page-end rejected after 1 visit
[OK] InImage oversize NumSegments rejected
```

All 10 tests pass with `WERROR=1 USE_SHARED_OBJS=1` against this tree.

vit9696:
Thank you for the fix. Sorry for the delay, wanted to spend some time reading this. In the end I believe maybe we could have simplified some checks, but that would likely have made things a bit less readable, so it is good.

There are several issues with this change in regard to alignment checks, e.g.

1. `Starts + SegOffset` is not guaranteed to be aligned for `MACH_DYLD_CHAINED_STARTS_IN_SEGMENT`.
2. `Buffer + SlotOffset` is not guaranteed to be aligned for `UINT64` (in two places).

While Intel supports unaligned reads and writes, C standard considers this undefined behaviour, and some compiler optimisations may sometimes introduce bugs when parsing malformed binaries. This may also cause crashes on some implementations of ARM architecture during userspace testing/fuzz-testing as well as UBSan warnings.

I am still merging this, but a follow-up PR would be nice if you still have enough resources.

PR #604, opened 2026-04-27
OcMainLib: Add HideVerbose option

ilikesn0w:
Starting with macOS Tahoe, users have begun using apfs_aligned.efi due to issues with FV2. However, this driver produces unnecessary console output. I propose adding a HideVerbose option to suppress this output, similar to how IgnoreVerbose works in OcApfsLib.

### DIFF
--- a/Library/OcMainLib/OpenCoreUefi.c
+++ b/Library/OcMainLib/OpenCoreUefi.c
@@ -67,6 +67,7 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 STATIC EFI_EVENT_NOTIFY  mOcExitBootServicesHandlers[OC_EXIT_BOOT_SERVICES_HANDLER_MAX+1];
 STATIC VOID              *mOcExitBootServicesContexts[OC_EXIT_BOOT_SERVICES_HANDLER_MAX];
 STATIC UINTN             mOcExitBootServicesIndex;
+STATIC EFI_SYSTEM_TABLE  *mNullSystemTable;
 
 VOID
 OcScheduleExitBootServices (
@@ -206,7 +207,7 @@ OcLoadDrivers (
       continue;
     }
 
-    if ((DriverArguments != NULL) && (DriverArguments[0] != '\0')) {
+    if (((DriverArguments != NULL) && (DriverArguments[0] != '\0')) || DriverEntry->HideVerbose) {
       Status = gBS->HandleProtocol (
                       ImageHandle,
                       &gEfiLoadedImageProtocolGuid,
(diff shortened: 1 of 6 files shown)


mikebeaton:
Can you check whether this works: https://github.com/acidanthera/bugtracker/issues/2499#issuecomment-4328921493 ?

ilikesn0w:
> Can you check whether this works: [acidanthera/bugtracker#2499 (comment)](https://github.com/acidanthera/bugtracker/issues/2499#issuecomment-4328921493) ?

It didn't change anything.

mikebeaton:
> > Can you check whether this works: [acidanthera/bugtracker#2499 (comment)](https://github.com/acidanthera/bugtracker/issues/2499#issuecomment-4328921493) ?
> 
> It didn't change anything.

Ah right, apologies. I added that option in EnableGop to suppress exactly these messages, but the loading order is wrong in normal OpenCore - the console control protocol doesn't get installed until after drivers are loaded, so unfortunately the setting won't change anything here.

mikebeaton:
Right, I had a closer look and I see what you're doing - basically allowing the existing `HideVerbose` logic from APFS jumpstart to be applied to any driver, on a per-driver basis.

I'd suggest some cleanups, if we want it:
 - I'd suggest changing `AllocateNullTextOutSystemTable` to keep track of whether the table has already been allocated, internally to itself; then the same call can be used in multiple places without repeating the allocation logic
   - Maybe rename from `Allocate...` to `Get...` to reflect the changed semantics
 - I would only fetch LoadedImageProtocol in your new code if we need it:
   - `((DriverArguments != NULL) && (DriverArguments[0] != '\0')) || DriverEntry->HideTextOut)`
 - I would probably suggest using the name `HideVerbose` instead of `HideTextOut` - as it's already been chosen for the same functionality

But I am not certain we do need it, with only this one use case, so far, which already has a [workable fix](https://github.com/acidanthera/bugtracker/issues/2499#issuecomment-3781101485) (from your good self).

Many thanks for providing the PR.

cc @vit9696

ilikesn0w:
Okay, just in case your team accepts this PR, I’ve made changes 2 and 3, as you requested. I think it would be better to make the changes to the AllocateNullTextOutSystemTable in a separate PR.

Library/OcMainLib/OpenCoreUefi.c
```diff
@@ -224,7 +225,19 @@ OcLoadDrivers (
         FreePool (Driver);
         continue;
       }
```
mikebeaton, line 227: Just noting, for other reviewers, that this is stronger than the APFS jumpstart HideVerbose code, which carries on if it can't get the loaded image protocol, but I don't think it matters.

mikebeaton:
> Okay, just in case your team accepts this PR, I’ve made changes 2 and 3, as you requested. I think it would be better to make the changes to the AllocateNullTextOutSystemTable in a separate PR.

I don't agree - it's fine, and probably better, in this PR, if it's going to happen. Maybe indeed as a separate commit. But I do agree there's no point making the change unless @vit9696 also thinks it's a good change.

(Btw for your commits so far, original plus revisions, you can squash them into one commit and force push to your branch. GitHub will track it and update the PR.)

PR #607, opened 2026-06-28
CI: Adapt for new GitHub and Homebrew 6 requirements

ilikesn0w:
Since GitHub has [announced ](https://github.blog/changelog/2026-05-14-github-actions-upcoming-image-migrations/)new changes to CI and Homebrew 6 was released, I suggest making the following changes:

- Change macOS runner to macos-15 since Tahoe causes compilation problems
- Change Windows runner to windows-2022 since AUDK currently doesn't support VS2026
- Add brew trust for 3rd party taps

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -153,7 +153,7 @@ jobs:
 
   build-windows:
     name: Windows VS2022
-    runs-on: windows-latest
+    runs-on: windows-2022
     env:
       HAS_OPENSSL_BUILD: 0
       CC: gcc
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Install Linux Toolchain
         run: |
           brew tap FiloSottile/homebrew-musl-cross
-          brew install musl-cross
+          brew install FiloSottile/musl-cross/musl-cross
 
       - name: Install Dependencies
         run: brew install mingw-w64


.github/workflows/build.yml
```diff
@@ -27,6 +27,8 @@ jobs:
 
       - name: Install Linux Toolchain
         run: |
+          brew trust aws/tap
+          brew trust FiloSottile/musl-cross
           brew tap FiloSottile/homebrew-musl-cross
           brew install musl-cross
```
mikebeaton: I think you should be able to switch to:
```
brew tap FiloSottile/homebrew-musl-cross
brew install FiloSottile/musl-cross/musl-cross
```
install by fully qualified name should trust automatically (https://docs.brew.sh/Tap-Trust).

.github/workflows/analyze.yml
```diff
@@ -23,6 +23,7 @@ jobs:
       # Due to brew adopting PEP 668 https://github.com/orgs/Homebrew/discussions/3404 pip installs should be in a Python venv
       - name: Install Dependencies
         run: |
+          brew trust aws/tap
```
mikebeaton: It might be giving a warning, but I believe we are not using anything from it - so we don't want to trust it.

.github/workflows/build.yml
```diff
@@ -27,8 +27,9 @@ jobs:
 
       - name: Install Linux Toolchain
         run: |
+          brew trust aws/tap
```
mikebeaton: I don't believe we need this line - I think the warning is coming from a tap included in the runner image - it would be nice to have no warnings, but not by trusting something we don't want to use.

mikebeaton:
Very slightly modified and split commits. Merged under your name. Thanks.

PR #611, opened 2026-08-13
ci: grant contents:write to build-macos job for release uploads

albert-mueller:
Fixes "Resource not accessible by integration" from svenstaro/upload-release-action, caused by the default read-only GITHUB_TOKEN.


vit9696:
Thank you!

PR #614, opened 2026-08-15
OcAppleKernelLib: fix unchecked overflow and align-before-check in PrelinkedContext.c

albert-mueller:
- PrelinkedContextInit: guard the KC __LINKEDIT SegmentEndOffset and PrelinkedLastLoadAddress additions with BaseOverflowAddU64, matching the overflow-checked style already used elsewhere in this file (e.g. PrelinkedFindLastLoadAddress).
- PrelinkedInjectPrepare: compute AlignedExpansion before validating it against PrelinkedAllocSize, instead of validating the unaligned LinkedExpansion and then zeroing/committing the larger aligned value.

### DIFF
--- a/Library/OcAppleKernelLib/PrelinkedContext.c
+++ b/Library/OcAppleKernelLib/PrelinkedContext.c
@@ -384,14 +384,30 @@ PrelinkedContextInit (
     //
     // In KC mode last load address is the __LINKEDIT address.
     //
-    SegmentEndOffset = Context->LinkEditSegment->Segment64.FileOffset + Context->LinkEditSegment->Segment64.FileSize;
+    if (BaseOverflowAddU64 (
+          Context->LinkEditSegment->Segment64.FileOffset,
+          Context->LinkEditSegment->Segment64.FileSize,
+          &SegmentEndOffset
+          ))
+    {
+      PrelinkedContextFree (Context);
+      return EFI_INVALID_PARAMETER;
+    }
 
     if (MACHO_ALIGN (SegmentEndOffset) != Context->PrelinkedSize) {
       PrelinkedContextFree (Context);
       return EFI_INVALID_PARAMETER;
     }
 
-    Context->PrelinkedLastLoadAddress = Context->LinkEditSegment->Segment64.VirtualAddress + Context->LinkEditSegment->Segment64.Size;
+    if (BaseOverflowAddU64 (
+          Context->LinkEditSegment->Segment64.VirtualAddress,
+          Context->LinkEditSegment->Segment64.Size,
+          &Context->PrelinkedLastLoadAddress
+          ))
+    {
+      PrelinkedContextFree (Context);
+      return EFI_INVALID_PARAMETER;
+    }
   }
 
   //
@@ -538,8 +554,16 @@ PrelinkedInjectPrepare (
     // For newer variant (KC mode) __LINKEDIT is last, and we need to expand it to enable
     // dyld fixup generation.
     //
-    if (  (Context->PrelinkedAllocSize < LinkedExpansion)
-       || (Context->PrelinkedAllocSize - LinkedExpansion < Context->PrelinkedSize))
+    //
+    // Align first and validate the aligned size, as that is what actually gets
+    // zeroed and committed below -- checking the unaligned LinkedExpansion here
+    // would let AlignedExpansion (rounded up) write past the validated bound.
+    //
+    AlignedExpansion = MACHO_ALIGN (LinkedExpansion);
+
+    if (  (AlignedExpansion < LinkedExpansion)
+       || (Context->PrelinkedAllocSize < AlignedExpansion)
+       || (Context->PrelinkedAllocSize - AlignedExpansion < Context->PrelinkedSize))
     {
       return EFI_OUT_OF_RESOURCES;
     }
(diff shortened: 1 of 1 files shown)


Library/OcAppleKernelLib/PrelinkedContext.c
```diff
@@ -384,14 +384,30 @@ PrelinkedContextInit (
     //
     // In KC mode last load address is the __LINKEDIT address.
     //
-    SegmentEndOffset = Context->LinkEditSegment->Segment64.FileOffset + Context->LinkEditSegment->Segment64.FileSize;
+    if (BaseOverflowAddU64 (
```
vit9696, line 387: This looks checked to me: https://github.com/acidanthera/OpenCorePkg/blob/164696d/Library/OcMachoLib/HeaderX.h#L1239-L1246

Library/OcAppleKernelLib/PrelinkedContext.c
```diff
@@ -384,14 +384,30 @@ PrelinkedContextInit (
...
+      return EFI_INVALID_PARAMETER;
+    }
 
     if (MACHO_ALIGN (SegmentEndOffset) != Context->PrelinkedSize) {
       PrelinkedContextFree (Context);
       return EFI_INVALID_PARAMETER;
     }
 
-    Context->PrelinkedLastLoadAddress = Context->LinkEditSegment->Segment64.VirtualAddress + Context->LinkEditSegment->Segment64.Size;
+    if (BaseOverflowAddU64 (
```
vit9696, line 402: This also looks checked to me: https://github.com/acidanthera/OpenCorePkg/blob/164696d/Library/OcMachoLib/HeaderX.h#L1248-L1250

Library/OcAppleKernelLib/PrelinkedContext.c
```diff
@@ -538,8 +554,16 @@ PrelinkedInjectPrepare (
...
     // dyld fixup generation.
     //
-    if (  (Context->PrelinkedAllocSize < LinkedExpansion)
-       || (Context->PrelinkedAllocSize - LinkedExpansion < Context->PrelinkedSize))
+    //
+    // Align first and validate the aligned size, as that is what actually gets
+    // zeroed and committed below -- checking the unaligned LinkedExpansion here
+    // would let AlignedExpansion (rounded up) write past the validated bound.
+    //
+    AlignedExpansion = MACHO_ALIGN (LinkedExpansion);
```
vit9696, line 562: Did you find an actual path to violate this? I am not yet sure this change is needed, but it will help if you confirm it.

vit9696 (commented):
Thank you for the patch. I am not fully positive all the changes are needed. Replied inline.

Library/OcAppleKernelLib/PrelinkedContext.c
albert-mueller: Yes, it is needed, just tested on my T2 MacBook Pro and for the first time on a T2 Mac it boots Tahoe via OpenCore Legacy Patcher T2 and now it is able to boot Tahoe without triggering Activation Lock or causing Needs authenticator panics. Just uploading a video showing my MacBook Pro 2020 booting OpenCore with a spoofed SMBIOS, with adding the latest changes I did in my fork on top:
https://youtu.be/_ACpaSJEt2Q?is=rXZPrSjH3SAsbL3_  - this video demonstrates my MacBook Pro 2020 4 thunderbolt 3 ports being able to boot via OpenCorePkg spoofed as MacBook Pro 2019 16 inch without much patches except the most obvious kexts and does get to the desktop.
On a Hackintosh system or a non-T2 Mac, you don’t have a T2 chip and as such this change only improves performance. But on T2 Macs trying to boot Tahoe via OpenCorePkg supported or not, this change is must-have to even boot.

vit9696: Well, hmm.

1. The first two changes do nothing, they are definitely checked prior to the code you changed.
2. The third change will abort booting, if the check you modified triggers. In fact, if the check triggers, you have corrupted kernel collection supplied by Apple, which has never been the case so far.

The fact your T2 system boots is dependent on something else, and these changes cannot affect it directly, I am afraid.

What is the origin of the changes?

albert-mueller: Fair, and thanks for checking — you're right on all three points.

Origin: these came out of an AI-assisted read-through of this file. It picked up on the overflow-checked style used elsewhere in PrelinkedContext.c and pattern-matched the same treatment onto these two additions, without tracing far enough to see they're already covered.

On 1 & 2: confirmed against MachoGetSegmentByName() → MachoGetNextSegment() directly. Context->LinkEditSegment can only ever come from there, and that path already checks FileOffset + FileSize (overflow, bounded against Context->FileSize) and VirtualAddress + Size (overflow) before a segment is ever returned. So both BaseOverflowAddU64 additions are unreachable, as you said.

On 3: I don't have a concrete path that violates the existing LinkedExpansion check, and since PrelinkedAllocSize is reserved with its own margin by the caller, I agree this is theoretical at best.

And I should retract the T2 boot claim — that's not explained by any of this. Whatever actually fixed that boot is unrelated to this file. Sorry for the noise, closing this.

vit9696: Right, thank you. Since 3 might happen, though very very unlikely, I made the change in master to make things cleaner anyway.

albert-mueller:
I’ve found that this error is still occurring mostly on unsupported T2 Macs: https://github.com/albert-mueller/OpenCore-Legacy-Patcher-T2/issues/130

PR #615, opened 2026-08-15
OcAppleKernelLib: fix Link.c symbol-by-value lookup silently skipping trailing symbols

albert-mueller:
InternalOcGetSymbolWorkerValue's unrolled fast loop computed SymbolsEnd from the level-specific NumSymbols (Kext->NumberOfSymbols, or NumberOfCxxSymbols for the C++-only search), rounded down to a multiple of 16. That correctly matches the comment's intent of rounding the *start* down for the C++-only case (a few harmless extra C symbols get scanned), but it also rounded the *end* down, so up to 15 real trailing symbols right before the true end of the table were never compared for a value match.

Fix: compute the fast-loop bound from Kext->NumberOfSymbols (the table's actual end) instead of the level-specific count, and add a scalar remainder loop that picks up the leftover (<16) symbols up to the table's real end.

### DIFF
--- a/Library/OcAppleKernelLib/Link.c
+++ b/Library/OcAppleKernelLib/Link.c
@@ -158,13 +158,30 @@ InternalOcGetSymbolWorkerValue (
     // Increasing the iteration block to more than 16 no longer pays off.
     // Note, lower loop is not on hot path.
     //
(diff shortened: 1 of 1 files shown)


albert-mueller:
I closed the PR due to bugs.

PR #618, opened 2026-08-23
Dockerfiles/oc-dev: fix flaky LLVM install (IPv4 + reset trust anchor…

albert-mueller:
Dockerfiles/oc-dev: fix flaky LLVM install (IPv4 + reset trust anchor on retry)
The llvm.sh retry loop could fail permanently instead of recovering:

1. apt.llvm.org resolves to both an A and an AAAA record. On runners/
   BuildKit sandboxes without working IPv6 egress, wget can pick the
   AAAA record and fail with "Network is unreachable" instead of
   falling back to IPv4.
2. If a wget call inside llvm.sh is interrupted by that failure while
   writing /etc/apt/trusted.gpg.d/apt.llvm.org.asc, it leaves behind an
   empty/corrupt key file. llvm.sh only (re)downloads the key when that
   file is absent, so every subsequent retry in the loop reuses the
   broken key and apt-get update fails with NO_PUBKEY, even once the
   network recovers.

Fix: prefer IPv4 for wget via /etc/wgetrc, and remove the trust anchor
and any partially-added apt.llvm.org sources list entry at the start of
every retry iteration, so a bad attempt cannot poison the ones after it.

CI log of the original failure: image 'Linux CLANGDWARF' job, docker
compose build-duet, all 5 llvm.sh attempts failing (2x unreachable
IPv6, 3x NO_PUBKEY from a stale key file left by a previous attempt).

This should fix this error: 
[Annotations.txt](https://github.com/user-attachments/files/31353407/Annotations.txt)


vit9696:
This is garbage, because e.g. /etc/wgetrc may not exist before your command.

Please stop pushing weird stuff here, this is not wanted.

PR #620, opened 2026-08-26
OcMainLib: treat empty Generic->SystemUUID like "OEM"

albert-mueller:
PlatformInfo->Generic->SystemUUID being empty is documented (Configuration.pdf, ValidatePlatformInfo.c) as legal and equivalent to "OEM", but OcLoadPlatformSupport()/OcGetLegacySecureBootECID() only recognised the literal "OEM" string. An empty value instead fell through to OcAsciiStrToRawGuid() on an empty string, whose failure status was not checked, leaving MacInfo->Oem.SystemUuid zeroed.

A zero UUID is skipped when writing the Data Hub entry and the system-id NVRAM variable, but SMBIOS_OVERRIDE_V falls back to the real hardware UUID from the original SMBIOS table instead of leaving it zero. So with an empty SystemUUID, PlatformInfo->SMBIOS ended up with the real hardware UUID while PlatformInfo->Generic (Data Hub/NVRAM/IOPlatformUUID) got nothing set, letting the OS mint its own value independently - exposing two different, sometimes per-boot-varying UUIDs instead of one consistent one.

Add PlatformInfoUuidIsOem() and use it at both call sites so empty and "OEM" are handled identically, matching the documented failsafe.

Again commited using Claude.
This is designed to attempt to fix this on unsupported T2 Macs:
This picture is from my 2018 Mac mini booted inside macOS 26's installer with the buggy OC version before this fix:
IMG_0857

Here's the EFI with the buggy version before this fix:
[Archiv.zip](https://github.com/user-attachments/files/31444888/Archiv.zip)

### DIFF
--- a/Library/OcMainLib/OpenCorePlatform.c
+++ b/Library/OcMainLib/OpenCorePlatform.c
@@ -35,6 +35,37 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 
 STATIC CHAR8  mCurrentSmbiosProductName[OC_OEM_NAME_MAX];
 
+/**
+  Determine whether an ASCII PlatformInfo->Generic->SystemUUID value should
+  be treated as OEM, i.e. extracted from the real, unspoofed SMBIOS table
+  instead of parsed as a user-specified UUID.
+
+  Per Configuration.pdf and OcValidateLib (see ValidatePlatformInfo.c), an
+  empty PlatformInfo->Generic->SystemUUID is documented as a legal value
+  equivalent to explicitly specifying "OEM". Previously only the literal
+  "OEM" string was recognised at the two call sites below, so leaving
+  SystemUUID empty silently fell through to OcAsciiStrToRawGuid() on an
+  empty string. That call fails and its status is not checked, so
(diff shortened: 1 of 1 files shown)


albert-mueller:
I recommend to review the changes carefully as these are written using Claude.

albert-mueller (commented):
.

albert-mueller:
I just tested it on my 2018 Mac mini and I found that with this latest commit, now at least it tries to mount the volume (it fails). To fix the mounting issue on T2 Macs it requires a bit more research.
But still T2 Macs benefit from this commit: now my Mac mini 2018 is running a little bit more stable inside the installer, for example.
From this commit mostly benefit only T1 Macs, non-T series Macs and Hackintoshes.

vit9696:
> PlatformInfo->Generic->SystemUUID being empty is documented (Configuration.pdf, ValidatePlatformInfo.c) as legal and equivalent to "OEM"

This is lies. It is documented as:

> Empty (OEM specified or not installed)

This reads if it was set to OEM value, then it remains set. I.e. it is left as is by default, not forcibly set to OEM in all variables like you might think of.

Please stop posting junk here. It takes time to review but gives little to no benefit to the project.

PR #623, opened 2026-08-30
OpenLegacyBoot: Add grub4dos (grldr) PBR detection support

bugprogrammer:
## Summary

Adds grub4dos (`grldr`) PBR detection to OpenLegacyBoot, allowing OC to
chain-load grub4dos as a standalone legacy boot entry.

## Background

OpenLegacyBoot currently detects `NTLDR`, `BOOTMGR`, and `ISOLINUX` PBR
signatures but not `grldr`. grub4dos is a legacy (MBR) boot loader
chain-loaded via OpenLegacyBoot to boot NT5 systems (Windows 2000/XP).

Unlike Vista/7, which can boot natively via UEFI+GPT (requiring CSM only
for video), NT5 requires a full legacy BIOS environment and cannot boot
through UEFI at all. In OpenDuet or CSM-based UEFI systems, `ntldr`'s
strict `DL=0x80` requirement often fails because the firmware's disk
enumeration differs from native BIOS. grub4dos serves as an intermediary
that performs the necessary drive mapping (`map () (hd0)`) before
chain-loading `ntldr`.

With grldr detection, each NT5 installation can have its own grub4dos
PBR and appear as an independent entry in the OC picker, rather than
relying on a single shared grub4dos menu with multiple entries.

## Changes

- `LegacyBootInternal.h`: Add `OcLegacyOsTypeGrldr` enum value.
- `LegacyBootSupport.c`: Detect both `"GRLDR"` and `"grldr"` PBR
  signatures. Case-insensitive search is required because FAT32 stores
  directory entries in uppercase while NTFS preserves the original
  lowercase filename.
- `OpenLegacyBoot.c`: Add `Grub4dos (legacy)` entry name. Flavour reuses
  `OC_FLAVOUR_WINDOWS` (consistent with all existing legacy entries), so
  no new theme assets are required.

The existing `NTLDR` detection branch is preserved unchanged — partitions
without grub4dos installed continue to appear as `Windows (legacy)`.

## Testing

- **EP45 (OpenDuet) + Windows 2000**: OC detects grldr PBR, chain-loads
  grub4dos, boots Win2000 successfully.
- **Z77 (UEFI + CSM) + Windows XP**: Same flow, boots successfully.
- Native BIOS boot (without OC): grub4dos menu.lst with conditional
  `map` works correctly.

## Recommended menu.lst

For environments where the target disk may not be `hd0` (e.g. OpenDuet
disk enumeration), a conditional map avoids errors on native BIOS boot:

```
find --set-root /ntldr
if not "%@root%"=="(hd0,0)" map () (hd0) && map --hook
chainloader /ntldr
```

### DIFF
--- a/Platform/OpenLegacyBoot/LegacyBootInternal.h
+++ b/Platform/OpenLegacyBoot/LegacyBootInternal.h
@@ -34,7 +34,8 @@ typedef enum OC_LEGACY_OS_TYPE_ {
   OcLegacyOsTypeNone,
   OcLegacyOsTypeWindowsNtldr,
   OcLegacyOsTypeWindowsBootmgr,
-  OcLegacyOsTypeIsoLinux
+  OcLegacyOsTypeIsoLinux,
+  OcLegacyOsTypeGrldr
 } OC_LEGACY_OS_TYPE;
 
 EFI_STATUS
--- a/Platform/OpenLegacyBoot/LegacyBootSupport.c
+++ b/Platform/OpenLegacyBoot/LegacyBootSupport.c
@@ -353,6 +353,10 @@ InternalGetPartitionLegacyOsType (
             || CheckLegacySignature ("isolinux", Buffer, BufferSize))
   {
     LegacyOsType = OcLegacyOsTypeIsoLinux;
+  } else if (  CheckLegacySignature ("GRLDR", Buffer, BufferSize)
+            || CheckLegacySignature ("grldr", Buffer, BufferSize))
+  {
+    LegacyOsType = OcLegacyOsTypeGrldr;
   } else {
     LegacyOsType = OcLegacyOsTypeNone;
     DEBUG ((DEBUG_INFO, "OLB: Unknown legacy bootsector signature\n"));
--- a/Platform/OpenLegacyBoot/OpenLegacyBoot.c
+++ b/Platform/OpenLegacyBoot/OpenLegacyBoot.c
@@ -34,7 +34,25 @@ GetLegacyEntryName (
   OC_LEGACY_OS_TYPE  LegacyOsType
   )
 {
-  return AllocateCopyPool (L_STR_SIZE ("Windows (legacy)"), "Windows (legacy)");
+  CONST CHAR8  *Name;
+
+  switch (LegacyOsType) {
+    case OcLegacyOsTypeIsoLinux:
+      Name = "Linux (legacy)";
+      break;
+
+    case OcLegacyOsTypeGrldr:
+      Name = "Grub4dos (legacy)";
+      break;
+
+    case OcLegacyOsTypeWindowsBootmgr:
+    case OcLegacyOsTypeWindowsNtldr:
+    default:
+      Name = "Windows (legacy)";
+      break;
+  }
+
+  return AllocateCopyPool (AsciiStrSize (Name), Name);
 }
 
 STATIC


vit9696:
The change simply changes the name of the boot entry from what I understand. I.e. previously grub4dos would be detected as `Windows (legacy)` now it would be called `Grub4dos (legacy)`. Does the boot entry title really matter given it is still essentially Windows, although booted via grub?

bugprogrammer:
> The change simply changes the name of the boot entry from what I understand. I.e. previously grub4dos would be detected as `Windows (legacy)` now it would be called `Grub4dos (legacy)`. Does the boot entry title really matter given it is still essentially Windows, although booted via grub?

This is not just a title change. The core of this PR is adding **grldr (grub4dos) PBR signature detection** to `InternalDetectLegacyOsType` in `LegacyBootSupport.c`.

Previously, OpenLegacyBoot only recognized three PBR signatures:

```
if (CheckLegacySignature ("BOOTMGR", Buffer, BufferSize)) {
    LegacyOsType = OcLegacyOsTypeWindowsBootmgr;
} else if (CheckLegacySignature ("NTLDR", Buffer, BufferSize)) {
    LegacyOsType = OcLegacyOsTypeWindowsNtldr;
} else if (  CheckLegacySignature ("ISOLINUX", Buffer, BufferSize)
          || CheckLegacySignature ("isolinux", Buffer, BufferSize))
{
    LegacyOsType = OcLegacyOsTypeIsoLinux;
} else {
    LegacyOsType = OcLegacyOsTypeNone;  // grldr fell here — partition was invisible
}
```

A partition with grub4dos installed to its PBR contains the `grldr` signature, which did not match any of the above, so `LegacyOsType` was `OcLegacyOsTypeNone` and the partition never appeared in the OC boot menu.

This PR adds a new detection branch:

```
} else if (  CheckLegacySignature ("GRLDR", Buffer, BufferSize)
          || CheckLegacySignature ("grldr", Buffer, BufferSize))
{
    LegacyOsType = OcLegacyOsTypeGrldr;
}
```

The boot entry title ("Grub4dos (legacy)") is adapted accordingly as a consequence, not the purpose of the PR.

The use case is booting Windows 2000/XP (NT5): OpenLegacyBoot's direct ntldr boot has a DL register mapping issue on non-0x80 disks (ntldr requires the boot disk to be `0x80`). grub4dos's `map` command solves this. With each NT5 system having its own grub4dos PBR, OC can list and boot them independently, rather than funneling everything through one grub4dos menu.

vit9696:
Right, might you want it to be detected as Windows for more native icons though? If not, I could merge as is.

bugprogrammer:
> Right, might you want it to be detected as Windows for more native icons though? If not, I could merge as is.

Please merge as-is. Keeping the title as "Grub4dos (legacy)" is intentional during the setup phase: grub4dos requires configuring `menu.lst` (disk mapping, chainloading targets, etc.), and being able to distinguish the grub4dos entry from native ntldr entries in the picker is essential for debugging.

Once configured, the entry name can be customized per-partition using OpenCore's built-in disk label mechanism, which already works with legacy boot entries:

1. Enable the disk label attribute in `config.plist`:
```
Misc -> PickerAttributes -> OC_ATTR_USE_DISK_LABEL_FILE (0x80)
```
2. Load `OpenNTFS.efi` driver so OC can read files from NTFS partitions.
3. Place a `.disk_label.contentDetails` file at the root of the target partition containing the desired display name (e.g. `Windows XP`).

OpenLegacyBoot already checks for this file via `LoadAppleDiskLabel()` and uses its content to override the default entry name. No code changes are needed for renaming — the default "Grub4dos (legacy)" title serves as a sensible fallback when no custom label is present.

PR #624, opened 2026-09-01
OpenCanopy: Add support for the optical drive icon

ilikesn0w:
Adds support for optical drive icon in OpenCanopy.

I'm also unsure about the icon name, maybe instead of "OpticalDrive," I should use "GenericCd" or something else?


vit9696:
Seems fine to me, thx.

CobanRamo:
I believe this hasn't been fully thought through. Due to this change, the "OpticalDrive.icns" and "ExtOpticalDrive.icns" files are now missing everywhere, which prevents OpenCanopy.efi from starting and causes it to fall back to internal mode. Wouldn't it be better to at least display an error message when a file is missing and then let the program continue running? In my opinion, the majority of users won't understand what's wrong without a debug build and log, which will only lead to unnecessary frustration.

Bildschirmfoto 2026-09-06 um 13 21 33


[OpticalDrive.zip](https://github.com/user-attachments/files/31879222/OpticalDrive.zip)

ilikesn0w:
@CobanRamo Yes, you're right, OpenCanopy won't launch without OpticalDrive.icns, but ExtOpticalDrive.icns is optional. To resolve this issue, I suggest adding ICON_GENERIC_CD after ICON_APPLE = ICON_NUM_MANDATORY, that way, if it's missing, the system will fall back to the HDD icon.
CC @vit9696 If you agree with what I said above, I'm ready to open a PR

vit9696:
Sounds good. It would also make sense to contribute OpticalDrive.icns to OcBinaryData.

ilikesn0w:
> contribute OpticalDrive.icns to OcBinaryData.

I can get it from the Mac firmware and open PR, but https://github.com/acidanthera/OcBinaryData/pull/7#issuecomment-830318009 says that only Acidanthera members can contribute to OcBinaryData.

vit9696:
For image-like resources it is fine. For device drivers we generally do not accept for safety reasons as we cannot control this.

PR #626, opened 2026-09-09
OcDevicePathLib: Fix unresolved ATAPI paths on NVMe Macs

rctphone:
On a MacBookPro14,2 with an original Apple NVMe SSD, macOS 15.7.9 creates a `boot-image` path containing `Ata(...)`, while firmware exposes that disk through Apple's NVMe node. The unresolved path causes `boot.efi` to fail its Block I/O lookup with `EFI_NOT_FOUND` and restart instead of restoring the suspended session.

Handle ATAPI expansion in the existing `OcDevicePathLib` resolver. Find a unique NVMe disk path under the same prefix, preserve all following nodes, and verify Block I/O resolution. This covers whole-disk, partition, APFS and file paths, including hibernation. Device-path structure excludes partition handles; media flags do not exclude removable NVMe devices or firmware with incomplete media metadata. Firmware device paths use the same validity assumptions as the surrounding code.

Allocation ownership follows the existing SATA expansion branch, including calls without a restore context. The generic resolver also releases its saved original path after successful expansion; the missing cleanup was reproduced in the existing VirtIO handler.

Validation for `7b1a034`:

- 29 public host scenarios pass on arm64 and x86_64 with ASan/UBSan, DEBUG, Werror and pool-allocation counters.
- Coverage includes extended ATAPI nodes, multiple instances, disk/partition/APFS/file suffixes, removable NVMe, ambiguous matches, rejected replacement resolution, rollback and cleanup, and ACPI correction followed by ATAPI expansion with no restore context.
- Removing the successful-path cleanup reproduces a leak in the existing VirtIO expansion. Removing nullable-context handling causes the ACPI-to-ATAPI scenario to fail under the sanitizers.
- Official Uncrustify and `git diff --check` pass. All nine executed CI checks pass for this revision, including [macOS XCODE5, Windows VS2022, Linux GCC, CLANGPDB, CLANGDWARF and Docs](https://github.com/acidanthera/OpenCorePkg/actions/runs/34945552243), codestyle, Python and shell checks. Coverity and Documentation analysis were skipped by the workflow conditions.

Earlier hardware validation covers the initial hibernation-specific implementation: DEBUG and RELEASE OpenCore 1.0.4 restored forced hibernation on MacBookPro14,2 / macOS 15.7.9 with FileVault enabled, retaining the kernel boot UUID and resuming the suspended process. This generalized revision has not been installed on the test Mac. Machine-specific fixtures, raw dumps and firmware binaries are excluded.


Library/OcBootManagementLib/AppleHibernate.c
```diff
@@ -88,6 +92,15 @@ OcActivateHibernateWake (
                           &BootImagePath,
                           &RemainingPath
                           );
+      if (InternalFixAppleHibernateDevicePath (&BootImagePath)) {
```
vit9696, line 95: Why a separate function is needed?

rctphone: I separated it to limit the initial change to the captured hibernation image path and avoid changing other callers of OcFixAppleBootDevicePath. The existing partition-based expansion does not cover this whole-disk path with an image offset/volume suffix.

There is no fundamental requirement for a separate call here, though; the fallback can live in OcDevicePathLib and be invoked by the existing resolver, retaining the strict path-shape and unique-disk checks and its allocation/restore handling.

vit9696: I just get a suspect the fix applies pretty much universally and there is no big need to separate, they are unlikely to conflict.

rctphone: Agreed. The repair now runs inside `OcFixAppleBootDevicePathNode`, so the existing resolver handles both hibernation and ordinary boot paths. I removed the extra hibernation call and the image-only suffix restriction. The replacement preserves disk, partition, APFS and file suffixes while retaining exact-prefix matching, unique-disk selection and Block I/O verification; already working paths remain unchanged.

The tests now exercise the generic resolver, including partition-level resolution and the existing restore API. They also caught a missing cleanup of the saved original path after a successful expansion, which this revision fixes. All 25 public scenarios, plus the private captured image path locally, pass on arm64 and x86_64 with sanitizers and allocation-counter checks. Hardware restore validation still covers the earlier implementation; this revision has not yet been installed on the test Mac.

CI update for `f832ec5`: macOS XCODE5, Windows VS2022, Linux GCC, Linux CLANGDWARF, Linux Docs, codestyle, Python and shell checks all passed. [Linux CLANGPDB](https://github.com/acidanthera/OpenCorePkg/actions/runs/34768287900/job/103753108499) failed while preparing its Docker environment, before compiling OpenCore: the downloaded apt.llvm.org installer reported Ubuntu 24.04.4 as unsupported after five attempts. GitHub rejected my rerun request with HTTP 403 because I lack upstream repository permissions. Could you please rerun that job? Hardware validation of this generalized revision is still pending.

Library/OcDevicePathLib/OcDevicePathLib.c
```diff
@@ -408,6 +409,156 @@ InternalExpandNewPath (
...
+  EFI_DEVICE_PATH_PROTOCOL  *MatchPath;
+  EFI_HANDLE                *Handles;
+  EFI_HANDLE                Match;
+  EFI_BLOCK_IO_PROTOCOL     *BlockIo;
+  EFI_STATUS                Status;
+  UINTN                     PrefixSize;
+  UINTN                     Index;
+  UINTN                     Count;
+
+  if (  (DevicePathNodeLength (*DevicePathNode) != sizeof (ATAPI_DEVICE_PATH))
```
vit9696: This looks like garbage to be honest:

* It forbids ATAPI expansion, but there is no reason for it.
* It checks for multi instance, but there is no other check in any function for it.

Please remove.

Library/OcDevicePathLib/OcDevicePathLib.c
```diff
@@ -408,6 +409,156 @@ InternalExpandNewPath (
...
+                    Handles[Index],
+                    &gEfiBlockIoProtocolGuid,
+                    (VOID **)&BlockIo
+                    );
+    if (  EFI_ERROR (Status) || (BlockIo->Media == NULL)
+       || !BlockIo->Media->MediaPresent || BlockIo->Media->LogicalPartition
+       || BlockIo->Media->RemovableMedia)
+    {
+      continue;
+    }
```
vit9696: This is technically garbage. You have already checked EfiBlockIoProtocol is present when you got a handle buffer, so handling a protocol is unnecessary.

As for checking for other properties, it is also unnecessary:
- MediaPresent is unused anywhere here, if anything does not set it, we never know and never care.
- LogicalPartition will be checked by device node type after prefix, you ensure it is the last one, and check against NVMe/SAS, thus it will filter out logical partitions.
- RemovableMedia looks just fine to me, what if we have Thunderbolt NVMe for instance.

Just drop this.

Library/OcDevicePathLib/OcDevicePathLib.c
```diff
@@ -408,6 +409,156 @@ InternalExpandNewPath (
...
+
+    Status = gBS->HandleProtocol (
+                    Handles[Index],
+                    &gEfiDevicePathProtocolGuid,
+                    (VOID **)&DiskPath
+                    );
+    if (  EFI_ERROR (Status) || (DiskPath == NULL) || !IsDevicePathValid (DiskPath, 0)
```
vit9696: We trust DevicePath to be valid everywhere else in the code. Please drop IsDevicePathValid check.

Library/OcDevicePathLib/OcDevicePathLib.c
```diff
@@ -408,6 +409,156 @@ InternalExpandNewPath (
...
+       || (GetDevicePathSize (DiskPath) != PrefixSize + sizeof (NVME_NAMESPACE_DEVICE_PATH) + END_DEVICE_PATH_LENGTH)
+       || (CompareMem (DiskPath, *DevicePath, PrefixSize) != 0))
+    {
+      continue;
+    }
+
+    DiskNode = (EFI_DEVICE_PATH_PROTOCOL *)((UINT8 *)DiskPath + PrefixSize);
```
vit9696: Reuse `Node` since you have it, extra var does not give you anything here.

Library/OcDevicePathLib/OcDevicePathLib.c
```diff
@@ -408,6 +409,156 @@ InternalExpandNewPath (
...
+  //
+  FixedPath = AppendDevicePath (MatchPath, Suffix);
+  if (FixedPath == NULL) {
+    return 0;
+  }
+
+  Node   = FixedPath;
+  Status = gBS->LocateDevicePath (&gEfiBlockIoProtocolGuid, &Node, &Match);
+  if (  EFI_ERROR (Status)
+     || ((UINTN)Node - (UINTN)FixedPath < PrefixSize + sizeof (NVME_NAMESPACE_DEVICE_PATH)))
```
vit9696: Checking for just EFI_ERROR is enough. This is how the rest of the code does it.

Library/OcDevicePathLib/OcDevicePathLib.c
```diff
@@ -408,6 +409,156 @@ InternalExpandNewPath (
...
+    FreePool (FixedPath);
+    return 0;
+  }
+
+  DEBUG ((DEBUG_INFO, "OCDP: Fixed ATAPI path using matching NVMe disk\n"));
+  if (RestoreContext != NULL) {
+    RestoreContext->OldPath = *DevicePath;
+  } else {
+    FreePool (*DevicePath);
+  }
```
vit9696: Drop the `if`, if we got here, RestoreContext is meant to be present. See e.g. InternalFixAppleBootDevicePathVirtioNode.

Library/OcDevicePathLib/OcDevicePathLib.c
```diff
@@ -913,6 +1071,16 @@ OcFixAppleBootDevicePath (
   ASSERT ((UINTN)*RemainingDevicePath < ((UINTN)*DevicePath) + DevicePathSize);
   DEBUG_CODE_END ();
 
+  //
+  // A successful expansion no longer needs the original path for rollback.
+  //
+  if (NodePatched != 0) {
+    OcFixAppleBootDevicePathNodeRestoreFree (
```
vit9696, line 1053: Is this a bugfix? We did not free it before, but I am starting to wonder if it was a mistake even then?

vit9696 (commented):
Left few comments here and there. Is this neuro-generated? Many checks look excessive, and style/workflow generally does not follow the existing codebase.

Why did not you just copy-paste something like InternalFixAppleBootDevicePathVirtioNode and adopt to your needs?

Anyway, I think we can more or less merge this after you cleanup the comments.

Library/OcDevicePathLib/OcDevicePathLib.c
rctphone: I checked this with an ACPI correction followed by ATAPI expansion: OcFixAppleBootDevicePath sets RestoreContextPtr to NULL after its first iteration, so the ATAPI call can receive NULL. The short-form caller in BootEntryManagement.c also explicitly passes NULL. Removing this handling makes the chained regression fail under UBSan/ASan with a null-pointer access.

I removed RestoreContext from the helper entirely and moved ownership handling into OcFixAppleBootDevicePathNode, following the existing SATA expansion branch. The tests now cover both the chained case and a direct call without a context.

Library/OcDevicePathLib/OcDevicePathLib.c
rctphone: Yes, this fixes a pre-existing leak. On a successful VirtIO expansion, InternalFixAppleBootDevicePathVirtioNode saves the original allocation in RestoreContext->OldPath, but the normal-success exit of OcFixAppleBootDevicePath previously did not free it.

I added a regression using the existing VirtIO handler. With the cleanup removed it reports an outstanding pool allocation; with the cleanup present it passes. This case runs before any ATAPI test, so it reproduces independently of the new handler.

rctphone:
Yes, AI tools were used for this patch. The earlier version did not follow the surrounding code closely enough.

Updated in 7b1a034: removed the ATAPI-length and multi-instance restrictions, redundant Block I/O handling and media filters, firmware-path validation, the extra node variable, and the replacement-offset check. Ownership now follows the existing SATA expansion branch. I added replies below with reproductions for the nullable context and the pre-existing cleanup bug.

All 29 host scenarios pass on arm64 and x86_64 with ASan/UBSan and allocation counters; Uncrustify also passes. Thank you for rerunning CLANGPDB for the previous revision.

CI update: all nine executed checks now pass for 7b1a034, including macOS, Windows and all three Linux toolchains.

PR #628, opened 2026-09-16
Misc: Revise EDK II metadata files

ilikesn0w:
- Removed `UEFI_HII_RESOURCE_SECTION` from INF's since this flag isn't needed
- Removed `EDK_RELEASE_VERSION` and `EFI_SPECIFICATION_VERSION` from OpenCore, OpenCanopy and Test drivers\apps since these flags are not needed
- Removed `gEfiHiiDatabaseProtocolGuid` from `ControlMsrE2.inf` since it isn't used
- Renamed entry points in CryptoTest to match general codestyle
- Updated INF's specs from `0x00010005` to `0x0001001B`
- Fixed INF file in OcMachoLib
- Updated DSC's specs from `0x00010006` to `0x0001001C`
- Updated DEC's specs from `0x00010005` to `0x0001001B`
- Removed unnecessary spaces and empty sections from INF files
- Converted GUIDs in some INF files to uppercase
- Fixed comment typos in some INF files


Library/OcPciIoLib/OcPciIoLib.inf
```diff
@@ -1,6 +1,6 @@
 ## @file
 #
-#  Component description file for the library producing the English version of Unicode Collation Protocol.
+#  Component description file for the library producing the Cpu IO 2 and PCI Root Bridge IO protocols overrides.
 #
 #  Copyright (c) 2006 - 2011, Intel Corporation. All rights reserved.<BR>
```
ilikesn0w, line 5: BTW, I think it would be a good idea to replace these copyrights with @xCuri0 copyrights, but I'm not sure if I have the right to do that.

vit9696: I guess he can change them if he wants, do not think anyone cares.

PR #630, opened 2026-09-17
OcCryptoLib: Implement HMAC support

ilikesn0w:
Implemented HMAC support in OcCryptoLib for MD5, SHA-1, SHA-256/384/512 and added HMAC tests to CryptoTest

I also noticed two things: the Cyrillic character `№` was used in `TestHash`, but since it isn't available in all console fonts, I replaced it with `#`, also, in `HashSamples,` the number 4 was used instead of `HASH_SAMPLES_NUM`.


Include/Acidanthera/Library/OcCryptoLib.h
```diff
@@ -51,6 +51,8 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 //
 // Block sizes.
 //
+#define MD5_BLOCK_SIZE     64
+#define SHA1_BLOCK_SIZE    64
```
ilikesn0w, line 55: Maybe move `MD5_BLOCK_SIZE` and `SHA1_BLOCK_SIZE` below `SHA384_BLOCK_SIZE` and make them `#define MD5_BLOCK_SIZE SHA256_BLOCK_SIZE` and `#define SHA1_BLOCK_SIZE SHA256_BLOCK_SIZE`?

vit9696:
Thank you for the patch! Could you please clarify before I start reviewing what are you trying to achieve with this change? I.e. who is the user for HMAC?

ilikesn0w:
I just want to expand the library's functionality, btw Apple uses HMAC SHA-256 in Tamper Resistant Boot protocol (https://reverse.put.as/2016/06/25/apple-efi-firmware-passwords-and-the-scbo-myth/#:~:text=Message%20Autentication%20Code%20(MAC)%20using%20SHA256) and I can't rule out the possibility that an implementation of this protocol might appear in OpenCore at some point

PR #632, opened 2026-09-19
Implement SimpleInputEx protocol on top of AppleEvent

### LINKED ISSUE #1080: Implement SimpleInputEx protocol on top of SimpleInput or AppleEvent
SimpleInputEx is unavailable on many older laptops as well as legacy Mac (prior to Skylake) firmwares. This makes `edit` Shell command completely unavailable on such machines. It will fail to initialise with an error. We could try implementing this protocol on top of an existing protocol to permit at least some kind of functioning.

ilikesn0w:
Implemented `SimpleInputEx` protocol on top of `AppleEvent`. Tested it with `edit` in OpenShell on `MacBookPro9,2`, and it fully works

I've commented on a few moments below where I'm unsure.

Closes: https://github.com/acidanthera/bugtracker/issues/1080


Library/OcSimpleTextInputLib/OcSimpleTextInputLibInternal.h
ilikesn0w, line 1: I'm not sure about this, do we really need an internal header?

